PR #13, opened 2014-11-22
Webhooks

jbrooksuk:
I've started work on Web Hooks which will allow Cachet to post to other services on updates.

### DIFF
--- a/app/models/WebHookContent.php
+++ b/app/models/WebHookContent.php
@@ -1,5 +1,7 @@
 <?php
 
 	class WebHookContent extends Eloquent {
-
+		public function hook() {
+			return $this->belongsTo('WebHook', 'id', 'hook_id');
+		}
 	}


RickB17:
Does Cachet support using webhooks to create/update incidents?

jbrooksuk:
We actually ended up removing web hooks. They will be making a comeback at some point for some services. In the meantime, you can use the API 😄

RickB17:
Thanks for the quick reply. The API's are working great, I was hoping for easy integration into JIRA via webhooks. I'll look at possibly using Zapier or some other service.

Do you have any recommendations for a solution I could look at that would not require going to an external network to process the webhook request and act as a connector between JIRA and Cachet?

jbrooksuk:
Well, you could look at https://packagist.org/packages/mrbase/cachet-slack-integration and see how @mrbase hooks into the event system. This will require some code changes though.

mrbase:
Yeah, the slack integration is a bit off at the moment tho. The event system changed since I last used it - but feel free :)

jbrooksuk:
v2.4 has a lot more events now 😉

PR #174, opened 2015-01-01
Update 404 to show only on dashboard and api

joecohens:
I think if is the users status page just redirect them to the status page, if it's a 404 on the dashboard then we show the 404 page. 

screen shot 2014-12-31 at 7 13 06 pm

### DIFF
--- a/app/start/global.php
+++ b/app/start/global.php
@@ -57,9 +57,13 @@ API::error(function (\Illuminate\Database\Eloquent\ModelNotFoundException $excep
 });
 
 App::missing(function ($exception) {
-    return Response::view('errors.404', [
-        'pageTitle' => 'That page went missing!'
-    ], 404);
+    if (Request::is('dashboard*') or Request::is('api*')) {
+        return Response::view('errors.404', [
+            'pageTitle' => 'That page went missing!',
+        ], 404);
+    }
+
+    return Redirect::route('status-page');
 });
 
 /*
--- a/app/lang/en/cachet.php
+++ b/app/lang/en/cachet.php
@@ -55,6 +55,9 @@ return [
         'logged_incidents'         => '{0} There are no incidents, good work.|You have logged one incident.|You have reported <strong>:count</strong> incidents.',
         'settings_saved'           => 'Settings saved.',
         'settings_not_saved'       => 'Settings could not be saved.',
+        'not_found'                => 'Page Not Found',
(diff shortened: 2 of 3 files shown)


app/start/global.php
```diff
@@ -57,9 +57,13 @@
 });
 
 App::missing(function ($exception) {
-    return Response::view('errors.404', [
-        'pageTitle' => 'That page went missing!'
-    ], 404);
+    if (Request::is('dashboard*') or Request::is('api*')) {
+        return Response::view('errors.404', [
+            'pageTitle' => 'That page went missing!'
```
GrahamCampbell: We need a trailing comma here to pass the cs checks.

GrahamCampbell: Or, at least, we would do if that fixer wasn't broken...

joecohens: Sure! I forget PSR-2 :)

joecohens: Hahaha

jbrooksuk:
Merge when ready :)

app/start/global.php
GrahamCampbell: That's not actually a psr-2 rule, but a rule from symfony's coding standards. I've got many of those fixers enabled on styleci because it enforces extra consistency, the same way I'm forcing short array syntax. psr-2 says nothing about that, and actually neither does symfony.

joecohens: That's right, they only set them if it's a multi-line array.

GrahamCampbell: yeh, and this is a mutliline array.

GrahamCampbell: it only has an element on one line, but it is split over 3 lines.

GrahamCampbell: Like this is a multiline function:
```php
function foo(
    $foo
( {
    //
}
```

joecohens: Ok ok ;)

GrahamCampbell: I've got a feeling we got that wrong in php-cs-fixer... We messed up the multiline function stuff before...

joecohens: Haha maybe. Last commit it was passing.

GrahamCampbell: it's still technically passing now

joecohens: Do you want me to leave the PR open to check fi the cs-fixer it's ok?

GrahamCampbell: Nah, just merge it. Trailing commas are really trivial.

joecohens: Ok !

PR #178, opened 2015-01-01
Prevent double submit, compose start.js, added messenger

joecohens:
- Added globals for ajax for csrf_token and auth with.
- CachetHQ namespace js.
- Added libs lodash, messenger.
- Prevent double submit on forms.
- Created CachetHQ Notifier object. Try (new CachetHQ.Notifier).notify('Error occurred');
- Runed gulp --production to minify css and js.


jbrooksuk:
This is so much better!

joecohens:
:)

PR #185, opened 2015-01-01
View refactor

jbrooksuk:
I've moved the dashboard view files into their own sub-folders to make it tidier. Plus, I used @joecohens `set_active` helper to make managing the sidebar easier.


app/helpers.php
```diff
@@ -23,3 +23,26 @@ function elixir($file)
         throw new InvalidArgumentException("File {$file} not defined in asset manifest.");
     }
 }
+
+if (! function_exists('set_active')) {
+
+    /**
+     * Set active class if request is in path.
+     *
+     * @param $path
```
GrahamCampbell: missing type hint

jbrooksuk: Will fix now.

PR #171, opened 2015-01-01
Banner can now be removed

Jamesking56:
This may need a bit of cleaning to be more inline with your project but I hope this is alright. Fixes #168

PR #187, opened 2015-01-01
Banner can now be removed

GrahamCampbell:
Closes #171.

### DIFF
--- a/app/views/dashboard/settings/app-setup.blade.php
+++ b/app/views/dashboard/settings/app-setup.blade.php
@@ -13,7 +13,7 @@
         <div class="content-wrapper">
             <div class="row">
                 <div class="col-sm-12">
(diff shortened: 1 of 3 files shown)


app/views/dashboard/settings/app-setup.blade.php
```diff
@@ -52,8 +52,10 @@
                                     <div class='form-group'>
                                         <label>Banner Image</label>
                                         @if($banner = Setting::get('app_banner'))
-                                        <div class='well'>
+                                        <div id="banner-view" class='well'>
                                             <img src='data:{{ Setting::get("app_banner_type") }};base64,{{ $banner }}' style='max-width: 100%' />
+                                            <br /><br />
```
GrahamCampbell: not sure about this

jbrooksuk: There isn't a proper class yet, which means there is no decent margin at the bottom. We'll sort that soon.

PR #195, opened 2015-01-01
Move code to src folder

GrahamCampbell:
This is the next part of my refactoring. Again, this will help us when we come to upgrade to laravel 5 at a later point. The best way to upgrade to laravel is actually to delete the entire app, and port your code over to the new structure. Following this refactoring, I'll be moving the controllers/models/filters into the src folder, and namespacing them.

NB, this refactor should not have actually changed any code so far, just the directory structure.


GrahamCampbell:
@jbrooksuk Are you ok this this, so I can continue?

jbrooksuk:
Yeah, this is fine. Thanks for the heads-up.

PR #179, opened 2015-01-01
[Proposal] Artisan command for one click installation services

joecohens:
This is a proposal to this issue https://github.com/cachethq/Cachet/pull/96

PR #199, opened 2015-01-01
Added a one click deploy command

GrahamCampbell:
Replaces #179. NB, I haven't actually checked this works, but it should do provided there are no typos.


app/.env.example.php
```diff
@@ -0,0 +1,9 @@
+<?php
```
GrahamCampbell: @joecohens Why did you put this in the app folder btw?

joecohens: Just has an example of a https://github.com/cachethq/Cachet/blob/master/INSTALL.md kind of what dotenv does but just to reference not any use.

GrahamCampbell: dotenv in laravel 5 doesn't do that i don't think

GrahamCampbell: probably better in the root folder

joecohens: Oh sorry my bad it should be on the root folder !

GrahamCampbell: i'll move it...

joecohens: No in Laravel 5 it does other things this is just a reference.

joecohens: Thanks!

jbrooksuk:
@joecohens can you test this please.

joecohens:
Sure!! Testing...

src/Providers/RoutingServiceProvider.php
```diff
@@ -7,22 +7,12 @@
...
-     * Organise Routes.
+     * Boot the service provider.
      *
-     * @param string $app
+     * @return void
      */
-    private function routesInDirectory($app = '')
+    public function boot()
     {
         $routeDir = app_path('routes/'.$app.($app !== '' ? '/' : null));
```
joecohens: Undefined variable: app

GrahamCampbell: one sec...

GrahamCampbell: was this code copied and pasted from somewhere?

GrahamCampbell: I'm going to totally remove it because it doesn't really do what we need, and replace it with something more useful.

joecohens: Nop I don't know why it has changes. https://github.com/cachethq/Cachet/blob/master/src/Providers/RoutingServiceProvider.php

GrahamCampbell: alrighty, i've rewritten that stuff now. way, way simpler

jbrooksuk: The original `routesInDirectory` was written by me.

GrahamCampbell: Ok, It was massively over complicated. How does my replacement look?

jbrooksuk: @GrahamCampbell based on [this](http://james-brooks.uk/caching-routes-with-laravel-4/). Yours looks great though!

src/Providers/ConsoleServiceProvider.php
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    /**
+     * Register the service provider.
+     *
+     * @return void
+     */
+    public function register()
+    {
+        $this->app->singleton('CachetHQ\Cachet\Commands\OneClickDeployCommand', function () {
```
joecohens: Missing `use ($app)`

GrahamCampbell: actually, just just be `function ($app)`

joecohens: :+1:

GrahamCampbell:
Does this work for you now @joecohens?

joecohens:
Woking now! Squash an merge?

PR #200, opened 2015-01-01
Improved JS

jbrooksuk:
Tidies up the JS source code against [this](https://codeclimate.com/github/cachethq/Cachet/app/assets/js/start.js).


joecohens:
We can add jshint to the project :) https://www.npmjs.com/package/laravel-elixir-jshint

jbrooksuk:
Yeah, we should do that!

PR #211, opened 2015-01-01
Cleaned Up The Models And Repositories

GrahamCampbell:
Note that this pull request doesn't actually namespace the models, just cleans them up.

### DIFF
--- a/app/models/Component.php
+++ b/app/models/Component.php
@@ -2,14 +2,21 @@
 
 use CachetHQ\Cachet\Transformers\ComponentTransformer;
 use Dingo\Api\Transformer\TransformableInterface;
+use Illuminate\Database\Eloquent\Builder;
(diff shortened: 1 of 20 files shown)


joecohens:
:+1: :+1:

GrahamCampbell:
How does this look now?

NB, this will probably need to be actually tested given that I've enforced some extra type hinting in some of the models based on the phpdoc, and it's possible that phpdoc could have been incorrect.

joecohens:
Looking good ! Have a question why string[] instead of array for php?

jbrooksuk:
Array of strings?

GrahamCampbell:
Yeh. We should always use more specific types instead of just `array` when possible.

For example, `Foo[]` is an array of Foo objects.

joecohens:
I had never done that in PHP I like it!

PR #226, opened 2015-01-01
fix typo

benwa:
`curl -XGET` is duplicated


docs/api/making-requests.md
```diff
@@ -8,7 +8,7 @@ The API is accessible from `http://<cachet-url>/api/<endpoint>`
 
 You can use curl to interact with the API. There is no authentication required for `GET` requests:
 
-`curl -XGET curl -XGET http://status.cachethq.io/api/components`
+`curl -XGET http://status.cachethq.io/api/components`
```
benwa: `-XGET` should be written as `-X GET`, however it will work like this (as well as without the option altogether)

PR #242, opened 2015-01-02
Fixed ignoring files

GrahamCampbell:


GrahamCampbell:
Github things this is a javascript project because of all the vendor files without this too. :/

jbrooksuk:
I think the build stuff can stay, because we have no setup yet to recompile. The actual assets can be removed though.

GrahamCampbell:
So what do you want to keep exactly?

jbrooksuk:
Anything in `public` basically. The `app/assets/bower_components/*` can be ignored though.

jbrooksuk:
They should be pulled down for development anyway.

jbrooksuk:
You can still remove `bower_components`.

GrahamCampbell:
Will just commit this to the master.

PR #244, opened 2015-01-02
Fix composer install

gjuric:
Fixes an error during installation, the error was:

      [RuntimeException]
        Could not scan for classes inside "tests/TestCase.php" which does not appear to be a file nor a folder

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,11 +38,6 @@
             "CachetHQ\\Cachet\\": "src/"
         }
     },
-    "autoload-dev": {
-        "classmap": [
-            "tests/TestCase.php"
-        ]
-    },
     "extra": {
         "paas": {
             "document-root": "public",


GrahamCampbell:
Install with `--no-dev`.

GrahamCampbell:
If you want dev dependencies, then you cannot download a zip from github. We exclude the tests from there.

gjuric:
I totally missed the fact that I haven't cloned the repo :) Installation instructions should probably be updated and linked to from somewhere on the main site.

GrahamCampbell:
Yeh, I'll make sure we recommend `--no-dev`.

GrahamCampbell:
There we go: https://github.com/cachethq/Cachet/commit/782b24210ee7f4f0a2c2e33759fd807ae095a086.

PR #246, opened 2015-01-02
Finally fixed the Gulp build process

jbrooksuk:
Right, this **absolutely** fixes the gulp build on my machine. I've ran `gulp` about 100 times now and not once has it kicked off about a missing file.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,64 +1,39 @@
 var elixir = require('laravel-elixir');
-var gulp = require('gulp');
-var changed = require('gulp-changed');
-var jshint = require('gulp-jshint');
-var stylish = require('jshint-stylish');
+var gulp   = require('gulp');
+var jshint = require('laravel-elixir-jshint');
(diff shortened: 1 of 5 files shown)


jbrooksuk:
Also, I've added an elixir compatible jshint check.

GrahamCampbell:
Does minification of the js work?

GrahamCampbell:
That would be very useful.

jbrooksuk:
Not yet, I'll do that now :)

GrahamCampbell:
That would fix github thinking this isn't a php repo too.

jbrooksuk:
I think the only way is to run `gulp --production`. Unless we look at Elixir...

GrahamCampbell:
That's probably fine then. Just make sure we always do that.

jbrooksuk:
I have an idea.

jbrooksuk:
Nailed it. Push incoming.

jbrooksuk:
Done.

PR #247, opened 2015-01-02
[WIP] Heroku build

jbrooksuk:
This branch will have Heroku automatically build itself with Gulp.


GrahamCampbell:
"Remove all build files" If you could do that after ignoring them, then it's not working...

jbrooksuk:
@GrahamCampbell I added it to the `.gitignore`, then had to remove them from the source, no?

GrahamCampbell:
No, ignoring them means git shouldn't even know you've deleted them. It literally means ignore, rather than don't store.

jbrooksuk:
But they're in the source currently, don't they need to be removed once they're added to the ignore list? If I didn't do that, then surely the existing files from the last commit would remain?

GrahamCampbell:
> don't they need to be removed once they're added to the ignore list

yes, but you ignored them first, so what i'm saying is your "ignore" can't be working correctly

jbrooksuk:
Ok, so talk me through what I should've done?

GrahamCampbell:
That's my point. I'm saying I don't know why git didn't ignore the files?

jbrooksuk:
It never does, I always add the pattern to `.gitignore`, commit it, but then the existing files aren't removed from the tree.

GrahamCampbell:
No, you not understanding what I mean by "ignore". I expect it to keep the existing files, and then whe you delete them, I expect it to do nothing.

jbrooksuk:
I think we have our wires crossed. If I ignore the files, I still want to remove the existing ones. I don't need to keep removing them.

jbrooksuk:
Maybe it's because I did it in two commits?

GrahamCampbell:
What i'm saying is if you were able to delete them after ignoring them, that means if you put files in that folder again, then it won't ignore them because it didn't before. See my concerne?

jbrooksuk:
I'm able to delete them because Git still has them committed from previously. They're ignored, but Git still knows they're in there.

GrahamCampbell:
Hmmm. Ok. Also, you made travis sad.

GrahamCampbell:
Maybe just call gulp in the one-click-setup command?

jbrooksuk:
Yeah, I think that's what we're going to have to do. I can't for the life of me make this work.

jbrooksuk:
My last commit didn't work either...

GrahamCampbell:
Yeh, I can sort this if you like?

jbrooksuk:
If you wouldn't mind? :)

GrahamCampbell:
Will do this soonish. Bit busy with other stuff atm.

composer.json
```diff
@@ -57,7 +57,8 @@
         "post-install-cmd": [
             "php artisan optimize",
             "chmod -R 755 app/storage",
-            "php artisan cachet:one-click-deploy"
+            "php artisan cachet:one-click-deploy",
+            "$(npm bin -q)/gulp"
```
GrahamCampbell: and what if we're on windows...

GrahamCampbell:
I think a better idea would be to get people to run the install with `--no-scripts` if then don't want to run the scripts. Quite simple really.

GrahamCampbell:
I'll prepare another pull request.

PR #248, opened 2015-01-02
Update documentation for PHP builtin server

Crash--:
To run Cachet from the PHP builtin server

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -102,6 +102,14 @@ $ php artisan db:seed
 
 # Running Cachet
 
+## PHP builtin server
+
+The easiest way to run Cachet for test or development purpose, it's to run it from the PHP builtin server. To achieve that:   
+```
+cd Cachet/
+php artisan serve
+```
+
 ## Apache
 
 Apache is one of the easier installations. We simply need to create a new Virtual Host and add it to our `HOSTS` file.


GrahamCampbell:
No, this is a bad idea. Also, the serve artisan command has been removed in laravel 5.

Crash--:
OK if the serve artisan has been removed in laravel 5. 

But I don't see why serving the app from the builtin server is a bad idea! It's a very common usage and it's pretty straightforward to test it...

GrahamCampbell:
I don't think we should encourage it given that it's removed, and running from the cli can give unexpected results, particularly with apcu if people don't configure it correctly.

jbrooksuk:
What you need to know is that when you run `artisan serve` you're not getting a proper web server, only an interface into PHP itself, so it'd be silly to support something that isn't ever properly used.

jbrooksuk:
And, FWIW, I'm able to run `artisan serve` and Cachet works (except for logging in).

GrahamCampbell:
> except for logging in

yeh, probably badly configured apc...

Crash--:
I'm logged within the app, and I can add incidents, etc without any error. (and without APC. I just needed to install apcu - you might add this requirement to your composer.json -). 

I just find that having to install / edit apache or nginx is "pretty complicated". Lot of applications use the builtin server for development or testing purpose (zf, sf...), and I can't imagine that Laravel will remove it at the end. 

With the builtin server, you have a proper web server to develop and to test... without no needed functionalities. 

Anyway, I'm not here to start a war but to contribute. I'll find something else to work on.

GrahamCampbell:
@Crash Yeh, apcu, that was what I was referring to. The driver is called apc.

jbrooksuk:
I'm all for having it in the docs as an unofficial support.

GrahamCampbell:
I don't like it. We'd only have to remove it again in 4 week's time.

jbrooksuk:
Ok :)

PR #251, opened 2015-01-02
BInput XSS protection using get instead of all - setup fails

ajtrichards:
When setting up a fresh install of Cachet it fails due to some missing arguments on `Binput::get()`. I don't think it should be using get so i've changed it back to `Binput::all()` and setup now works correctly.

### DIFF
--- a/src/Http/Controllers/SetupController.php
+++ b/src/Http/Controllers/SetupController.php
@@ -42,7 +42,7 @@ class SetupController extends Controller
      */
     public function postIndex()
     {
-        $postData = Binput::get();
+        $postData = Binput::all();
 
         $v = Validator::make($postData, [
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Thanks. Yeh, laravel allows get with no args, but I don't.

jbrooksuk:
@GrahamCampbell should we update Binput to allow that?

GrahamCampbell:
Nope. I've done it on purpose. I don't like how laravel allows it. I think get MUST be for getting a single item. and all MUST be for getting everything. :)

ajtrichards:
FWIW I agree with Graham's idea of getting a single item and being specific.

jbrooksuk:
Ok, cool!

PR #252, opened 2015-01-02
Feature/responsive sidebar

joecohens:
This is working on the main menu.

Need to make the same for the sub-menu and implement animate.css


GrahamCampbell:
Conflicts already?

joecohens:
Yea sorry I'm not in town since yesterday I just push it has a reference. I'll try to finish it tonight :)

jbrooksuk:
:+1:

joecohens:
I've rebased and compiled with --production. Responsive it's working on main menu but still needs some work on the settings sub menu but this ir ready to merge. @jbrooksuk Let me know what you think!

bower.json
```diff
@@ -1,16 +1,18 @@
+
```
GrahamCampbell: extra line needs removing

GrahamCampbell:
Also, `--production` shouldn't be required anymore. @jbrooksuk enabled it by default.

bower.json
joecohens: Thanks! Missed that on the rebase.

joecohens:
By default won't let us debug with out source maps on dev.. But I guess it's ok for now.

jbrooksuk:
@joecohens if you want to debug, then run `gulp --production=false` I think that'll work.

joecohens:
@jbrooksuk Thanks!

jbrooksuk:
The semi-responsive looks great!

jbrooksuk:
Although there could be less margin below the user profile.

joecohens:
Thanks! I think so too, also it would be cool to add some kind of animation.

jbrooksuk:
:+1: for the animation, but not too obtrusive.

Also, what about mobile? The nav disappears.

joecohens:
No just like size of the icons it would look fancy haha. And you're right I have to add a button to show on mobile.

jbrooksuk:
I love the mini-sidebar :)

joecohens:
:)

PR #258, opened 2015-01-03
[WIP] API Auth Key

jbrooksuk:
I've been playing around with using an api key for authorisation on the API. I've not fully tested the API yet, but when you supply an incorrect key, you do get an error back.

Needs more work:

- [ ] Generate default api key for users who may not currently have one.
- [x] Generate the api key at setup for the user.
- [x] Check the api still works with basic auth.

### DIFF
--- a/src/Models/User.php
+++ b/src/Models/User.php
@@ -7,6 +7,7 @@ use Illuminate\Auth\Reminders\RemindableTrait;
 use Illuminate\Auth\UserInterface;
 use Illuminate\Auth\UserTrait;
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
 use Illuminate\Support\Facades\Hash;
 
 /**
@@ -48,6 +49,20 @@ class User extends Model implements UserInterface, RemindableInterface
     protected $guarded = [];
 
     /**
+     * Overrides the models boot method.
+     *
+     * @return void
+     */
+    public static function boot()
+    {
+        parent::boot();
+
+        self::creating(function ($user) {
+            $user->api_key = self::generateApiKey();
+        });
+    }
+
+    /**
      * Hash any password being inserted by default.
      *
      * @param string $password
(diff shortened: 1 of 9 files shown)


jbrooksuk:
@GrahamCampbell have I registered the service provider properly here?

src/Models/User.php
```diff
@@ -72,4 +73,23 @@ public function getGravatarAttribute($size = 200)
...
         return sprintf('https://www.gravatar.com/avatar/%s?size=%d', md5($this->email), $size);
     }
+
+    /**
+     * Find by api_key, or throw an exception.
+     *
+     * @param string $api_key
+     * @param mixed  $columns
+     *
+     * @throws ModelNotFoundException if no matching User exists.
```
GrahamCampbell: should use the fqcn here

src/Models/User.php
```diff
@@ -72,4 +73,23 @@ public function getGravatarAttribute($size = 200)
     {
         return sprintf('https://www.gravatar.com/avatar/%s?size=%d', md5($this->email), $size);
     }
+
+    /**
+     * Find by api_key, or throw an exception.
+     *
+     * @param string $api_key
```
GrahamCampbell: should be `$apiKey`

GrahamCampbell: laravel should be clever enough to still work

src/Providers/AuthServiceProvider.php
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    /**
+     * Register the service provider.
+     *
+     * @return void
+     */
+    public function register()
+    {
+        $this->app->bindShared('CachetHQ\Cachet\Auth\ApiKeyAuthenticator', function ($app) {
```
GrahamCampbell: don't need `$app` there. it can be deleted

GrahamCampbell:
Just rebased this and pushed some fixes too.

GrahamCampbell:
NB, you were returning the user id for some reason when the description in the dingo interfaces says you should return an instance of the user.

jbrooksuk:
Thanks buddy.

jbrooksuk:
... Like an idiot I was trying something and committed that too. Will fix when I continue work.

GrahamCampbell:
> have I registered the service provider properly here?

yeh, looks good

joecohens:
:+1:

jbrooksuk:
@GrahamCampbell do you think a `post-update-cmd` as a Laravel command would be good enough, so we can update all existing Heroku deployments who don't have an API key?

jbrooksuk:
I've added a button to the user panel to regenerate the API key anyway, so perhaps we advise that.

jbrooksuk:
Will merge for now, with a note for generating API keys.

jbrooksuk:
I've merged it, but now it's gone from the main code?

GrahamCampbell:
Errrr, what>

GrahamCampbell:
https://github.com/cachethq/Cachet/commits/master

GrahamCampbell:
I can still see it?

jbrooksuk:
I just re-merged it. It wasn't in `master`.

GrahamCampbell:
Grrr. Is this something to do with your git setup. This has happened before?

jbrooksuk:
I think it must be. I'm going to have to check that out.

jbrooksuk:
Are we missing anything else?

GrahamCampbell:
Not sure. Haven't been watching carefully.

jbrooksuk:
I think it's all there.

PR #282, opened 2015-01-05
Drop gulp-rimraf in favour of del. Closes #276

### LINKED ISSUE #276: gulp-rimraf is deprecated
Deploying to Heroku tells me this:

```
npm WARN deprecated gulp-rimraf@0.1.1: Use npmjs.org/del instead, see https://github.com/gulpjs/gulp/blob/master/docs/recipes/delete-files-folder.md
```

jbrooksuk:


package.json
```diff
@@ -1,8 +1,8 @@
 {
   "dependencies": {
     "bower": "1.3.12",
+    "del": "1.1.1",
```
GrahamCampbell: shouldn't lock dependencies like this

GrahamCampbell: is `~1.1.1` if the project is non-semver, or `^1.1.1` if it is semver.

package.json
```diff
@@ -1,8 +1,8 @@
 {
   "dependencies": {
     "bower": "1.3.12",
```
GrahamCampbell: same here

package.json
jbrooksuk: How come? Want me to use `^1.1.1`

package.json
jbrooksuk: Ok, I'll fix them up.

package.json
GrahamCampbell: It makes no sense to lock it at that version. What's wrong with 1.1.2 when it's released?

jbrooksuk: I've set it to `~1.1.1`.

PR #288, opened 2015-01-05
Allow deleting of a component group. Closes #286

### LINKED ISSUE #286: [Feature Request] Component group deletion
It would be nice to be able to delete component groups.

I'd say, if and only if there is no component associated with a group. However, if a component can be group-less, we could delete any component group, and make its components group-less.

jbrooksuk:
Simple `GET` request to delete.

### DIFF
--- a/app/views/dashboard/components/groups/index.blade.php
+++ b/app/views/dashboard/components/groups/index.blade.php
@@ -23,8 +23,7 @@
                             <strong>{{ $group->name }}</strong>
                         </div>
                         <div class='col-md-4 text-right'>
-                            <a href='#' class='btn btn-default'>{{ trans('forms.edit') }}</a>
-                            <a href='#' class='btn btn-danger'>{{ trans('forms.delete') }}</a>
+                            <a href='/dashboard/components/groups/{{ $group->id }}/delete' class='btn btn-danger confirm-action' data-method='DELETE'>{{ trans('forms.delete') }}</a>
                         </div>
                     </div>
                     @empty
--- a/app/assets/js/start.js
+++ b/app/assets/js/start.js
@@ -1,5 +1,4 @@
 $(function() {
-
     // Ajax Setup
     $.ajaxPrefilter(function(options, originalOptions, jqXHR) {
         var token;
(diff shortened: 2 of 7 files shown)


GrahamCampbell:
This is bad.

GrahamCampbell:
GET must not have an "action". This violates the http spec. We must use `DELETE`.

GrahamCampbell:
http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html

jbrooksuk:
Yeah, we need to tidy up the routes, components and incidents are deleted with a `GET` request too.

GrahamCampbell:
Yeh, I just noticed. Very bad. Note that laravel support pretend DELETE requests.

jbrooksuk:
I didn't know that?

jbrooksuk:
To be honest, the whole `dashboard.php` routes file needs cleaning up.

GrahamCampbell:
> I didn't know that?

One sec, I'll find the code, then explain what it means.

GrahamCampbell:
Ok then, so laravel internally calls `Request::enableHttpMethodParameterOverride();` in https://github.com/laravel/framework/blob/v4.2.16/src/Illuminate/Foundation/start.php#L195 which is this: https://github.com/symfony/symfony/blob/v2.5.8/src/Symfony/Component/HttpFoundation/Request.php#L680.

This means you can send a POST request, and set the `_method` to `DELETE` and we're good.

GrahamCampbell:
See the internals of how the method is determined by symfony: https://github.com/symfony/symfony/blob/v2.5.8/src/Symfony/Component/HttpFoundation/Request.php#L1246.

GrahamCampbell:
I guess this will do for now. We can fix that issue later.

GrahamCampbell:
This needs rebasing.

jbrooksuk:
Done. @GrahamCampbell I'll take a proper look at Laravel faking these `DELETE` requests.

GrahamCampbell:
I actually just use a small bit of javascript to do this in my projects.

GrahamCampbell:
Feel free to lift the code out of my cms. :)

jbrooksuk:
Sweet, thanks. I'll have a look now.

GrahamCampbell:
https://github.com/BootstrapCMS/CMS/blob/v0.8.2-alpha/app/assets/js/cms-restfulizer.js
https://github.com/BootstrapCMS/CMS/blob/v0.8.2-alpha/app/views/events/delete.blade.php#L13

jbrooksuk:
Looks easy enough :)

jbrooksuk:
@GrahamCampbell mocked! Your code is basically unchanged too. Once this is merged, I'll make all of the delete routes work the same.

app/views/dashboard/components/groups/index.blade.php
```diff
@@ -24,7 +24,7 @@
                         </div>
                         <div class='col-md-4 text-right'>
                             <a href='#' class='btn btn-default'>{{ trans('forms.edit') }}</a>
```
GrahamCampbell: should we not just remove this button?

jbrooksuk: Yeah we should for now.

GrahamCampbell:
> Once this is merged, I'll make all of the delete routes work the same.

Might be better to just do it here, and rename this pull.

jbrooksuk:
Oh, ok.

jbrooksuk:
Done.

jbrooksuk:
@GrahamCampbell all done. Anything else before I merge?

GrahamCampbell:
Looks good. :)

jbrooksuk:
Whoop.

PR #294, opened 2015-01-05
[BACKWARDS INCOMPATIBLE] Redo migrations

jbrooksuk:
I've rewritten all existing migrations to merge into one migration per table. I have however removed the web hook tables and models, we'll come back to that properly when we're ready to work on them.

**Please note that all existing installations will be unable to migrate past this point. This is a backwards incompatible change.**

We've been experiencing issues when upgrading PgSQL databases which prevent components from being added without groups. This change will fix the migrations.


app/database/migrations/2015_01_05_203014_CreateSettingsTable.php
```diff
@@ -16,7 +16,7 @@ public function up()
         Schema::create('settings', function (Blueprint $table) {
             $table->increments('id');
             $table->string('name');
-            $table->text('value')->nullable()->default(null);
+            $table->longText('value');
```
joecohens: Should this have a default?

jbrooksuk: Nah, settings will always have a value, won't they?

GrahamCampbell: We shouldn't abuse default values.

joecohens: Ok ok. :)

joecohens:
:+1:

jbrooksuk:
Rebased.

jbrooksuk:
@GrahamCampbell ok to merge?

app/database/migrations/2015_01_05_203341_CreateUsersTable.php
```diff
@@ -15,10 +15,11 @@ public function up()
     {
         Schema::create('users', function (Blueprint $table) {
             $table->increments('id');
-            $table->string('username', 200)->nullable(false);
-            $table->string('password')->nullable(false);
-            $table->rememberToken();
-            $table->string('email')->nullable()->default(null);
+            $table->string('username');
+            $table->string('password');
+            $table->string('remember_token')->default('');
```
GrahamCampbell: this should be nullable

jbrooksuk: Still with a default of `''`?

GrahamCampbell: also should be unique

GrahamCampbell: we don't want empty string when there is no token, we surely want null?

app/database/migrations/2015_01_05_203341_CreateUsersTable.php
```diff
@@ -15,10 +15,11 @@ public function up()
     {
         Schema::create('users', function (Blueprint $table) {
             $table->increments('id');
-            $table->string('username', 200)->nullable(false);
-            $table->string('password')->nullable(false);
-            $table->rememberToken();
-            $table->string('email')->nullable()->default(null);
+            $table->string('username');
```
GrahamCampbell: should be unique

app/database/migrations/2015_01_05_203341_CreateUsersTable.php
jbrooksuk: Good point.

app/database/migrations/2015_01_05_203341_CreateUsersTable.php
jbrooksuk: We want email to be unique as well then since that's what we login with.

jbrooksuk: And api_key

GrahamCampbell: yeh

GrahamCampbell: or better, an index

GrahamCampbell: rather than just unique

joecohens: :+1:

jbrooksuk: A unique index?

GrahamCampbell: indexes have to be unique don't they?

jbrooksuk: http://laravel.com/docs/4.2/schema#adding-indexes

jbrooksuk: Done. Check now.

app/database/migrations/2015_01_05_203341_CreateUsersTable.php
```diff
@@ -15,10 +15,11 @@ public function up()
     {
         Schema::create('users', function (Blueprint $table) {
             $table->increments('id');
-            $table->string('username', 200)->nullable(false);
-            $table->string('password')->nullable(false);
-            $table->rememberToken();
-            $table->string('email')->nullable()->default(null);
+            $table->string('username')->unique();
```
GrahamCampbell: we should just add this an "index" shouldn't we?

jbrooksuk: Index the username as `unique`?

jbrooksuk: That's fine if it's the case, so I assume it's the same with the rememberToken().

jbrooksuk:
@GrahamCampbell ok, I've switched back to `rememberToken` I forgot about that.

GrahamCampbell:
This looks ok. We can always just edit the migrations at a later point and cause more rage, lol.

jbrooksuk:
Prepare your bunkers lads, we're merging.

GrahamCampbell:
lol, i get plenty of abuse whenever i close more than 10 issues to do with eloquent over on laravel/framework anyway. I closed 100 the other week. Big mess, lol.

jbrooksuk:
:hammer:

joecohens:
Haha :)

PR #301, opened 2015-01-06
Link incidents to components

jbrooksuk:
Linking an incident to a component has no effect on the component status yet.

### DIFF
--- a/app/lang/en/forms.php
+++ b/app/lang/en/forms.php
@@ -27,6 +27,7 @@ return [
     'incidents' => [
         'name'         => 'Name',
         'status'       => 'Status',
+        'component'    => 'Component',
         'message'      => 'Message',
         'message-help' => 'You may also use Markdown.',
 
@@ -98,5 +99,8 @@ return [
     'delete' => 'Delete',
     'submit' => 'Submit',
     'cancel' => 'Cancel',
-    'remove' => 'Remove'
+    'remove' => 'Remove',
+
+    // Other
+    'optional' => '* Optional',
 ];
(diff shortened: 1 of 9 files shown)


willdurand:
:+1:

jbrooksuk:
Updating a components status isn't easy (pragmatically it is) but it's the mapping between an incident status and a component status that isn't.

willdurand:
Maybe, you should leave this as is, and only mention that an incident is tied to a component in the UI, because both can be decorrelate IMO.

jbrooksuk:
Yeah I think that Cachet should link the statuses, but what we could do is this:

1. User selects a component from the dropdown.
2. Cachet displays the inline status editor.
3. On save, the incident gets created and the component status will update too.

If no status is chosen from the inline status picker, nothing happens to it.

Thoughts?

willdurand:
Sounds good to me. Cachet should let users:

* create incidents without any components linked
* change component statuses without having to create a new incident nor link to an existing incident

jbrooksuk:
I'm glad you agree. As far as I'm concerned, Cachet shouldn't do anything unexpected.

In which case, I'm going to leave this PR open until I've had time to implement the inline status editor within the incident pages.

app/lang/en/forms.php
```diff
@@ -21,6 +21,7 @@
     'incidents' => [
         'name'         => 'Name',
         'status'       => 'Status',
+        'component'    => 'Component',
```
GrahamCampbell: missing from fr

app/lang/en/forms.php
```diff
@@ -86,5 +87,8 @@
     'delete' => 'Delete',
     'submit' => 'Submit',
     'cancel' => 'Cancel',
-    'remove' => 'Remove'
+    'remove' => 'Remove',
+
+    // Other
+    'optional' => '* Optional',
```
GrahamCampbell: missing from fr

PR #310, opened 2015-01-06
French translation

foxted:
Quick translations & sync check.

### DIFF
--- a/app/lang/fr/setup.php
+++ b/app/lang/fr/setup.php
@@ -5,7 +5,7 @@ return [
     'title'             => 'Installer Cachet',
     'service_details'   => 'Details du service',
     'status_page_setup' => 'Installation de la page de statut',
-    'show_support'      => 'Show support for Cachet? <small>Places a link in the footer linking to Cachet.</small>',
+    'show_support'      => 'Montrer votre support pour Cachet? <small>Affiche un lien vers Cachet dans le pied de page.</small>',
     'admin_account'     => 'Compte administrateur',
     'complete_setup'    => 'Terminer l\'installation',
     'completed'         => 'Cachet a été configuré avec succès !',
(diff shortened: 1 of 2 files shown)


jbrooksuk:
Thanks!

app/lang/fr/setup.php
```diff
@@ -5,7 +5,7 @@
     'title'             => 'Installer Cachet',
     'service_details'   => 'Details du service',
     'status_page_setup' => 'Installation de la page de statut',
-    'show_support'      => 'Show support for Cachet? <small>Places a link in the footer linking to Cachet.</small>',
+    'show_support'      => 'Montrer votre support pour Cachet? <small>Affiche un lien vers Cachet dans le pied de page.</small>',
     'admin_account'     => 'Compte administrateur',
     'complete_setup'    => 'Terminer l\'installation',
     'completed'         => 'Cachet a été configuré avec succès !',
```
GrahamCampbell: shouldn't be a space before the !

willdurand: Before the `!` nope, but before `?` yes, for sure!

jbrooksuk: Unless that's how the French do grammar ;)

GrahamCampbell: there shouldn't be spaces anywhere.

willdurand: In French, there are spaces before any punctuation :)

GrahamCampbell: Hmmm, it seems to be a stylistic choice. It does seem to be preferred in that language. I think we shouldn't do it though. It spoils the consistency.

foxted:
True, there should be a space before the ?, missed that!

app/lang/fr/setup.php
willdurand: Allow me to disagree, it is not a stylistic choice. See #313

GrahamCampbell: Ok then. I'm far from an expert. :)

Crash--: In French, we have to add a space before every "double ponctuation" (i.e. : ; ! ?) it's the rule, not a stylistic choice.

jbrooksuk: @GrahamCampbell if that's how French write, then we should do it. It should be a natural translation.

PR #316, opened 2015-01-06
Notification system

Crash--:
We can now enable Slack or Twilio notification when an incident is saved.


.gitignore
```diff
@@ -19,3 +19,4 @@ public/js
 # Packages
 npm-debug.log
 .vagrant
+composer.lock
```
foxted: :+1:

PR #324, opened 2015-01-07
Added npm to the apt-get install

iMstr:
npm is not included by default on Ubuntu.


iMstr:
Also added "array-union" which is asked for in gulp but not in the package list.

jbrooksuk:
Have you done a test deploy to Heroku?

iMstr:
Done that, alsi I have not deployed to Heroku but tested on my local installation.

jbrooksuk:
Yep - worked for me :)

Issue #281, opened 2015-01-05
Add locale, timezone & url settings to config/app.php after application setup


foxted:
I think it would be nice to modify the `config/app.php` file at setup to save the **locale**, **timezone** and **url (domain)** settings.

I think it will just make it faster to install with the desired language & timezone.

jbrooksuk:
This is definitely something we need to address. We actually ask for the `app_domain` at setup, but never use it in `app.php`.

foxted:
How would you implement this?

jbrooksuk:
We'd have to have default settings with it, so that before it's installed, migrations etc will continue to work.

foxted:
I have been playing around with a package I found: [Phil-F/Setting](https://github.com/Phil-F/Setting). It seems pretty cool, you can set/get/forget settings in a json file wherever you want, and it has a fallback on the Laravel default if you use the same path (e.g. `app.locale`, `app.timezone`, etc...). 

Just need to be careful as his facade is called `Setting` and we already have a model with this name (which is setting the App name, we could store that info in json as well instead of using the database?!) 

What do you think?

jbrooksuk:
@foxted the problem with that package is that if you were to redeploy the app on Heroku, you'd lose your settings.

foxted:
Except if you change the path on the config file to something inside you versionning.

jbrooksuk:
Potentially you'd lose them if you followed the current upgrade guide.

GrahamCampbell:
@foxted Our current settings system is fine. You edit them online, and they're stored in the db. We should not start pulling in more packages to cachet (ex ones we know will be laravel 5 ready) otherwise I'm going to have to rewrite every package we use that's not L5 compatible when we upgrade...

jbrooksuk:
@GrahamCampbell is there any reason we can't replace these values in `app.php` with `Setting::get()` so long as the model can take a default value? Also, as a side question, `Setting` should be a repository right?

GrahamCampbell:
No. Repositories are totally overused in laravel. They make no sense because we're only using eloquent. The whole point of repositories is to allow you to interact with multiple totally different totally different database systems. I think we should remove our repositories. All the code in them belongs in our models.

jbrooksuk:
Ok, so the models are fine as they are? I know you can use repositories for changing to NoSQL and mocking, I was just wondering if it's also the kind of thing we should use for default with default values. If not, that's cool.

foxted:
[...]

Issue #192, opened 2015-01-01
Show exact incident time on status page (on hover/click)


Ehesp:
If a service goes down and effects a 3rd party, they might need to do a report if things effected their services with specific outage times.

jbrooksuk:
We could also add [timeago](http://timeago.yarp.com/) so that the readable times are updated in realtime.

GrahamCampbell:
Yeh, I use that library on bootstrap cms. I also have an html macro setup that I use with it.

jbrooksuk:
That'd be good to add!

GrahamCampbell:
```php
        HTML::macro('ago', function (Carbon $carbon, $id = null) {
            if ($id) {
                return '<abbr id="'.$id.'" class="timeago" title="'
                    .$carbon->toISO8601String().'">'.$carbon->toDateTimeString().'</abbr>';
            } else {
                return '<abbr class="timeago" title="'
                    .$carbon->toISO8601String().'">'.$carbon->toDateTimeString().'</abbr>';
            }
        });
    }
```

jbrooksuk:
I'll create a `macros.php` and whack it in there.

GrahamCampbell:
NB, I actually did this in a service provider, and didn't use the facade.

GrahamCampbell:
Real code: https://github.com/GrahamCampbell/Laravel-Core/blob/master/src/CoreServiceProvider.php#L63.

jbrooksuk:
@GrahamCampbell we should implement that into Cachet - it'd be a nice little "polish" feature IMHO.

PR #325, opened 2015-01-07
Set timezone and date from setup and settings

joecohens:
Closes https://github.com/cachethq/Cachet/issues/281

After this we need a presenter that use the datetime package. The lang thing it's automatic.

### DIFF
--- a/app/config/app.php
+++ b/app/config/app.php
@@ -127,6 +127,7 @@ return [
         'GrahamCampbell\Markdown\MarkdownServiceProvider',
         'Roumen\Feed\FeedServiceProvider',
         'Thujohn\Rss\RssServiceProvider',
+        'Jenssegers\Date\DateServiceProvider',
 
         'CachetHQ\Cachet\Providers\AuthServiceProvider',
         'CachetHQ\Cachet\Providers\ConsoleServiceProvider',
--- a/app/lang/en/forms.php
+++ b/app/lang/en/forms.php
@@ -4,11 +4,13 @@ return [
 
     // Setup form fields
     'setup' => [
-        'email'       => 'Email',
-        'username'    => 'Username',
-        'password'    => 'Password',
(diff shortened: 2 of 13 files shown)


joecohens:
This works also with current installations :)

jbrooksuk:
@joecohens we should implement #192 at the same time as your suggestion.

joecohens:
@jbrooksuk Sure! I like https://github.com/laracasts/Presenter they are really extensible and easy, but I think @GrahamCampbell has also one for auto presenters.

GrahamCampbell:
Yeh, use mine, lol. We actually have 2 maintainers. It's Shawn's pacakge originally. :)

joecohens:
@GrahamCampbell Lol

GrahamCampbell:
Ours support dependency injection into presenters. :P

jbrooksuk:
I need to read about presenters.

jbrooksuk:
So presenters are away of performing logic on a view, before the data goes into the view?

joecohens:
@jbrooksuk Correct.

joecohens:
@GrahamCampbell Just sent a PR to don't use facades since we don't have them registered https://github.com/ShawnMcCool/laravel-auto-presenter/pull/61 help me merge it haha :)

jbrooksuk:
Ok, so an example is that we'd use a Presenter for formatting the timestamps on the incident list?

joecohens:
Yes that would be an example.

jbrooksuk:
Cool! Then yeah, it's absolutely something we should be doing then :)

GrahamCampbell:
I use it for that. NB, we should be using the same api presenters in our api layer instead of duplicating code in transformers.

joecohens:
Definitely!

jbrooksuk:
Makes sense to me. One API format to rule them all.

PR #332, opened 2015-01-08
Incident form is now prefill with the originally data

### LINKED ISSUE #330: Incident form error should prefil the originally submitted data
If you submit an incident but leave off the status, an error is shown and all form fields are emptied.

Crash--:
Close #330

### DIFF
--- a/app/views/dashboard/incidents/add.blade.php
+++ b/app/views/dashboard/incidents/add.blade.php
@@ -27,7 +27,8 @@
                     <fieldset>
                         <div class="form-group">
                             <label for="incident-name">{{ trans('forms.incidents.name') }}</label>
-                            <input type="text" class="form-control" name="incident[name]" id="incident-name" required />
+                            <input type="text" class="form-control" name="incident[name]" id="incident-name" required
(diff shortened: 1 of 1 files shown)


jbrooksuk:
I think you could actually use `Input::old()` here.

Crash--:
As we already store the previous $incident in the Session and we already do a check $incident->isValid() what will be the benefice of doing that with Input::old()? (it's really a question, not a sarcasm or something else)

jbrooksuk:
Actually, we should be using `Binput::old` which is @GrahamCampbell's safe version of it.

What it does is use the `Session::flash` code to maintain the form content. We probably could just use `$incident` but I've always preferred that.

jbrooksuk:
@joecohens is this an ok way of doing things?

@Crash-- this needs rebasing.

joecohens:
I don't think so, this should be `Binput::old('incident.name', $incident->name)`

jbrooksuk:
Thought so :)

jbrooksuk:
@Crash-- can you rebase and use `Input::old` please (`Binput` isn't aliased, so that won't work).

jbrooksuk:
Closed by c3e8993.

Issue #319, opened 2015-01-07
MySQL 5.6 compatibility


wdalmut:
I don't know if mysql 5.6 is in the compatibility list but with that version of mysql during migrations i receive that error

```
Migration table created successfully.
                                                                                                                                                                                              
  [Illuminate\Database\QueryException]                                                                                                                                                         
  SQLSTATE[42000]: Syntax error or access violation: 1101 BLOB/TEXT column 'description' can't have a default value (SQL: create table `components` (`id` int unsigned not null auto_incremen  
  t primary key, `name` varchar(255) not null, `description` text not null default '', `link` text not null default '', `tags` text not null default '', `status` int not null, `order` int n  
  ot null, `group_id` int not null, `user_id` int not null, `created_at` timestamp default 0 not null, `updated_at` timestamp default 0 not null, `deleted_at` timestamp null) default charac  
  ter set utf8 collate utf8_unicode_ci)                                                         
  [PDOException]                                                                                                     
  SQLSTATE[42000]: Syntax error or access violation: 1101 BLOB/TEXT column 'description' can't have a default value  
                                                                   
migrate [--bench[="..."]] [--database[="..."]] [--force] [--path[="..."]] [--package[="..."]] [--pretend] [--seed]
```

With mysql 5.5 all works correctly.

I tried with docker pulling the mysql:5.6 repository

```
docker run --name cachet-mysql -p 3306:3306 -e MYSQL_ROOT_PASSWORD=root -d mysql:5.6
```

jbrooksuk:
Can you try running the latest version 873f726671473bf1e06f31ca7c65cdc0bf2c927b please.

FWIW I run MySQL 5.6 locally for development and have had no issues. We've recently rewritten all of the migrations too, so that may help.

Crash--:
I'm also running 5.6(.22) locally, and had encounter no issues with migrations today.

wdalmut:
I have always the same issue. Here is the whole error:

```
13:02 $ docker run -e MYSQL_ROOT_PASSWORD=root --link cachet-mysql:cachet -it mysql:5.6 mysql -uroot -proot -hcachet                                                                                                                                                           
Warning: Using a password on the command line interface can be insecure.
Welcome to the MySQL monitor.  Commands end with ; or \g.
Your MySQL connection id is 3
Server version: 5.6.22 MySQL Community Server (GPL)

Copyright (c) 2000, 2014, Oracle and/or its affiliates. All rights reserved.

[...]

PR #334, opened 2015-01-08
Fixes 319 - MySQL 5.6 compatibility

wdalmut:
Removed the default field during migrations for text fields - Issue #319

### DIFF
--- a/app/database/migrations/2015_01_05_201444_CreateComponentsTable.php
+++ b/app/database/migrations/2015_01_05_201444_CreateComponentsTable.php
@@ -16,9 +16,9 @@ class CreateComponentsTable extends Migration
         Schema::create('components', function (Blueprint $table) {
             $table->increments('id');
             $table->string('name');
-            $table->text('description')->default('');
-            $table->text('link')->default('');
-            $table->text('tags')->default('');
+            $table->text('description');
+            $table->text('link');
+            $table->text('tags');
             $table->integer('status');
             $table->integer('order');
             $table->integer('group_id');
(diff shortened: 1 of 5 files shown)


jbrooksuk:
If you run this migration, then try and add an incident without a message, what happens?

wdalmut:
It reply with that the message is a mandatory field.

```
Whoops. Something went wrong with the incident. {"message":["The message field is required."]}
```

add an incident dashboard cachet

jbrooksuk:
And if you do a component without a description?

jbrooksuk:
@wdalmut rather than making the fields required, you can make use of the `$attribute` property within the models, see [here](https://github.com/cachethq/Cachet/blob/master/src/Models/Component.php#L63) for existing attributes.

You'd need to add all of the fields for where you've removed the `default` value from.

wdalmut:
Components allows also empty description.

jbrooksuk:
Ok, so long as nothing has broken, then it should be ok. This will only work for newer installations mind.

wdalmut:
:+1:

wdalmut:
I have squashed my commits

jbrooksuk:
Thanks mate.

GrahamCampbell:
Also, our messages aren't too helpful. The end user should not be seeing json, lol.

joecohens:
Lol yeah that's a thing we must be fixing :)

jbrooksuk:
Yeah, I keep meaning to make the error messages apply to the input that is a problem.

joecohens:
Maybe stop using the Session to display the errors and use Laravel $errors with a partial?

jbrooksuk:
Create an issue :)

PR #339, opened 2015-01-08
Added a ssl fix for /auth/login

noahbass:
added a new filter: SslFixFilter
added SslFixFilter to auth route
added SslFixFilter to RoutingServiceProvider

fixes ssl on the auth login page (https://status.cachethq.io/auth/login) by simply replacing `action="http://*"` with `action="//*"`. useful because cachethq.io uses cloudflare ssl + heroku.


jbrooksuk:
Hey @noahbass thanks for this.

I'd say rather than wasting CPU and replacing the output, why not just fix the asset URL itself?

noahbass:
firstly, I agree on the cpu issue.

secondly, its not the asset url that is the issue, its the url in the form action.

the current login page will throw the auth login page:

mixedcontentwarning-production

or

mixedcontentwarning-production-ff

[because of a password field being in a form where the form action and/or the page is not https](https://developer.mozilla.org/en-US/docs/Security/InsecurePasswords)

Now we could just do this:

```php
{{ Form::open(['url' => secure_url('auth/login')]) }}
```

this would work on a `https://` site, but not a `http://` site.


The next best would be to change line 9 in [`auth/login.blade.php`](https://github.com/cachethq/Cachet/blob/master/app/views/auth/login.blade.php#L9) to use a relative url:

```php
{{ Form::open(['url' => '/auth/login']) }}
```

jbrooksuk:
Last option please. We're in a position to use it on sub domains.

jbrooksuk:
Out of interest, can't you route the form to `auth.login`?

noahbass:
I was thinking the same thing, but unfortunately:

What we want is something like:

```html
<form action="/auth/login">
```

or

```html
<form action="//<domain>/auth/login">
```

But..

```php
{{ Form::open(['route' => 'auth.login']) }}
```

throws an error. but good news! this sort of works:

```php
{{ Form::open(['route' => 'login']) }}
```

But this does not work because it outputs the entire url (ex: http://cachet.io.dev/auth/login)

jbrooksuk:
Ok, just go with `/auth/login` :)

noahbass:
cool.

Whats going on with this Scrutinizer check?

jbrooksuk:
It just checks for various common bad things etc. We good to merge?

noahbass:
#shipit

jbrooksuk:
Cheers!

app/views/auth/login.blade.php
```diff
@@ -6,7 +6,7 @@
             <div class="welcome-logo">
                 <img class="logo" height="50" src="{{ url('img/cachet-logo.svg') }}" alt="Cachet"/>
             </div>
-            {{ Form::open() }}
+            {{ Form::open(['url' => '/auth/login']) }}
```
GrahamCampbell: We should avoid coding urls like this because it breaks everything when people don't run cachet on the root domain.

We should always call the url helper method.

// cc @cachethq/owners

joecohens: I prefer using named routes, if we have them defined I think it's handy for changes on the URI's also.

GrahamCampbell: Yeh, named routes are best. I'm also saying this for things like linking to images too.

noahbass:
@GrahamCampbell 100% agree on url helper

But using

'route' => 'login'

Will give the absolute url in http://* which is fine for http sites but not https sites (and vice versa). 

The best solution would be to have urls in the //* format so they work in http and https sites. Is there a helper that will do this?

Issue #326, opened 2015-01-07
Multi-Factor Authentication


benwa:
Support for MFA would be awesome.  According to [Authy](https://authy.com/developers), it's doable very quickly if you use their REST endpoint, though they also have [libraries](https://github.com/authy) if that would be more appropriate.

jbrooksuk:
Yeah, this is something we could add quite easily. It'd be opt in of course.

nodesocket:
Is you use a service such as Authy that would require each person who wants to run Cachet to signup for them. It would be better to use a fully native Google Authenticator library. Unfortunately I haven't been able to find any decent ones in php. I'm really tempted to use a node.js project (https://github.com/markbao/speakeasy) which seems solid and renders the QR codes.

kfrankperson:
I would like to see 2FA for Cachet but I think it would need to be Google's 2FA as @nodesocket has said.

nodesocket:
FYI, we [Commando.io](https://commando.io) currently use [GAuthify](https://www.gauthify.com) but we also looking for solid php native library without relying on a 3rd party service.

jbrooksuk:
I was looking yesterday but everything returns Authy. Is there really no library that handles it?

nodesocket:
Most libraries are old and seem abandoned and not maintained:

https://github.com/PHPGangsta/GoogleAuthenticator
https://github.com/chregu/GoogleAuthenticator.php

Of course there are solid ruby gems, but not finding anything good for php:

https://github.com/jaredonline/google-authenticator

jbrooksuk:
I see another project coming on... /s

joecohens:
:+1:

nodesocket:
@jbrooksuk we'd be willing to sponsor some development of a **solid** and **secure** php implementation of Google Authenticator including generating QR codes with no 3rd party dependencies besides the call to google images for the QR. How long did you think it would take?

jbrooksuk:
What about [this](https://github.com/scheb/two-factor-bundle) package?

jbrooksuk:
Actually, [this](https://github.com/antonioribeiro/google2fa).

nodesocket:
Winner, winner, chicken dinner.

https://github.com/antonioribeiro/google2fa

Looks solid! Thanks, will play around with it.

jbrooksuk:
Saves re-inventing the wheel and you guys some dollar :) we could definitely implement it into Cachet too.

[...]

PR #342, opened 2015-01-09
[WIP] Two factor authentication

jbrooksuk:
This is a biggy for #326 and is a massive WIP.

- [x] Add package.
- [x] Setup Google 2FA Secret on setup.
- [x] Show QR code in user panel.
- [x] Enable/disable 2FA in user panel.
- [x] If user logs in and has 2FA enabled, show code form (currently working on).
- [x] Language file updates.


jbrooksuk:
Phew. Two factor authentication is implemented. Can a couple of you try it out please? I've been able to log in normally without two-factor turned on and if I turn it on, I can only login by supplying a token.

@cachethq/owners FYI.

joecohens:
@jbrooksuk Testing now ...

joecohens:
@jbrooksuk It's working nice work! But I think the flow is not the best because you can get locked out easily.

I'll remove the enable two factor auth from the setup page and leave it only on the profile. If we leave the button we have to redirect after login to setup MFA code directly. I think some people wont know what to do after checking the box. 

Also when you go to the teams page and you see yourself we can setup a button like edit your profile or something to remember people how to.

What do you think?

app/views/dashboard/user/index.blade.php
```diff
@@ -42,6 +42,27 @@
...
                         </div>
+                        <hr />
+                        <div class="form-group">
+                            <label class="checkbox-inline">
+                                <input type="hidden" name="google2fa" value="0" />
+                                <input type='checkbox' name="google2fa" value="1" {{ !empty(Auth::user()->google_2fa_secret) ? "checked" : "" }} />
+                                {{ trans('forms.setup.enable_google2fa') }}
+                            </label>
+                        </div>
+                        @if(Auth::user()->google_2fa_secret)
```
joecohens: Can we make this validation on the user like:

Model\User
````
function hasEnabled2FA()
{
    return (bool) (trim($this->google_2fa_secret) !== '');
}
````

Comment (jbrooksuk) [inline app/views/dashboard/user/index.blade.php]: Good idea.

**Comment by jbrooksuk** [comment]
@joecohens good point. If you enable it, but then don't get the QR code you can indeed lock yourself out.

Can you push those suggestions? I'm not able to right now.

**Comment by joecohens** [comment]
@jbrooksuk Sure!

**Comment by jbrooksuk** [comment]
Done. Time to rebase and then merge!

**Comment by jbrooksuk** [comment]
Also we need to move the code for the QR code into the controller. It shouldn't be in the view.

File: app/config/app.php [hunk 2]
```diff
@@ -172,17 +173,17 @@
...
-        'Form'     => 'Illuminate\Support\Facades\Form',
-        'Input'    => 'Illuminate\Support\Facades\Input',
-        'Redirect' => 'Illuminate\Support\Facades\Redirect',
-        'Request'  => 'Illuminate\Support\Facades\Request',
-        'Response' => 'Illuminate\Support\Facades\Response',
-        'Route'    => 'Illuminate\Support\Facades\Route',
-        'Session'  => 'Illuminate\Support\Facades\Session',
-        'Setting'  => 'CachetHQ\Cachet\Models\Setting',
-        'Str'      => 'Illuminate\Support\Str',
+        'App'       => 'Illuminate\Support\Facades\App',
```
GrahamCampbell: this indentation should not have changed.

jbrooksuk: Whoops. I did that when I originally added the new library.

app/views/auth/two-factor-auth.blade.php
```diff
@@ -0,0 +1,31 @@
+@extends('layout.clean')
+
+@section('content')
+    <div class="login">
+        <div class="col-xs-12 col-xs-offset-0 col-sm-6 col-sm-offset-3 col-md-4 col-md-offset-4 text-center">
+            <div class="welcome-logo">
+                <img class="logo" height="50" src="{{ url('img/cachet-logo.svg') }}" alt="Cachet"/>
```
GrahamCampbell: you should use the "asset" helper for assets

app/views/dashboard/team/index.blade.php
```diff
@@ -21,7 +21,7 @@
                 <div class="user-grid">
                     @foreach($teamMembers as $member)
                     <div class="user col-sm-3 col-xs-6">
-                        <a href="/dashboard/team/{{ $member->id }}">
+                        <a href="@if(Auth::user()->id == $member->id) {{ url('dashboard/user') }} @else /dashboard/team/{{ $member->id }}  @endif">
```
GrahamCampbell: too many spaces before the closing endif

app/views/auth/two-factor-auth.blade.php
jbrooksuk: The `asset` url breaks things when you're on HTTPS.

jbrooksuk:
Leaving for @joecohens to fix, I've gotta leave for a bit,

src/Http/Controllers/SetupController.php
```diff
@@ -126,12 +126,11 @@ public function postStep2()
             // Pull the user details out.
             $userDetails = array_pull($postData, 'user');
 
-            // TODO: Do we want to just use Model::unguard() here?
             $user = User::create([
-                'username' => $userDetails['username'],
-                'email'    => $userDetails['email'],
-                'password' => $userDetails['password'],
-                'level'    => 1,
+                'username'          => $userDetails['username'],
```
GrahamCampbell: incorrect indentation

src/Models/User.php
```diff
@@ -133,4 +133,14 @@ public function getIsAdminAttribute()
...
     }
+
+    /**
+     * Returns if a user has enabled two factor authentication.
+     *
+     * @return bool
+     */
+    public function getHasEnabled2FAAttribute()
+    {
+        return (bool) (trim($this->google_2fa_secret) !== '');
```
GrahamCampbell: no need to cast this to a bool

app/views/auth/two-factor-auth.blade.php
GrahamCampbell: That's a bummer. :(

GrahamCampbell: It's also odd. That helper function works for me on styleci and on bootstrap cms, using L4.2 and L5.0.

joecohens: Yeah I saw you where using the `assets` on StyleCI.

joecohens:
This is done :) the only thing missing is the suggestion of @GrahamCampbell to change te `getHasEnabled2FAAttribute` method name.

jbrooksuk:
Does this stop the user password from being changed too?

joecohens:
@jbrooksuk Now it does :)

jbrooksuk:
@joecohens what's left?

jbrooksuk:
So we should be able to login with oAuth, disable/enable it and it will no longer lose the password?

joecohens:
@jbrooksuk Yes!

PR #344, opened 2015-01-09
Updated Dependencies

GrahamCampbell:
I've also changed the version constraints to strictly specify the minor version because I don't trust people not to break things. I think it would be useful for us to have full control over the minor versions of our dependencies.


joecohens:
:+1:

jbrooksuk:
👍

Issue #346, opened 2015-01-10
Heroku install setup fails on Complete Setup Step


Forge-Media:
Hi.

I've used the one-click Heroku install to install Cachethq.io

However, once install finishes successfully and I navigate to the appname.herokuapp.com address to proceed with the setup here I encounter a problem.

Once I hit the Administrator Account and enter the required details and click the Complete Setup, nothing happens. Inspect elements in chrome gives the following error: 

POST http://appname.herokuapp.com/setup/step2 500 (Internal Server Error)all-4073a820.js:3 X.cors.e.crossDomain.sendall-4073a820.js:3 Z.extend.ajaxall-4073a820.js:3 Z.(anonymous function)all-4073a820.js:11 (anonymous function)all-4073a820.js:2 Z.event.dispatchall-4073a820.js:2 m.handle
all-4073a820.js:11 Uncaught TypeError: Cannot read property 'errors' of undefinedall-4073a820.js:11 (anonymous function)all-4073a820.js:2 dall-4073a820.js:2 c.fireWithall-4073a820.js:3 nall-4073a820.js:3 (anonymous function)

PR #347, opened 2015-01-10
Fixes error on postgresql on 2FA column

joecohens:
Fixes https://github.com/cachethq/Cachet/issues/346


GrahamCampbell:
I think this caused issues with mysql 5.6.

joecohens:
I've tested on the latest version of Homestead with MySQL Distrib 5.6.19, for debian-linux-gnu (x86_64) and (PostgreSQL) 9.4.0. Maybe for another dist?

GrahamCampbell:
Also, this shouldn't be a separate migration file anyway. We should just be editing the existing migrations until we actually tag v1.0.0.

joecohens:
Yeah that's true, this is only fixing one click install on Heroku for now.

jbrooksuk:
Cool.

Issue #99, opened 2014-12-28
Steps to launch


jbrooksuk:
Once Cachet is setup, I think that we could display "steps to launch" in the dashboard index, it'll basically walk them through the things they've left to do:

1. Setup Cachet (ticked obviously)
2. Login to dashboard (ticked, because they should be auto logged in).
3. Add components (ticked once there are > 0 components in the DB).
4. Add first incident (ticked once there are > 0 incidents in the DB).
5. Customize Cachet with theme settings.

Then a button to hide the steps.

By adding this kind of feature we're showing the user how to use Cachet in a non-invasive way.

I'm thinking of something like this:

![screenshot 2014-12-28 10 27 55](https://cloud.githubusercontent.com/assets/246103/5563620/7fc96e3e-8e7c-11e4-9b3a-95c5ec1dceb7.png)

joecohens:
See https://github.com/cachethq/Cachet/pull/403

PR #362, opened 2015-01-13
[WIP] Steps to configure

joecohens:
Not sure if it would be better to store checks on settings like `has_setup_components`, `has_setup_incidents`, `has_setup_customize`. So we can set a notification bar at the top to help with the steps your missing.

screen shot 2015-01-13 at 12 55 18 am

Closes https://github.com/cachethq/Cachet/issues/99

### DIFF
--- a/src/Http/Controllers/DashboardController.php
+++ b/src/Http/Controllers/DashboardController.php
@@ -3,6 +3,8 @@
 namespace CachetHQ\Cachet\Http\Controllers;
 
 use CachetHQ\Cachet\Models\Component;
+use CachetHQ\Cachet\Models\Incident;
+use CachetHQ\Cachet\Models\Setting;
 use Illuminate\Routing\Controller;
 use Illuminate\Support\Facades\View;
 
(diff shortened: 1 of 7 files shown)


jbrooksuk:
@joecohens you'll also need to copy the `en` translation to the other three to keep them in sync for now. When we're closer to finishing I'll push everything to be translated.

jbrooksuk:
With that said @joecohens all of the text which is within the views need moving into the `./app/lang/en/` files :)

joecohens:
@jbrooksuk Yeah I haven't  done the translations yet, will do them now. What do you think about the way I'm checking if you have done steps? Or do you like better the proposal I've made on the comment?

src/Http/Controllers/DashboardController.php
```diff
@@ -24,6 +26,101 @@ public function showDashboard()
...
+     * Shows missing configurations view.
+     *
+     * @return \Illuminate\View\View
+     */
+    public function showConfiguration()
+    {
+        $hasComponents = (Component::withTrashed()->count() > 0);
+        $hasIncidents = (Incident::withTrashed()->count() > 0);
+        $hasCustomized = (Setting::get('style_background_color') || Setting::get('style_text_color'));
+
```
joecohens: @cachethq/owners Not sure about this checks what do you think? Maybe we could have a Setting for each check like `has_customized => true`?

jbrooksuk: That's not a bad idea.

jbrooksuk: Also made better by your new setting caching.

jbrooksuk: Also, the `$hasCustomized` could include a custom stylesheet.

jbrooksuk:
@joecohens any update on this? It's so close :)

joecohens:
See https://github.com/cachethq/Cachet/pull/403

PR #367, opened 2015-01-13
Dynamic language list within setup.

### LINKED ISSUE #366: make language-chooser dynamic
It would be nice if new translation would be available wihtout [further interventions](https://github.com/cachethq/Cachet/pull/365) as soon as there is a directory in app/lang/.

This also includes that the languages' name have to be included in the lang-files.

jbrooksuk:
Closes #366.

Nice and easy really. It does mean that we need PECL intl >= 1.0.0 though. Thoughts?

### DIFF
--- a/src/Http/Controllers/SetupController.php
+++ b/src/Http/Controllers/SetupController.php
@@ -34,12 +34,13 @@ class SetupController extends Controller
      */
     public function getIndex()
     {
-        $langs = [
-            'de'    => 'Deutsch',
-            'en'    => 'English',
-            'fr'    => 'Français',
-            'pt-BR' => 'Portuguese, Brazilian',
-        ];
+        $langs = array_map(function ($lang) {
+            $locale = basename($lang);
+
+            return [$locale => ucwords(locale_get_display_name($locale, $locale))];
+        }, glob(app_path('lang').'/*'));
+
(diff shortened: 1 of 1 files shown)


joecohens:
:+1:

joecohens:
We have a problem here, `locale_get_display_name` can only be used if you have the PECL intl extension installed so it should be required or find another way to get the locale name.

joecohens:
Sorry I didn't saw the PR comment it's too late haha :). Maybe we can set a config file with the array mapping, how often are you going to install a new lang once they are done, instead of requiring a hole extension for such a small feature. Thoughts?

jbrooksuk:
Without having a big array somewhere it's really only that function. We could add it to require within composer.json though.

jbrooksuk:
Heh, I missed your second comment.

I guess for adding one language the dynamic language functionality is a bit overkill. Sure, it's a pain, but it's a one-time job.

joecohens:
Haha happens :). Yeah I think it's better than requiring another PHP just for that.

jbrooksuk:
Ok, lets do that then.

joecohens:
Sure! I'm sending a PR to update the SP of config right now, I'll include it there :)

jbrooksuk:
Awesome, sounds good to me!

PR #372, opened 2015-01-14
Create container for Cachet

byxorna:
In order to make it easier for people to try out Cachet, I created a Dockerfile. Hopefully you all find this useful and can set up a trusted build on the docker hub.

@jbrooksuk CC. If you want to play with the image, you can pull down ```byxorna/cachet:latest```. It is the build as of 953ece0.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+node_modules
+.git
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM php:5.4-apache
+FROM php:5.6-apache
 
 # before building
 # run npm install
@@ -40,10 +40,10 @@ RUN docker-php-ext-install zip && \
     docker-php-ext-install mcrypt && \
     docker-php-ext-install mbstring && \
     docker-php-ext-install pdo_mysql
-RUN curl http://pecl.php.net/get/APC-3.1.13.tgz -o /usr/src/php/ext/apc.tar.gz && \
-    tar xzvf /usr/src/php/ext/apc.tar.gz -C /usr/src/php/ext && \
-    mv /usr/src/php/ext/APC-3.1.13 /usr/src/php/ext/apc && \
-    docker-php-ext-install apc
+RUN curl http://pecl.php.net/get/apcu-4.0.7.tgz -o /usr/src/php/ext/apcu.tar.gz && \
+    tar xzvf /usr/src/php/ext/apcu.tar.gz -C /usr/src/php/ext && \
+    mv /usr/src/php/ext/apcu-4.0.7 /usr/src/php/ext/apcu && \
+    docker-php-ext-install apcu
 RUN curl -sS https://getcomposer.org/installer | php
 RUN php composer.phar install --no-dev -o
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,49 @@
+FROM php:5.4-apache
+
+# before building
+# run npm install
+# run node_modules/.bin/bower install
+# run node_modules/.bin/gulp
+
+ENV DB_DRIVER=mysql \
+    ENV=production \
+    DB_DATABASE=cachet \
+    DB_HOST= \
+    DB_USERNAME= \
+    DB_PASSWORD=
+COPY . /var/www/html/
+WORKDIR /var/www/html/
+
+# ensure the assets have been compiled
+RUN for d in public/{build,css,js} ; do test ! -d public/build && \
+  echo "Run 'gulp' before building container" >&2 && exit 1 || : ; done
+
+RUN apt-get update && \
+  apt-get install -y curl libmcrypt-dev zlib1g-dev libxml2-dev \
+  git sqlite libsqlite3-dev --no-install-recommends && \
+  rm -r /var/lib/apt/lists/*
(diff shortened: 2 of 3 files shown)


CaptPhunkosis:
:floppy_disk:

joecohens:
:boom: Awesome!! Do you think we could install an nginx server instead of apache?

joecohens:
And I guess PHP 5.5 :)

byxorna:
@joecohens I used 5.4 as that was what the readme said to use :). Ill update the image to pull in 5.5 in the next commit.

As far as nginx, https://registry.hub.docker.com/u/library/php/ provides only an apache image and FPM (which requires a second container to actually handle requests and send them into the socket, so not very turnkey). I figured using the official PHP image would be preferable to a home-rolled image.

byxorna:
@joecohens also, I saw that in a previous version of the install docs, there was a hard dependency on APC. Is that no longer the case? (I assume so, if you are on 5.5).

GrahamCampbell:
What about 5.6?

> I saw that in a previous version of the install docs, there was a hard dependency on APC. Is that no longer the case?

PHP 5.5+ still has apc though "apcu". We're just using apc as the default cache driver. People can use redis or something if they really want to.

jbrooksuk:
@byxorna it should be >= 5.4 :)

joecohens:
Yeah I think if we are providing people a way to install their own server I would use the latest version of the requirements.

joecohens:
@byxorna I guess we can stick with apache for now, the only other way is to create a custom box with php-fpm and install nignx on it.

joecohens:
https://github.com/fideloper/docker-nginx-php

byxorna:
@joecohens yea there are loads of nginx+php containers out there; I wanted to use the most supportable container I could though, because I trust the official PHP distro more than some random dude on docker hub :). Updated the PR with php-5.6 (test with byxorna/cachet:php-5.6).

jbrooksuk:
:+1:

joecohens:
:+1:

jbrooksuk:
@byxorna are you happy that this is complete?

byxorna:
@jbrooksuk yep, LGTM.

jbrooksuk:
Ok, I don't use Docker, but I'm sure some people will.

byxorna:
@jbrooksuk just realized the readme talks about the tag byxorna/cachet:test, which doesn't exist anymore. This should be updated to whatever your trusted build is, once it is set up. (I.e. cachet/cachet:latest)

jbrooksuk:
As in, we need to tag the release?

byxorna:
You will need to set up an account on dockerhub, and create an automated build hooked up to the cachet git repo, so when this repo changes, a new build is triggered and automatically tagged.

jbrooksuk:
Setting it up now :)

jbrooksuk:
`jbrooksuk/Cachet`.

Can you update the readme please. I signed up to Dockerhub not knowing it'd link to my username (rather, I was thinking it'd use the org name).

byxorna:
You should be able to set up the repo under your organization as well. (I.e. I've got byxorna/collins and tumblr/collins, which are 2 separate repos). If you don't have an org yet, you can create one from your profile (I don't see a cachet org in the hub).

jbrooksuk:
Awesome, I'll try creating an org now.

Out of interest are you planning on using Cachet within Tumblr?

byxorna:
We are experimenting with it. Its a hell of a lot easier to experiment with a container than physically install in a VM or physical machine, thus this PR :)

jbrooksuk:
I've created a Dockerhub org :) `cachethq/cachet` it's building now.

Sweet - let us know if you need anything.

jbrooksuk:
Building gives us this error:

```
The command [/bin/sh -c for d in public/{build,css,js} ; do test ! -d public/build &&   echo "Run 'gulp' before building container" >&2 && exit 1 || : ; done] returned a non-zero code: 1
```

Any advice?

byxorna:
Ach right. When I was doing builds, I ran assets locally before the docker build. This way I wouldn't have to clutter up the dockerfile with nodejs deps that are not needed at build time. I suppose either you can make the cachet repo manually built and you push up new versions manually (you would have to replace the auto build repo with a normal repo I think), or we tweak the dockerfile to pull in nodejs and run assets during the build.

One issue with the assets build is I couldn't figure out how to make gulp run once and exit; it seems to stay running forever until interrupt, which isn't ideal for a container build...

jbrooksuk:
Ok, that makes it less maintainable for us. Are you able to make a PR for this? None of us use Docker, so it's not something we're good at.

byxorna:
@jbrooksuk yea I can look into adding the nodejs deps into the container to do the asset build during build. Can you see if there is a way to make gulp run assets in a oneshot mode? I think this will require a new gulp task to be implemented.

jbrooksuk:
It's strange because the gulp command runs then finishes for me. We do use Elixir, so that could be causing an issue.

byxorna:
@jbrooksuk odd, it finishes for me as well during container build, just not i run it on my laptop. It just hangs out forever after finishing publish. :confused: 

Ill have a PR in shortly with adding asset compilation.

Issue #359, opened 2015-01-12
Finalize V1 of the API


jbrooksuk:
There is still a few endpoints missing from the API:

- [x] Metrics #4 
- [x] Component tags
- [x] Update components
- [x] Created, edit and delete metrics themselves.
- [x] Delete components.

jbrooksuk:
Added component tags.

jbrooksuk:
@GrahamCampbell you mentioned removing of the transformers etc? Is this something we'd benefit from immediately?

GrahamCampbell:
> Is this something we'd benefit from immediately?

No. Removing them would prevent code from diverging. The same effect we were observing with the lang files might be observed here if we're not careful. That was what I was getting at. Only in this case, we really don't need to duplicate this stuff unlike the lang files.

jbrooksuk:
I've made some changes to the Metrics API:

- Endpoints have changed slightly:
    - `GET /api/metrics`
    - `GET /api/metrics/:id`
    - `GET /api/metrics/:id/points`
    - `POST /api/metrics`
    - `POST /api/metrics/:id/points`
- You no longer need to supply the `id` in the JSON body when creating a point.

Other than that, this issue is now done.

PR #373, opened 2015-01-14
Implement API deletes

joecohens:
First step of https://github.com/cachethq/Cachet/issues/359


jbrooksuk:
Looks good! If you fix up StyleCI issues, we can merge.

CC @GrahamCampbell after V1, we should remove the transformers.

PR #376, opened 2015-01-14
Further fixes #375 - Adding Australia to the Dashboard

### LINKED ISSUE #375: Australia not available in timezones
I can't find any options for any cities in Australia for the timezone settings, I selected Bougainvile (I think) which had roughly the same timezone however none of our capital cities (Sydney, Melbourne etc) are available.

Mooash:
Adding Australia to the Dashboard


jbrooksuk:
Whoops, good catch. We should probably make this into a view composer or something re-usable.

joecohens:
:+1:

PR #377, opened 2015-01-14
Make view composers into service providers

jbrooksuk:
Rather than `composers.php` we now have them within our `src/Composers` directory.


joecohens:
:+1:

Issue #371, opened 2015-01-13
After changing language, the success message is in the previous language


jbrooksuk:
1. I changed the language to French and I got **Settings saved.**
2. Then I changed to English and I got **Réglages sauvegardés.**

jbrooksuk:
This could be fixed by @joecohens when he moves the locale stuff to a service provider.

PR #381, opened 2015-01-14
Move load config to ServiceProvider and Lang array.

joecohens:
Closes https://github.com/cachethq/Cachet/issues/371

### DIFF
--- a/src/Composers/TimezoneLocaleComposer.php
+++ b/src/Composers/TimezoneLocaleComposer.php
@@ -4,6 +4,7 @@ namespace CachetHQ\Cachet\Composers;
 
 use DateTime;
 use DateTimeZone;
+use Illuminate\Support\Facades\Config;
 use Illuminate\Support\Facades\View;
 
 class TimezoneLocaleComposer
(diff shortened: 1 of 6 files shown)


src/Providers/LoadConfigServiceProvider.php
```diff
@@ -28,4 +26,14 @@ public function filter(Request $request)
         Config::set('app.locale', $appLocale ?: Config::get('app.locale'));
```
GrahamCampbell: should be `$this->app->config->set(...)`

src/Providers/LoadConfigServiceProvider.php
```diff
@@ -28,4 +26,14 @@ public function filter(Request $request)
...
         Lang::setLocale($appLocale);
```
GrahamCampbell: again, the the app property instead of the facade

joecohens: Right :)

jbrooksuk:
After you've made the changes @GrahamCampbell said, merge :)

src/Composers/TimezoneLocaleComposer.php
```diff
@@ -17,10 +18,12 @@ class TimezoneLocaleComposer
      */
     public function compose(\Illuminate\View\View $view)
     {
-        $langs = array_map(function ($lang) {
+        $enabledLangs = Config::get('langs');
```
GrahamCampbell: NB, we should generally try to avoid regretting Config:: and Setting:: within our code. The view composer shouldn't know about how we store the config.

jbrooksuk: It's within a view composer, is that just as bad?

GrahamCampbell: Yeh. I only do it in service providers. I consider anywhere else to be bad.

PR #382, opened 2015-01-14
Enhance settings performance queries

joecohens:
Changed the way we get Settings by using a service to query only once the settings a cache them to a variable so each time we call Setting::get('app_domain') and Setting::get('app_url') we only query the DB once. Little performance boost :)

// cc @cachethq/owners


jbrooksuk:
This is a good idea! A lot of the settings will likely not change. Is it worth throwing them into apcu whilst we're at it?

jbrooksuk:
Ready to merge?

joecohens:
Just rebased should we merge?

jbrooksuk:
I'm happy :)

GrahamCampbell:
This is a misuse of a service class imo. This would be good for a "caching repository" or something.

jbrooksuk:
Is it work merging now, to rewrite as a repository?

GrahamCampbell:
That said, for now, we could just rename it to something else given that repositories are overkill.

GrahamCampbell:
IE, this class is badly named, rather than actually needs rewriting.

joecohens:
Yeah I can move it to the Repos folder

GrahamCampbell:
We can call it "ConfigRepository". The code doesn't need to change. Don't move it to the repositories namespace please.

GrahamCampbell:
We don't need to huddle all the repositories on one place and try to make them the same. That's not really the idea of repositories. NB, laravel's internal config implementation uses a "config repository".

joecohens:
Should I leave it on the Services folder?

GrahamCampbell:
We don't really need a services namespace. Hmmm...

GrahamCampbell:
Maybe create a "Config" namespace. And simply call the class "Repository"?

joecohens:
That sounds good

GrahamCampbell:
We can still alias it as "Setting" in the global aliases if you like. :)

joecohens:
Yeah I think it makes the difference between Laravel "Config" and app settings

joecohens:
I rebased this and apply a StyleCI patch on Proxies list :), this is ready to merge.

jbrooksuk:
I'm happy :) merge!

PR #390, opened 2015-01-15
added date format help link to settings

noahbass:
untitled

links to http://php.net/manual/en/function.date.php so users can figure out how to format their dates if they want to change it

### DIFF
--- a/app/views/dashboard/settings/app-setup.blade.php
+++ b/app/views/dashboard/settings/app-setup.blade.php
@@ -66,7 +66,10 @@
                             <div class="row">
                                 <div class="col-xs-12">
                                     <div class="form-group">
-                                        <label>{{ trans('forms.settings.app-setup.date-format') }}</label>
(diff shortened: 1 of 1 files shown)


jbrooksuk:
Thanks @noahbass!

app/views/dashboard/settings/app-setup.blade.php
```diff
@@ -66,7 +66,10 @@
                             <div class="row">
                                 <div class="col-xs-12">
                                     <div class="form-group">
-                                        <label>{{ trans('forms.settings.app-setup.date-format') }}</label>
+                                        <label>
+                                            {{ trans('forms.settings.app-setup.date-format') }}
+                                            <a href="http://php.net/manual/en/function.date.php"><i class="icon ion-help-circled"></i></a>
```
GrahamCampbell: should use https here

PR #392, opened 2015-01-16
Rewrite the tags implementation. Closes #391

### LINKED ISSUE #391: Rewrite component tags
Currently the tag system is completely unusable and I'm an idiot for not thinking about this. To have the tags become useful, we need a `tags` table and then a subsequent `component_tags` table that links each component to a tag.

We probably need to pull this feature out for now because it's absolutely unusable and will serve no purpose in the future.

jbrooksuk:
1. I love the `sync` method.
2. It's completely replaces the existing tags implementation.


jbrooksuk:
This will mean we can now use Selectize.js or whatever to suggest them to the user later on.

jbrooksuk:
Once merged, the component API will need updating. I'm wondering how best to send through tags in the API request?

src/Http/Controllers/DashComponentController.php
```diff
@@ -116,6 +118,19 @@ public function updateComponentAction(Component $component)
                 ->with('errors', $component->getErrors());
         }
 
+        // The component was added successfully, so now let's deal with the tags.
+        $tags = str_replace(', ', ',', $tags); // Clean up.
+        $tags = explode(',', $tags);
```
joecohens: This two lines can be replaced with `$tags = preg_split('/ ?, ?/', $string);`

src/Http/Controllers/DashComponentController.php
```diff
@@ -160,6 +178,19 @@ public function createComponentAction()
...
```
joecohens: Same here :)

src/Http/Controllers/DashComponentController.php
jbrooksuk: Would you be ok to do that? I've got to nip out :(

joecohens: Sure NP :)

app/database/migrations/2015_01_16_084806_AlterTableComponentsDropTagsColumn.php
```diff
@@ -0,0 +1,32 @@
+<?php
```
GrahamCampbell: this file should not exist. Just edit the existing migrations.

src/Models/Tag.php
```diff
@@ -0,0 +1,43 @@
...
+     * Overrides the models boot method.
+     *
+     * @return void
+     */
+    public static function boot()
+    {
+        parent::boot();
+
+        self::creating(function ($tag) {
+            $tag->slug = Str::slug($tag->name);
```
GrahamCampbell: This is bad. Slugs should actually be stored in the db. The laravel "slug" function may actually change, so cannot be relied upon.

app/database/migrations/2015_01_16_084806_AlterTableComponentsDropTagsColumn.php
jbrooksuk: Existing installations?

src/Models/Tag.php
jbrooksuk: The slug is stored in the DB. This is setting the value to it. If the API changes how a slug is made then we will write our own. A slug is a slug though.

app/database/migrations/2015_01_16_084806_AlterTableComponentsDropTagsColumn.php
GrahamCampbell: Yeh, we shouldn't worry about them till we actually release 1.0. That way our migrations aren't a mess before we've even released.

src/Models/Tag.php
GrahamCampbell: This shouldn't be done on boot though.

jbrooksuk: I always thought it was a clever solution :smile: where then?

GrahamCampbell: > If the API changes how a slug is made then we will write our own.

I'm saying this because I think it will change. It's been discussed before.

jbrooksuk: A slug is a slug though?

GrahamCampbell: No, there was debate about what do with punctuation and multiple spaces/tabs etc.

jbrooksuk: Shame, we've always used slugs with alpha, spaces turned into dashes, everything else striped.

GrahamCampbell: This is probably fine for now. I'll let you know if this changes in Laravel. I'll try to make sure it goes into 5.1 rather than a 5.0 patch release if it is changed.

jbrooksuk: Thanks buddy :)

app/database/migrations/2015_01_16_084806_AlterTableComponentsDropTagsColumn.php
jbrooksuk: Ok, will fix this later - working on client stuff :)

jbrooksuk:
All done. Merging.

app/views/dashboard/components/add.blade.php
```diff
@@ -50,7 +50,7 @@
                         </div>
                         <div class="form-group">
                             <label>{{ trans('forms.components.tags') }}</label>
-                            <textarea name="component[tags]" class="form-control" rows="2"></textarea>
+                            <input name="component[tags]" class="form-control" />
```
GrahamCampbell: shouldn't be a `/`

jbrooksuk: We need to do that on a lot of pages. I'll do it now.

GrahamCampbell: I changed quite a few a week ago.

jbrooksuk: It's most likely me, I've **always** ended inputs like that.

Issue #394, opened 2015-01-16
Non-Heroku install is failing setup at step2


neomelonas:
In the same vein as #346, I am getting 500s on POST for /setup/step2.

Using sqlite and nginx, setup fails when you try to Complete Setup after entering administrator login info.

I'm on 65d0e4f5f15c256bed51241541a74f7d7931f5db

joecohens:
Thanks!

neomelonas:
Still receiving a 500 on POST from step2.
```
POST http://.../setup/step2 500 (Internal Server Error)     all-92eb510e.js:3
 X.cors.e.crossDomain.send                                  all-92eb510e.js:3
 Z.extend.ajax                                              all-92eb510e.js:3
 Z.(anonymous function)                                     all-92eb510e.js:3
 (anonymous function)                                       all-92eb510e.js:11
 Z.event.dispatch                                           all-92eb510e.js:2
 m.handle                                                   all-92eb510e.js:2
```

joecohens:
I've just tested the latest version with PostgreSQL and working ok. @jbrooksuk Can you help me testing with sqlite?

jbrooksuk:
Certainly my man, give me a couple of minutes.

joecohens:
Thanks :)

jbrooksuk:
I've changed my `.env.local.php` to `sqlite` and it just hates me. Sorry for the delay.

jbrooksuk:
Doesn't work on SQLite.

jbrooksuk:
Should be closed in #398

PR #395, opened 2015-01-16
Use model on setup no facade

joecohens:
Fixes https://github.com/cachethq/Cachet/issues/394


GrahamCampbell:
What if our repository facade supported setting config too?

joecohens:
Yeah it would be ideal :)

joecohens:
Merging this for now since the setup is broken.

GrahamCampbell:
:+1:

PR #404, opened 2015-01-18
[WIP] Scheduled statuses

jbrooksuk:
- [x] Update incident form.
- [x] Build in date picker.
- [x] Store scheduled statuses.
    - [x] Fix bug where `publish_at` date doesn't get stored.
- [ ] Language files.
- [x] Continue to show incidents but which don't have the same `publish_at` as they do `created_at` - optional?
- [ ] Show scheduled statuses above incidents.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -13,7 +13,8 @@
         "animate-sass": "~0.6.2",
         "moment": "~2.9",
         "livestampjs": "~1.1.2",
-        "chartjs": "~1.0.1"
+        "chartjs": "~1.0.1",
+        "bootstrap-datepicker": "~1.3.1"
     },
     "resolutions": {
         "jquery": "~2.1.1"
--- a/src/Http/Controllers/HomeController.php
+++ b/src/Http/Controllers/HomeController.php
@@ -5,7 +5,6 @@ namespace CachetHQ\Cachet\Http\Controllers;
 use CachetHQ\Cachet\Facades\Setting;
 use CachetHQ\Cachet\Models\Component;
 use CachetHQ\Cachet\Models\Incident;
-use CachetHQ\Cachet\Models\Metric;
 use Carbon\Carbon;
 use Exception;
 use GrahamCampbell\Binput\Facades\Binput;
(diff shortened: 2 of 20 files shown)


bower.json
```diff
@@ -12,7 +12,8 @@
     "Sortable": "~1.0.0",
     "animate-sass": "~0.6.2",
     "moment": "~2.8.4",
-    "livestampjs": "~1.1.2"
+    "livestampjs": "~1.1.2",
+    "bootstrap-datepicker": "1.3.1"
```
GrahamCampbell: should be ~1.3.1

jbrooksuk: I remember doing this after realising I'd forgotten - but I know why now. The tilde character in OSX is a modifier and if you don't provide a character to modify the ~ will disappear.

jbrooksuk:
I'm wondering if we should actually split the scheduled statuses into a different page? The reason for this is that scheduled statuses wouldn't likely have a status - you'd use them for maintenance etc. The other way is when a `published_at` date is chosen, we could disable the statuses and just send through `0.

jbrooksuk:
CC @cachethq/owners

src/Http/Controllers/HomeController.php
```diff
@@ -48,7 +48,7 @@ public function showIndex()
 
         foreach (range(0, $incidentDays) as $i) {
             $date = $startDate->copy()->subDays($i);
-            $incidents = Incident::whereBetween('created_at', [
+            $incidents = Incident::notScheduled()->whereBetween('created_at', [
```
joecohens: This will disappear the status once the published date is today, we need a way to put it back to the normal flow.

joecohens: Maybe use the `published_at` date on the where here.

jbrooksuk:
Also, we should offer scheduled incidents at a certain time.

jbrooksuk:
Rebased.

jbrooksuk:
I'm going to move scheduled statuses into a new form, rather than dirtying the existing one.

jbrooksuk:
@GrahamCampbell what's going on with your Markdown package?

GrahamCampbell:
I told you this already. You need to clear out your composer cache. It's using a version that doesn't even exist.

jbrooksuk:
I did.

GrahamCampbell:
Did you run these commands?

```
composer self-update
composer clear-cache
rm -rf vendor
rm composer.lock
composer install
```

NB, composer update doesn't cut it because composer is stupid.

jbrooksuk:
I rebased `master` on to this branch and pushed so hopefully Travis will pass now.

jbrooksuk:
Rebased.

jbrooksuk:
I've messed up a rebase here and lost a few things from `master`. Will need to redo this now :angry:

PR #408, opened 2015-01-21
[WIP] Add metrics

joecohens:


jbrooksuk:
Awesome! I'm going to check this out now.

jbrooksuk:
I know this is a WIP, but couple of things:

1. "Create a metric point" should be something like "Create metric". The metric itself is what the current form does, whereas the point is a measurement of that metric. So you'd add a point to the metric, make sense?
2. Can we show the metric description on the Metrics page?

joecohens:
Sure for both things! I think the "Create a metric point" lang string, was already there before this PR, thanks for spotting it :)

jbrooksuk:
Needs rebasing.

jbrooksuk:
How far off are we on this? :)

joecohens:
I just rebased, pushing. The only thing missing is to set the graph data.

jbrooksuk:
So just the metric points themselves now?

jbrooksuk:
@joecohens can you update the original issue with a task list for what's left?

joecohens:
Moved to https://github.com/cachethq/Cachet/pull/449.

PR #410, opened 2015-01-22
Update ComponentTableSeeder.php

coreation:
Tags were removed from the component table.

### DIFF
--- a/app/database/seeds/ComponentTableSeeder.php
+++ b/app/database/seeds/ComponentTableSeeder.php
@@ -24,7 +24,6 @@ class ComponentTableSeeder extends Seeder
                 "order"       => 0,
                 "group_id"    => 0,
                 "link"        => "",
-                "tags"        => "",
             ], [
                 "name"        => "Payments",
                 "description" => "Backed by Stripe",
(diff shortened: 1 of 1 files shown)


jbrooksuk:
:beers:

coreation:
I have a question about the usage of this platform, say I want to monitor the status of a certain API or URL how do I get started with that? I see you can add a component with a link for example, but I want the status of that API link depending on what HTTP response that link says (e.g. 200 = ok still running, 500 = oops, not ok).

jbrooksuk:
That's now what Cachet is, for something like that you should look at Uptime Robot or Pingdom. What you can do though is integrate either of those services with Cachet, by using the API to automatically update the statuses or add an incident.

coreation:
Aha, thanks ^^

On Thu, Jan 22, 2015 at 6:07 PM, James Brooks <notifications@github.com>
wrote:

> That's now what Cachet is, for something like that you should look at
> Uptime Robot or Pingdom. What you can do though is integrate either of
> those services with Cachet, by using the API to automatically update the
> statuses or add an incident.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/cachethq/Cachet/pull/410#issuecomment-71056568>.
>



-- 
Met vriendelijke groeten,

Jan Vansteenlandt

jbrooksuk:
:+1:

PR #413, opened 2015-01-23
Added a list of things to what Cachet is not

jbrooksuk:
Same questions, different people. Let's clear some things up.


joecohens:
:+1:

PR #414, opened 2015-01-23
Anonymous tracking with Segment.com

### LINKED ISSUE #91: Tracking
I'm wondering whether it'd be worth adding some kind of internal tracking that gets sent to directly back to us, to see how often something happens.

We'd only need to track:

- Installs
- Components; added, edited, deleted
- Incidents; added, edited, deleted
- Notifications

Then we'd have enough data to use as "we've had x installs" etc.

Of course, there are things like:

- Security
- Privacy

That we'd need to take care of.

jbrooksuk:
Closes #91.

We track a few things, but could easily add more, plus it's all anonymous!

# Things to do:

- [x] Setting of the Segment.com `write_key` should be based on an API result we fetch and store somewhere, this allows us to revoke it at any time from all status pages.
- [x] What happens if the API key is completely wrong and the new one is not installed? Does Segment lose its shit? **Nothing. It just doesn't work - this is good.**

### DIFF
--- a/app/config/app.php
+++ b/app/config/app.php
@@ -127,16 +127,16 @@ return [
          * Packages Service Providers...
          */
         'Dingo\Api\Provider\ApiServiceProvider',
-        'GrahamCampbell\Security\SecurityServiceProvider',
+        'Fideloper\Proxy\ProxyServiceProvider',
         'GrahamCampbell\Binput\BinputServiceProvider',
-        'GrahamCampbell\Throttle\ThrottleServiceProvider',
         'GrahamCampbell\Markdown\MarkdownServiceProvider',
-        'Roumen\Feed\FeedServiceProvider',
-        'Thujohn\Rss\RssServiceProvider',
+        'GrahamCampbell\Security\SecurityServiceProvider',
+        'GrahamCampbell\Throttle\ThrottleServiceProvider',
         'Jenssegers\Date\DateServiceProvider',
         'McCool\LaravelAutoPresenter\LaravelAutoPresenterServiceProvider',
         'PragmaRX\Google2FA\Vendor\Laravel\ServiceProvider',
-        'Fideloper\Proxy\ProxyServiceProvider',
+        'Roumen\Feed\FeedServiceProvider',
+        'Thujohn\Rss\RssServiceProvider',
 
         /*
          * Application Service Providers...
@@ -148,6 +148,8 @@ return [
         'CachetHQ\Cachet\Providers\ViewComposerServiceProvider',
         'CachetHQ\Cachet\Providers\LoadConfigServiceProvider',
         'CachetHQ\Cachet\Providers\SettingsServiceProvider',
+        'CachetHQ\Cachet\Providers\SegmentApiServiceProvider',
+        'CachetHQ\Segment\SegmentServiceProvider',
 
     ],
 
--- a/app/start/global.php
+++ b/app/start/global.php
@@ -61,3 +61,14 @@ App::missing(function ($exception) {
 App::down(function () {
     return Response::make("Be right back!", 503);
 });
+
+/*
+|--------------------------------------------------------------------------
+| Analytics
+|--------------------------------------------------------------------------
+|
+| We handle all Segment.com tracking here. The function call will only do
+| something if tracking is enabled.
+|
+*/
+segment_identify();
--- a/src/Providers/LoadConfigServiceProvider.php
+++ b/src/Providers/LoadConfigServiceProvider.php
@@ -25,6 +25,10 @@ class LoadConfigServiceProvider extends ServiceProvider
             // Don't throw any errors, we may not be setup yet.
         }
 
+        // Set the Segment.com settings.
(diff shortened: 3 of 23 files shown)


app/config/app.php
```diff
@@ -137,6 +137,7 @@
         'McCool\LaravelAutoPresenter\LaravelAutoPresenterServiceProvider',
         'PragmaRX\Google2FA\Vendor\Laravel\ServiceProvider',
         'Fideloper\Proxy\ProxyServiceProvider',
+        'CachetHQ\Segment\SegmentServiceProvider',
```
GrahamCampbell: Needs to be alphabetically sorted,

app/start/global.php
```diff
@@ -61,3 +63,9 @@
 App::down(function () {
     return Response::make("Be right back!", 503);
 });
+
+/**
+ * Segment.com tracking.
+ * Since every user is identified under one site, we can "identify" them here.
```
GrahamCampbell: We need another new line before the start of the long description.

app/start/global.php
```diff
@@ -61,3 +63,9 @@
...
+ */
+segment_identify(Config::get('app.key'));
```
GrahamCampbell: This means we're setting this up on every request. Probably not what we want to do.

app/config/app.php
jbrooksuk: Ok :)

app/start/global.php
jbrooksuk: Gotcha.

app/start/global.php
jbrooksuk: What I could do is move this line into the `segment_` helpers?

jbrooksuk:
I've setup Keen.io integration which is much better for our usage.

jbrooksuk:
@GrahamCampbell I'm storing the write key in a private Gist file. It's nothing bad ass, but since we have to release it at some point, it's the easiest way.

What do you think is the best way for us to load this? I'm thinking of having it store as a setting, then if the `created_at` is more than a day old, check again. We wouldn't lose too much data if it then had to be changed.

jbrooksuk:
CC @joecohens (see above).

jbrooksuk:
1. Needs rebasing.
2. If we can figure out how to remotely fetch the write key every few hours, then update the setting, we're able to merge.

jbrooksuk:
Rebased :)

src/Providers/LoadConfigServiceProvider.php
```diff
@@ -26,6 +26,11 @@ public function boot()
             // Don't throw any errors, we may not be setup yet.
         }
 
+        // Set the Segment.com settings.
+        $repository = $this->app->make('CachetHQ\Cachet\Segment\RepositoryInterface');
+        $segmentWriteKey = $repository->fetch();
+        $this->app->config->set('cachethq/segment::write_key', $segmentWriteKey);
```
joecohens: Inline?

src/Segment/CacheRepository.php
```diff
@@ -0,0 +1,47 @@
...
+
+use CachetHQ\Cachet\Facades\Setting;
+use CachetHQ\Cachet\Models\Setting as SettingModel;
+use Carbon\Carbon;
+
+class CacheRepository implements RepositoryInterface
+{
+    protected $repository;
+
+    public function __construct(RepositoryInterface $repository)
```
joecohens: Comments

src/Providers/LoadConfigServiceProvider.php
jbrooksuk: ?

src/Segment/RepositoryInterface.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace CachetHQ\Cachet\Segment;
+
+interface RepositoryInterface
+{
+    public function fetch();
```
joecohens: Comments

src/Providers/LoadConfigServiceProvider.php
jbrooksuk: Gotcha. Will fix.

src/Segment/RepositoryInterface.php
jbrooksuk: I knew I'd forgotten something :P

src/Providers/LoadConfigServiceProvider.php
joecohens: `$segmentRepository = $this->app->make('CachetHQ\Cachet\Segment\RepositoryInterface');`
`$this->app->config->set('cachethq/segment::write_key', $segmentRepository->fetch());`

jbrooksuk:
Just needs a rebase now.

joecohens:
@jbrooksuk Needed rebasing again lol.

jbrooksuk:
I take it you fixed style CI?

joecohens:
Yep I've fixed StyleCI, rebased and merged :)

PR #418, opened 2015-01-24
Use phpunit.xml.dist instead of phpunit.xml

dosten:
The project must distribute a phpunit.xml.dist instead of a phpunit.xml, the reason is that phpunit use a phpunit.xml file and if not exists use a phpunit.xml.dist. this is useful if someone tries a new phpunit configuration, only has the need to create a phpunit.xml.


GrahamCampbell:
Thanks.

PR #425, opened 2015-01-26
Use CrowdIn's In-Context Localization Feature

jbrooksuk:
This is pretty cool, but will require some maintenance on my side.

Essentially if you set the locale to `en-UD` (via the dashboard) you'll then be shown the In-Context feature, where you can translate to a target language within the application. You have to set some kind of psuedo-language for the translation to sit within, I've chosen `en-UD` but it could be anything. This is just the simplest.

image

Then you'll get all of this:

image

And now you can submit translations.

# Things to do

- [x] Turn CrowdIn JS into a partial file.
- [x] Perhaps move the pseudo language to something more obscure than `en-US`.

~~Once CrowdIn have fixed using of the short-array syntax this will be so much more awesome for us!~~

### DIFF
--- a/app/views/layout/error.blade.php
+++ b/app/views/layout/error.blade.php
@@ -32,6 +32,8 @@
     </style>
     @endif
 
+    @include('partials.crowdin')
+
     <script src="{{ elixir('js/all.js') }}"></script>
 </head>
 <body class="error-page">
--- a/app/views/partials/dashboard/head.blade.php
+++ b/app/views/partials/dashboard/head.blade.php
@@ -23,6 +23,8 @@
     <link href="//fonts.googleapis.com/css?family=Lato:300,400,700" rel="stylesheet" type="text/css">
     <link rel="stylesheet" href="{{ elixir('css/all.css') }}">
 
+    @include('partials.crowdin')
+
     <script type="text/javascript">
         var Global = {};
         Global.locale = '{{ Setting::get('app_locale') }}';
--- a/app/config/langs.php
+++ b/app/config/langs.php
@@ -4,6 +4,7 @@ return [
     // Enabled langs
     'de'    => 'Deutsch',
     'en'    => 'English',
+    'en-UD' => 'CrowdIn - InContext Localization',
     'fr'    => 'Français',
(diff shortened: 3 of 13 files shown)


joecohens:
It seems pretty cool!

jbrooksuk:
It'll definitely help out with translations.

jbrooksuk:
CrowdIn have now fixed the short-array syntax issues which means it'll be much easier for us to update and create translations!

pedrommone:
Awesome!

jbrooksuk:
I've switched to using `en-UD` (upside-down) for the pseudo language. This works much better for us.

jbrooksuk:
Rebased and ready to go if @cachethq/owners believe this is a good idea.

joecohens:
:+1:

app/views/partials/dashboard/head.blade.php
```diff
@@ -23,6 +23,8 @@
     <link href="//fonts.googleapis.com/css?family=Lato:300,400,700" rel="stylesheet" type="text/css">
     <link rel="stylesheet" href="{{ elixir('css/all.css') }}">
 
+    @include('partials.crowdin')
+
```
joecohens: We should do `@if(Setting::get('app_locale') === 'en-UD')`?

app/views/layout/error.blade.php
```diff
@@ -32,6 +32,8 @@
     </style>
     @endif
 
+    @include('partials.crowdin')
+
```
joecohens: We should do `@if(Setting::get('app_locale') === 'en-UD')`?

jbrooksuk: That's within the partial?

app/views/partials/dashboard/head.blade.php
jbrooksuk: It's within the partial.

app/views/layout/error.blade.php
joecohens: Oh ok you are right!

joecohens:
Merging

jbrooksuk:
:beers:

PR #427, opened 2015-01-26
Patch.components

JMousqueton:
There is a bug when you add component which remain  in the group_id 0 
There is a bug when you edit component which remains you to change the group_id 

all bugs are due to an hidden field group_id at the bottom of the HTML form


jbrooksuk:
The hidden group field is there so that if you don't submit set a group, it will default to 0. This means all database drivers are taken care of.

jbrooksuk:
Fixed in 92e2ec188cca6e5cc9d950926891aec743d33659.

The hidden field just needs to go above the selects.

jbrooksuk:
Thanks though :)

Issue #419, opened 2015-01-24
Incident not obeying timezone


jenolan:
I am in Oz and have set my timezone, when I add an issue the incident is showing as yesterday. Screen shot below, I adding it Oz time 0900 2015-01-25 but it time travelled back to the 24th, I am sure Mr Hawking said that shouldn't happen ;-)
Site is https://jenolan.herokuapp.com/

![screenshot 2015-01-25 09 06 56](https://cloud.githubusercontent.com/assets/1505228/5889275/bd8d31aa-a471-11e4-9b34-642124f48bd8.png)

jbrooksuk:
What timezone did you pick specifically?

jenolan:
Australia/Sydney

jbrooksuk:
That *could* be the problem then. The server time is in America, so the date doesn't change but the time does?

jenolan:
Well the server itself should be UTC, but the PHP may be set to local time, then Cachet is fiddling that again?

jbrooksuk:
In the background Cachet is just setting `app.timezone` to whatever is selected.

CC @GrahamCampbell any advice on this?

jenolan:
By looking at the incident time, it is saying 24/1 2201 that's the UTC time that I posted the incident so the code is calling a UTC function rather than local time?

jbrooksuk:
Seems to be, I think we need to call `Carbon::setTimezone()` or something too. @joecohens any idea on this as you wrote the `LoadConfigServiceProvider`, do you experience it in Mexico?

jbrooksuk:
Alternatively all models need a view presenter for dates, then each date will be chained with:

```php
return $this->resource->created_at->tz(Setting::get('app_timezone'));
```

For example. Thoughts?

jbrooksuk:
Ok, that does seem to make the incident time tooltip show the right time, but the incident is still pulled under the 24th.

Looks like the database server now needs to be aware of the timezone?

Jamesking56:
I think when I had this problem with Laravel before I solved it by setting Laravel's config to UTC and then using Carbon to set the user's timezone when it pulls data from the DB

GrahamCampbell:
@Jamesking56 Yeh, that's what you're meant to do. :)

jbrooksuk:
Ok, so in `LoadConfigServiceProvider` we should call `Carbon::setTimezone()`?

GrahamCampbell:
Maybe, yeh.

[...]

PR #435, opened 2015-01-27
Fix timezones presenting

joecohens:
Using Carbon again to query and fix presenters to use timezone. Needs confirmation.

Fixes https://github.com/cachethq/Cachet/issues/419

### DIFF
--- a/src/Http/Controllers/HomeController.php
+++ b/src/Http/Controllers/HomeController.php
@@ -5,6 +5,7 @@ namespace CachetHQ\Cachet\Http\Controllers;
 use CachetHQ\Cachet\Facades\Setting;
 use CachetHQ\Cachet\Models\Component;
 use CachetHQ\Cachet\Models\Incident;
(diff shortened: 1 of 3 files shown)


jbrooksuk:
@joecohens you'll also probably want to remove the timezone setting in `LoadConfigServiceProvider` as we should remain at UTC.

joecohens:
Not sure if it makes difference, let me check.

jbrooksuk:
I think we should be keeping the timezone as UTC, but it's worth checking.

joecohens:
I'm your right I don't think it's making the change since Laravel sets the default date time https://github.com/laravel/framework/blob/36487a19f399fbe246fd9170d4c2dba969307ae8/src/Illuminate/Foundation/start.php#L167 

So the override is not working at all :stuck_out_tongue_winking_eye:

jbrooksuk:
Let's get this bad boy merged!

PR #441, opened 2015-01-29
Easier licensing terms

jbrooksuk:
This is the same license that Sentry uses.


GrahamCampbell:
You need to update the license attribute in the composer.json too.

jbrooksuk:
Good point :)

PR #455, opened 2015-02-18
Component can be created with tags from API

joecohens:
//cc @jbrooksuk


jbrooksuk:
Sweet! Nice on @joecohens.

Issue #448, opened 2015-02-09
Docker should use nginx


jbrooksuk:
This would make it more lightweight and easier to configure. Plus, we should really be advising use of a better web server anyway.

joecohens:
:+1:

sammcj:
+1

jbrooksuk:
I've no experience with Docker, so if somebody fancies doing this, that'd be lovely.

GrahamCampbell:
Docker is apparently an arse to configure. For example, if we had our db in a different container, we've got no way of knowing that other container's ip or hostname before we've booted it so our web container is kind of screwed.

sammcj:
We've got lots of docker & nginx experience @infoxchange but almost no PHP experience - I'll take a look though and see what I can come up with this week unless work gets busier.

jbrooksuk:
@GrahamCampbell can you get on Slack for a minute? :)

sammcj:
@grahamcampbell - it's not an arse to configure at all? Docker is very easy to work with compared to say - deploying with multiple package managers for all the crazy languages out there.  

With the database, you can just require the use of another upstream database container which you connect to via environment variables or a config file presented to the container.

Anyway, if I get time this week I'll take a look at it.

GrahamCampbell:
@sammcj That's not what I've been told. I too have little experience with docker though.

sammcj:
One of its main advantages is that it makes deployment easy and reproducible.

We moved all our applications, old and new alike from VMs into containers last year many of them are highly complex with many relaionships and dependant services - I'll tell you what - it's been a BIG win for us, it's sped up development, increased the frequency of feature deployments, made CI and testing easier, made load balancing an failover easier, reduced DR times by over 600% and the list goes on.

Don't get me wrong, it's not all roses, but it certainly makes it easier for devs and ops to do the right thing.

Maybe not the best example but: http://codesamplez.com/web-server/dockerize-an-application-from-scratch

sammcj:
@jbrooksuk ,

I've just given @frconil from our systems engineering team @infoxchange an initial one day on sorting out Nginx and the Dockerfile for Cachet.

jbrooksuk:
You rock! Thanks @sammcj :beers:

frconil:
Hi,

I've done some work in https://github.com/frconil/Cachet/commit/d132eec54b2c35eb4cb906b9eb525bdad16bc7a5 

[...]

PR #456, opened 2015-02-19
Use nginx instead of apache in docker image

frconil:
As discussed in https://github.com/cachethq/Cachet/issues/448

This image use debian:jessie instead of php:56-apache (which uses jessie as well)

I'm sure it could be optimised further, but it's working as expected. I also amended the documentation as I could not get artisan migrate to work without passing it --force.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,4 @@
-FROM php:5.6-apache
-
-# before building
-# run npm install
-# run node_modules/.bin/bower install
-# run node_modules/.bin/gulp
+FROM debian:jessie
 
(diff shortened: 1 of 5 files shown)


sammcj:
I have tested this on a clean Debian 7 VM running Docker 1.5.0
It builds and functions correctly as far as I could see - anyone else want to test?

jbrooksuk:
Awesome. I don't know enough about Docker to really test this, but I'm happy to offer a solution to those of you who do use it.

Thanks for this :beers:

sammcj:
ONWARD! :horse_racing:

Issue #463, opened 2015-02-20
Include compiled assets in repo, drop building with Node


jbrooksuk:
We've had **a lot** of reoccurring issues with Heroku deployment due to not including the compiled assets within the repo.

This should give us the following benefits:

- Hopefully be able to drop the custom buildpack.
- Reduce the size of the deployed application on Heroku.
- Removes the need for people installing Cachet to have Node, Bower, Gulp etc installed.

joecohens:
:+1:

GrahamCampbell:
:+1:

GrahamCampbell:
This only works if there's not much going on, otherwise we've going to have a crap load of merge conflicts again.

jbrooksuk:
I said to @joecohens that we can probably do this for V1 - or start now as there is less changes happening.

jbrooksuk:
@sammcj @frconil would you guys be able to help with the Docker changes that would be required by this?

frconil:
Sure, I'll submit a PR when you move to prebuilt assets.

On 21 Feb 2015, at 08:57, James Brooks <notifications@github.com<mailto:notifications@github.com>> wrote:


@sammcj<https://github.com/sammcj> @frconil<https://github.com/frconil> would you guys be able to help with the Docker changes that would be required by this?

-

jbrooksuk:
Thanks!

I hope you're not annoyed, since you've just done it, but after all the hassle I've had with asset compilation today, it's not worth it - only for development.

frconil:
It's ok, I was complaining about npm timing out all the time while building the docket image.

I will have a look at it first thing Monday.

On 21 Feb 2015, at 10:30, James Brooks <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks!

I hope you're not annoyed, since you've just done it, but after all the hassle I've had with asset compilation today, it's not worth it - only for development.

-

sammcj:
I'm so glad this is happening, so sick of all the NPM and Composer building, especially when I get rate-limited by Github :angry:

jbrooksuk:
Composer will still need to run, but NPM and Bower won't :)

Issue #384, opened 2015-01-15
Dashboard polish


jbrooksuk:
The dashboard needs some polish on it :lipstick:

- [x] Increase font sizes for small text.
- [x] Layout forms in the same way, giving them a universal design.
- [ ] The active sidebar item is slightly taller, which when you jump between pages means the sidebar items move slightly. Some kind of `box-sizing: border-box` should fix this.
- [x] Forms need tidying up.
- [x] Show the `ion-social-markdown` icon in the bottom right of the supported textareas rather than having the `.help-block` element below it.
- [x] Inline component status needs cleaning up, making everything uniform and aligned correctly.
- [x] Use `.btn-group` when we have two or more buttons that are linked to something.
- [x] The sub-sidebar found on some pages only ever has a fixed height. If the page scrolls, the sub-sidebar height remains fixed at the top.
- [x] Consistency, buttons, layouts etc.
- [x] Font thickness. I think generally it's too thin all around.
- [x] Switch to Open Sans over Lato. I regret that choice.
- [ ] Make the user profile link more obvious.

jbrooksuk:
@thomasHolman can I pick your brains one of the things on here, on Monday please?

thomasHolman:
Sure 👍

jbrooksuk:
Thanks :)

PR #465, opened 2015-02-21
Store only build compiled assets on public

joecohens:
//cc @jbrooksuk @GrahamCampbell 

Closes https://github.com/cachethq/Cachet/issues/463 and inline components on https://github.com/cachethq/Cachet/issues/384


src/Models/Incident.php
```diff
@@ -93,13 +93,13 @@ public function getIconAttribute()
     {
```
joecohens: This should be moved to the presenter on other PR

jbrooksuk: Agreed!

jbrooksuk:
Done.

jbrooksuk:
@joecohens I've pulled and this doesn't actually work anymore. JS assets are no longer compiled. Did I break that?

joecohens:
@jbrooksuk Have you reinstalled the npm modules?

PR #469, opened 2015-02-22
update Dockerfile to use pre-built assets

frconil:
As discussed in https://github.com/cachethq/Cachet/issues/463, here is a newly updated Dockerfile that works with pre-built assets.


jbrooksuk:
:beers: thanks so much :)

joecohens:
This looks so much simpler :smile:

PR #476, opened 2015-02-26
Further reduce size of Docker image

sammcj:
* Removing man pages, logs generated by the image build process
* Add apt-get autoremove to cleanup unused packages at build
* Move environment variable for non-interactive mode to ENV


jbrooksuk:
This is awesome! I really need to start using things like Docker.

PR #478, opened 2015-02-26
Catch invalid validation on api requests

joecohens:
// cc @jbrooksuk What do you think?


GrahamCampbell:
It would be better to do this using a try catch in a middleware imo.

jbrooksuk:
@GrahamCampbell any suggestions on how that'd work?

joecohens:
@GrahamCampbell Lol no middlewares on L4 yet but you are right might be better on a global catch or filter

GrahamCampbell:
> Lol no middlewares on L4

No, L4 does have middlewares. :)

PR #506, opened 2015-03-11
Fixes scheduled incidents timezones

joecohens:
/cc @jbrooksuk First we get the user local time, then we convert it to the app (UTC) tz to store it to the DB, that way the incident presenter and the created_at and updated_at will be on the same tz. Hope it makes sense :)


jbrooksuk:
Ah-ha! Nice :+1:

GrahamCampbell:
:ship:

Issue #507, opened 2015-03-11
Incident not obeying timezone


jenolan:
See #419 dates are still not right.

To replicate .. change timezone to Sydney/Australia between about 0600-0900 add an incident, then go to the status page and you will see it has moved to yesterday.

jenolan:
Not scheduled standard incident

jbrooksuk:
@joecohens FYI.

joecohens:
@jenolan This should fix it let us know if you have any trouble. https://github.com/cachethq/Cachet/pull/508

PR #508, opened 2015-03-11
Use Carbon to query scheduled incidents

joecohens:
/cc @jbrooksuk


jbrooksuk:
Should fix #507 completely?

jbrooksuk:
Merge if you're happy @joecohens

joecohens:
Yeah this should fixit completely :) I would like confirmation from @jenolan

Issue #515, opened 2015-03-15
Metric points not displayed on graph


matejkramny:
I'm writing a [third party plugin](https://github.com/CastawayLabs/cachet-monitor) that hits a url and `POST`s the lag to cachet as a metric point (every second).

I found that the points get created in the mysql database, but are never displayed on the actual graph.

I started logging the metric points to cachet's demo, you can get the metrics via `https://demo.cachethq.io/api/metrics/1/points` (you need to use Postman or curl, setting the api token).

Any ideas what could be wrong?

jbrooksuk:
Are you testing this solely on the demo instance or do you have your own version?

matejkramny:
I installed a local version and that doesn't work either.I tailored it for the demo so it is easier to test.Matej KramnyOn Mon, Mar 16, 2015 at 9:36 am,  James Brooks <notifications@github.com> wrote:            Are you testing this solely on the demo instance or do you have your own version?

jbrooksuk:
You need to [setup a metric](https://demo.cachethq.io/dashboard/metrics) to actually send data to. Remember that the demo resets every half hour.

matejkramny:
James, i did create a metric, and the api call returned 200 with json representation of the metric point.

The points were also present in the database.

Please reopen?

matejkramny:
I am not near my laptop, so writing from my phone.

Steps to reproduce:

1. Log into cachet demo
2. Create a metric, note the id
3. Go to settings and enable graphs to be shown on the homepage
4. Start Postman app (available in chrome app store)
5. Set url to POST to https://demo.cachethq.io/api/metrics/1/points
6. Add the required header for authentication
7. Click the Raw tab, enter {"value":100} and then send
8. Make a get with the same url to confirm it has been created
9. See cachet uptime dashboard (root url). The graph is there but no data.

joecohens:
@matejkramny Is the raw tab sending the request data has JSON? If it is you should set the `Content-Type: application/json` header if not you need to send it like form values ex. `value=10`.

matejkramny:
I am on my laptop so i'll try & verify.

matejkramny:
It will if you select JSON from the dropdown (its by the `...| raw | binary` tabs).

What i get back from `/api/metrics/2/points`:

```
{
  "metric_points": [
    {
      "id": "2482",
      "metric_id": "2",
      "value": "320.000",
      "created_at": "2015-03-16 19:38:44",
[...]

PR #520, opened 2015-03-17
Allow metrics to calculate the sum or average. Closes #517.

### LINKED ISSUE #517: Metrics adding up instead of averaging
I am trying to set up so that metrics are pushed every second to the server (ping). The server is just adding them up instead of averaging them out.

Image of graph area
Image of points

jbrooksuk:
Also fixes a bug found in #515 where only the last graph would render.

### DIFF
--- a/app/views/dashboard/metrics/add.blade.php
+++ b/app/views/dashboard/metrics/add.blade.php
@@ -32,6 +32,13 @@
                             </div>
                         </div>
                         <div class="form-group">
+                            <label>{{ trans('forms.metrics.calc_type') }}</label>
+                            <select name="metric[calc_type]" class="form-control" required>
+                                <option value="0" selected>{{ trans('forms.metrics.type_sum') }}</option>
+                                <option value="1">{{ trans('forms.metrics.type_avg') }}</option>
+                            </select>
+                        </div>
+                        <div class="form-group">
                             <label for="metric-default_value">{{ trans('forms.metrics.default-value') }}</label>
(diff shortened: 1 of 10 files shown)


jbrooksuk:
@joecohens this is also backwards compatible with existing metrics :+1:

joecohens:
Sweet! :+1:

app/views/dashboard/metrics/add.blade.php
```diff
@@ -32,6 +32,13 @@
                             </div>
                         </div>
                         <div class="form-group">
+                            <label>{{ trans('forms.metrics.type') }}</label>
```
joecohens: Should be `forms.metrics.calc_type` or we need to change the lang key :)

app/views/dashboard/metrics/edit.blade.php
```diff
@@ -32,6 +32,13 @@
                             </div>
                         </div>
                         <div class="form-group">
+                            <label>{{ trans('forms.metrics.type') }}</label>
```
joecohens: Same here.

app/views/dashboard/metrics/add.blade.php
jbrooksuk: Whoops!

app/views/dashboard/metrics/edit.blade.php
jbrooksuk: This is a weird change.

jbrooksuk:
All done :)

PR #522, opened 2015-03-18
Allow supplying of metric point timestamps. Closes #519.

### LINKED ISSUE #519: Supply timestamps for metric points
Metrics need to be able to be sent a timestamp so that data can be backfilled.

jbrooksuk:
If you supply a `timestamp` (in unix) it'll be applied to the `created_at` and `updated_at` values, so you can backfill metric points.


joecohens:
:+1:

PR #531, opened 2015-03-22
Add Addons section

### LINKED ISSUE #527: Server Monitoring app
I've decided to write a [server app](https://github.com/CastawayLabs/cachet-monitor) (for fun) that integrates Cachet's API.

It is now at a stage where 'it works' and creates an incident if a monitor goes offline. It then resolves the incident should the monitor go back up.. There are more features such as ping logging however that seems broken at the moment.

- Would you think its a good idea to list the addon in readme?
- Can you share some guidance that matches your vision for cachet?
- Your thoughts on third party plugins. I can't find any at the moment

The issues created in the project are goals for v1 (irrelevant of cachet's version).

matejkramny:
Resolves #527


jbrooksuk:
Smashing, thanks!

PR #536, opened 2015-03-23
more precise french translation for "watching"

anarcat:
`Analyse en cours` means `Under analysis`, which is basically the same as status `1` (`Investigating`), so a little misleading. I chose `Sous surveillance` (litterally `Under surveillance`) because it is more explicit. `Surveillé` (litterally `surveilled`) seemed a bit weird.


jbrooksuk:
Ok, that makes sense to me.

PR #539, opened 2015-03-24
'composante' is feminine in french, also adjust other fr trans

anarcat:
'Il s'est passé quelquechose...' doesn't actually mean anything...


anarcat:
well that was fast, thanks!

PR #545, opened 2015-03-25
add maintenance translations

anarcat:
it's unclear to me why those parts of the array weren't present in the first place.

it would be easier to maintain translations if it was using standard formats like the de-facto `gettext` standard: https://en.wikipedia.org/wiki/Gettext


jbrooksuk:
Translations are powered by Symfony's translator, which is what Laravel uses. They weren't in the array (but should've been) because it'll just fallback to `en`.

jbrooksuk:
Thanks :beer:

PR #559, opened 2015-03-31
Update .env.example

Duffleman:
DB_DRIVER is default set to sqlite. Anyone starting the project intending to use MySQL will have to look around to see to add that declaration here. Best add it to the example file to save them the hassle.


jbrooksuk:
Welcome to da club :notes:

PR #561, opened 2015-04-01
OpenShift Deployment Support

StartledPhoenix:
- OpenShift One Click Deployment Button
- Openshift deployment hooks
- env.local.php for OpenShift
- Updating of composer.json to allow both apc/apcu


StartledPhoenix:
Edit: wrong button link, will fix and do a new PR

PR #565, opened 2015-04-03
Drop Heroku support

jbrooksuk:
Heroku support is now dropped under L5.


jbrooksuk:
Whoops, I should've pushed the first commit first. Never mind.

Issue #447, opened 2015-02-07
Chinese is unable to display


wfjsw:
Due to fonts reason, when an Asian language is installed, the char will just disappeared. They cannot display properly in that environment.

I am using the Crowdin's translate file.

jbrooksuk:
Thanks for reporting this. I guess that's to do with Open Sans? Or is the font displaying as Lato for you? It depends on if you pulled recently.

wfjsw:
first, open sans never supports chinese.
second, it seems that google fonts' charset have some problem with processing it :(

jbrooksuk:
I'm not sure how we can fix this then I'm afraid.

wfjsw:
simply change an charset-friendly font when selecting asian language.

jbrooksuk:
I'm wondering if we make it so that if the language is Chinese or anything with other unsupported diacritics etc use a different font.

We can't drop Open Sans easily as it's part of the consistent branding for Cachet.

wfjsw:
Fine.

jbrooksuk:
> I'm wondering if we make it so that if the language is Chinese or anything with other unsupported diacritics etc use a different font.

What do you think about that @wfjsw

wfjsw:
Exactly the same. System Fonts did well in it.

jbrooksuk:
Do you have any ideas on how we can fix this without losing the font for the other languages?

wfjsw:
I am now wondering what cause that problem. Google Fonts or CSS?
My solution is using php to dynamic output an override-css ('!important' may help) file when using these languages. 

And another problem is that google is blocked in Mainland China. Any query to their services will slow down whole page. (We use mirrors or complete remove them)

jbrooksuk:
Closed by #570.

wfjsw:
@jbrooksuk 
Crazy things happened on Crowdin: My translations all disappeared:
![sshot1](https://cloud.githubusercontent.com/assets/2220320/7016336/d525ae60-dd14-11e4-8edb-77c01118815a.JPG)
![sshot2](https://cloud.githubusercontent.com/assets/2220320/7016337/d727a376-dd14-11e4-893d-868481117fb1.JPG)
But......
![sshot3](https://cloud.githubusercontent.com/assets/2220320/7016339/db0e9f6c-dd14-11e4-9301-f0dcd8efc35b.JPG)
What is happening in the world......

jbrooksuk:
[...]

PR #570, opened 2015-04-05
zh-Hant and zh-Hans Translation

applelam:
Translation for Chinese (Traditional) and Chinese (Simplified)


jbrooksuk:
Wow - thank you **so much** for these translations. This is incredible!

wfjsw:
so.. you have already solve the messy fonts problem? seems great.

jbrooksuk:
It could be the translations fixing this :) Let's close #447.

wfjsw:
All right I am thinking about if fuzzy 'charset' make that happens...... I will check & fix that soon.

PR #591, opened 2015-04-18
Fix L10n *again*

wfjsw:


jbrooksuk:
Don't worry about StyleCI, we can fix that.

PR #593, opened 2015-04-20
Update Dockerfile for moved app config file

EdeMeijer:
It seems like the app config was moved, but the Dockefile wasn't updated. This makes building work again.


jbrooksuk:
Thanks :smile:

PR #595, opened 2015-04-20
Refactored the way incidents are pulled out of the database on the homepage

Lavoaster:
Rather than run a query for _each_ day, this will pull all incidents for the given range out at once and then add the days with no incidents in it.


jbrooksuk:
Thanks! I've been meaning to do this for a while now :smile:

Lavoaster:
I'll be sending in a PR for this in a few hours. 

The collection needs re-ordering, any days with events will be ordered higher than days without events.

jbrooksuk:
I should revert this commit then?

Lavoaster:
Yeah

Issue #524, opened 2015-03-19
reordering components fails


anarcat:
i can reorder components in the list in `/dashboard/components` but the change never gets saved.

jbrooksuk:
Interesting, what browser, OS and host are you using? Re-ordering of components works on the demo.

anarcat:
chromium on debian jessie, self-hosted.

anarcat:
hmmm... and now it works.

odd.

anarcat:
actually, no, it doesn't work. i can reorder them and it says "components updated" but the changes do not reflect in the user-facing site. when i come back to the dashboard interface, the changes are gone.

jbrooksuk:
Does it work for you in the demo app?

anarcat:
yes, it does (except for #525 of course ;).

jbrooksuk:
Is this still a problem for you? Do you get any errors in the console when you've re-ordered your components? Are you using MySQL?

anarcat:
yes. the console shows a 500 error, and the network inspector in chrome tells me that the resulting page is `Whoops, looks like something went wrong.`

```
POST http://example.net/dashboard/api/components/order 500 (Internal Server Error)
```

i am using mysql.

jbrooksuk:
It's yourself whom has problems with the CSRF tokens, right?

anarcat:
you mean #525? that was me yes. but i didn't have cookies enabled then, now cookies *are* enabled here.

jbrooksuk:
This should've been closed when we merged #599. @anarcat if you're still experiencing this issue, let us know.

anarcat:
i can't confirm this, git pull blew up in my face (see #633).

PR #599, opened 2015-04-22
controller class names made case sensitive for router

naabster:
The internal dashboard API routes (sorting components, changing component status) were giving 500 errors.

Turns out, my server is pretty strict about case sensitive class names.
The class is called **ApiController**, but was referenced as **APIController** in the router.

I did not have this problem on my OSX dev machine, but it might also be the problem of #524


Lavoaster:
Yes, I can confirm this issue and I had to update this manually.

jbrooksuk:
Thanks! :smile:

PR #636, opened 2015-05-15
Add id in langs

LeoColomb:
Fixes crash on install


wfjsw:
@jbrooksuk You may merge this ASAP coz this problem can screw everythings up

Issue #625, opened 2015-05-06
Metric Points are not shown in Graph


Arcticwolf:
Hi,

I have a small problem with saving and showing metric points. I think in the end there is a mix of two problems.

1. When saving new metric points the created_at and updated_at fields of the points are not the same like local time. I added two new points today add 11:40 local time and when I retrieve the metric points via a GET request, I get:

```json
{"metric_points": [
      {
      "id": "1",
      "metric_id": "1",
      "value": "90.000",
      "created_at": "2015-05-06 09:40:43",
      "updated_at": "2015-05-06 09:40:43"
   },
      {
      "id": "2",
      "metric_id": "1",
      "value": "80.000",
      "created_at": "2015-05-06 09:40:56",
      "updated_at": "2015-05-06 09:40:56"
   }
]}
```

The time differs about two hours. So first I thought that I had maybe set the wrong time zone, but I checked it in the configurations and I have set the correct one.

Additionally, why do these values here have a different format than the other created_at and updated_at properties (e.g. of the components and the metrics)?

2. Even though the points don't have exactly the correct time, they are still not shown in the graph:
![figure](https://cloud.githubusercontent.com/assets/673494/7491208/63dd5192-f3e7-11e4-968f-4488d4f5f2fb.png)

About my setup: I use the docker image referenced at the README, so that I could quickstart to use this application.

I think that I just oversee something and that this is the root problem. Maybe you could help me?

Best regards,
Felix

Arcticwolf:
Additional Information: I POST points to the URL {cachetUrl}/api/metrics/1/points with the correct token header and the json message:
{
  "value": 80
}

jbrooksuk:
Is this still an issue on the latest version? We've changed how metric points are calculated, so this may be fixed.

Arcticwolf:
Hi, I've tried it so far only with the docker image. I hope I can manage to try it out with the latest version today. I'll let you know.

Kind regards,
Felix

Arcticwolf:
Ok, I've checked out this project, built the docker file and started the image like described in the README file. Everything starts, the components and metrics I've added are all there. Also the dashboard works. But when I try to access the REST api, I only get 404 Errors. But the paths are exactly the same like the ones I used with the docker image from your server (e.g. localhost/api/components). Are there now different paths?

jbrooksuk:
Ah, if you're using the new master, you'll need to version the API. `localhost/api/v1/components`.

Arcticwolf:
[...]

PR #638, opened 2015-05-16
Make presenters work with the api

joecohens:
This is just a proof of concept of how can we make the presenters work with the API responses. What do you think @jbrooksuk @GrahamCampbell ?

Also this fixes two bugs on this PR https://github.com/cachethq/Cachet/issues/625

### DIFF
--- a/app/Config/Repository.php
+++ b/app/Config/Repository.php
@@ -44,9 +44,9 @@ class Repository
     /**
      * Returns a setting from the database.
      *
-     * @param string $name
-     * @param string $default
-     * @param bool   $checkEnv
+     * @param string      $name
+     * @param string|null $default
+     * @param bool        $checkEnv
      *
      * @return string|null
      */
--- a/app/Config/Repository.php
+++ b/app/Config/Repository.php
@@ -45,11 +45,12 @@ class Repository
      * Returns a setting from the database.
      *
      * @param string $name
+     * @param string $default
      * @param bool   $checkEnv
      *
      * @return string|null
      */
-    public function get($name, $checkEnv = true)
+    public function get($name, $default = null, $checkEnv = true)
     {
         // if we've not loaded the settings, load them now
         if (!$this->settings) {
@@ -65,6 +66,8 @@ class Repository
         if ($checkEnv) {
             return $this->settings[$name] = env(strtoupper($name));
         }
+
+        return $default;
     }
 
     /**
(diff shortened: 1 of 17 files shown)


app/Http/Controllers/Api/AbstractApiController.php
```diff
@@ -35,9 +38,9 @@
     /**
      * The HTTP response data.
      *
-     * @var array
+     * @var mixed
```
GrahamCampbell: mixed should be avoided when possible

jbrooksuk:
Looks great to me!

app/Http/Controllers/Api/AbstractApiController.php
joecohens: Yeah you are right, do you think of a better way to accept the content data?

app/Config/Repository.php
```diff
@@ -45,11 +45,12 @@ public function __construct(Setting $model)
      * Returns a setting from the database.
      *
      * @param string $name
+     * @param string $default
```
GrahamCampbell: string|null

app/Presenters/MetricPointPresenter.php
```diff
@@ -0,0 +1,37 @@
...
+use Jenssegers\Date\Date;
+
+class MetricPointPresenter extends AbstractPresenter
+{
+    use TimestampsTrait;
+
+    /**
+     * Convert the presenter instance to an array.
+     *
+     * @return array
```
GrahamCampbell: can we be more specific here?

app/Presenters/MetricPresenter.php
```diff
@@ -0,0 +1,37 @@
...
+use Jenssegers\Date\Date;
+
+class MetricPresenter extends AbstractPresenter
+{
+    use TimestampsTrait;
+
+    /**
+     * Convert the presenter instance to an array.
+     *
+     * @return array
```
GrahamCampbell: and in all places :)

app/Presenters/MetricPointPresenter.php
joecohens: Can you give me an example?

GrahamCampbell: `string[]`?

joecohens: Lol ok :)

joecohens:
@jbrooksuk All done :) merge at will.

jbrooksuk:
Yes! :D

jbrooksuk:
:dancer:

PR #642, opened 2015-05-18
Toggle component groups

### LINKED ISSUE #572: Make component groups collapsible
https://github.com/cachethq/Cachet/issues/393 introduced component groups, but they aren't collapsible. It would be useful if by default groups are collapsed and only those with issues are shown.

jbrooksuk:
This doesn't use cookies yet, so it's only a session based thing for now.

sta9slwalg


joecohens:
Awesome! :+1:

PR #645, opened 2015-05-19
Better handling of the custom theme settings.

jbrooksuk:
The idea is that we don't need the empty `style` block if the user hasn't setup a theme.

### DIFF
--- a/app/Providers/ViewComposerServiceProvider.php
+++ b/app/Providers/ViewComposerServiceProvider.php
@@ -32,6 +32,7 @@ class ViewComposerServiceProvider extends ServiceProvider
     {
         $this->app->view->composer('*', 'CachetHQ\Cachet\Composers\LoggedUserComposer');
         $this->app->view->composer('index', 'CachetHQ\Cachet\Composers\IndexComposer');
+        $this->app->view->composer('index', 'CachetHQ\Cachet\Composers\ThemeComposer');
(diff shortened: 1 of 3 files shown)


app/Providers/ViewComposerServiceProvider.php
```diff
@@ -31,6 +31,7 @@ public function boot()
     public function register()
     {
         $this->app->view->composer('*', 'CachetHQ\Cachet\Composers\LoggedUserComposer');
+        $this->app->view->composer('*', 'CachetHQ\Cachet\Composers\ThemeComposer');
```
joecohens: Maybe only include it on the status page template?

joecohens:
Nice!!

app/Providers/ViewComposerServiceProvider.php
jbrooksuk: Probably a good idea, for some reason I was thinking of more pages than what we have...

jbrooksuk:
Done :)

app/Providers/ViewComposerServiceProvider.php
joecohens: In that case it could be on the master layout :)

jbrooksuk: Doesn't need to be yet :) In V2 it can be, once we start giving incidents and components their own endpoints.

joecohens: Agreed :)

PR #646, opened 2015-05-19
Fixes #644 - Incident presenter respects the date format.

### LINKED ISSUE #644: Superfluous ordinal indictor timeline incidents
When hovering over the '.. time ago' message in the timeline, there is a superfluous ordinal indicator when using Cachet in another language (in my case Dutch). 

schermafbeelding 2015-04-30 om 14 59 31

It would be nice if this date would be shown in the format as set in the settings in the dashboard.

jbrooksuk:
I've updated the default format to `l jS F Y H:i:s` too.


jbrooksuk:
@joecohens ready.

PR #647, opened 2015-05-19
Allow changing of incident dates. Closes #615

### LINKED ISSUE #615: Allow change the incident time when add/edit  incident
For migrate history data from other statuspage-like system.

jbrooksuk:
Take two.


jbrooksuk:
This does allow you to pick a future date, **but** it won't become scheduled. IMHO this is fine.

PR #650, opened 2015-05-19
Allows ordering of component groups

jbrooksuk:
Essentially it's the same code as we use for ordering of components themselves. Works fine :)

I wasn't planning on doing this feature till V2, but I actually needed it earlier.


jbrooksuk:
@joecohens merging this now. It's the same code as we have for components anyway :smile:

joecohens:
Great :+1:!

Issue #602, opened 2015-04-24
Allow Team Members to Set Incident Visibility


beardedbrew:
Would be nice to be able to set the visibility of an incident to hidden / visible. Multiple use cases -- need to write the incident report and maintain it in the system without having it visible to the public page.

jbrooksuk:
Hmm, I'll add it to V2 list.

PR #651, opened 2015-05-20
Incident visibility

jbrooksuk:
Working on #602 but cannot finish this until we can see if the user is logged in when using the API as we'll need to default to showing public incidents, unless a valid `X-Cachet-Token` is received in which case we'll send the hidden ones too.


jbrooksuk:
@joecohens @GrahamCampbell what do you guys think is the best way of modifying the API so that the right data is returned from the repositories?

jbrooksuk:
Ping @cachethq/owners

PR #659, opened 2015-05-23
Env config setup

jbrooksuk:
The first stage of the setup is now setting the env values.

image

This also calls the `key:generate` command if the `APP_KEY` value is `SomeRandomString` (copied from `.env.example` possibly...).

I did have this setting the database values too, but figured that they'd of had to setup the DB already to run the migrations.


joecohens:
It looks very good @jbrooksuk :)

PR #678, opened 2015-05-28
Started working on a test suite.

jbrooksuk:
Not an urgent feature, but it's good to have, especially for testing the API out.


composer.json
```diff
@@ -38,6 +38,7 @@
         "barryvdh/laravel-cors": "^0.5"
     },
     "require-dev": {
+        "graham-campbell/testbench": "~2.0",
```
GrahamCampbell: 2.1 might be better to start off with

jbrooksuk: Whoops, I was supposed to check if `2.1` was a `5.1` only version?

GrahamCampbell: Also, it would make sense to use `^2.1` to match the other notation in here. :)

jbrooksuk: Updating now.

GrahamCampbell:
> especially for testing the API out.

This will be way better in L5.1. We can actually check error responses are good too.

composer.json
GrahamCampbell: > Whoops, I was supposed to check if 2.1 was a 5.1 only version?

2.1 supports both 5.0 and 5.1. :)

jbrooksuk:
> This will be way better in L5.1. We can actually check error responses are good too.

I'm tempted to upgrade to L5.1 and rebase the repository stuff against it so that I can actually test the API  so that I don't break any output etc in the new version.

jbrooksuk:
> `^2.1`?

Done.

jbrooksuk:
Good to merge? :)

PR #691, opened 2015-06-05
Don't allow php 7 installs

GrahamCampbell:
We can't support php5 and php7 at the same time due to the fact carbon doesn't work. We need to choose a time to drop php5 and move to php7 in the future. I suggest maybe 20 Jun 2016 is a good date to do this. That's when php 5.5 stops reviewing security updates, and php 7.0 should be available on ubuntu 16.04 LTS by then. :)

N.B. I haven't updated the lock file yet since this was done from github's gui.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         }
     ],
     "require": {
-        "php": ">=5.5.9",
+        "php": "^5.5.9",
         "laravel/framework": "5.1.*",
         "cachethq/segment": "^2.1",
         "doctrine/dbal": "^2.5",


jbrooksuk:
20th June 2016 we will officially drop PHP 5 support, I'm happy with that.

GrahamCampbell:
Infact, I've got an idea. Make a 3.0 branch and make it 7.x only in Spring 2016 (or whatever), then actually release 3.0.0 in june 2016 and drop the 2.x version.

GrahamCampbell:
@jbrooksuk is this ok to merge then?

jbrooksuk:
Yeah, go for it :)

PR #702, opened 2015-06-10
Closes #700 - Subscribers can now be turned on and off

### LINKED ISSUE #700: Better subscribers setup
- [ ] Add a setting to disable subscribers.
- [ ] Add a middleware for the subscriber routes which will prevent access if it's not configured properly.
- [ ] Setting of `.env` values from within the dashboard.

jbrooksuk:
We need to add a way to configure the email settings through the dashboard, rather than fiddling with the environment.


app/Providers/ViewComposerServiceProvider.php
```diff
@@ -1,41 +0,0 @@
-<?php
```
jbrooksuk: I'm not sure why this file was still here...

app/Http/Middleware/SubscribersConfigured.php
```diff
@@ -0,0 +1,40 @@
...
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+        $isEnabled = Setting::get('enable_subscribers', false);
+        $mailAddress = env('MAIL_ADDRESS', false);
+        $mailFrom = env('MAIL_NAME', false);
+
+        if (!($isEnabled && $mailAddress && $mailFrom)) {
+            return Redirect::route('status-page');
```
GrahamCampbell: This redirecting on error we're doing a lot is probably a bad practice. Redirecting is telling the client, "this resource has moved here". We should really be showing an error page instead.

jbrooksuk: I was thinking about this as I was doing it... In most of our existing cases, we should show an error page.

With this though, I don't want to *slow down* access to a status page, just because a feature isn't enabled. I don't know, to be honest I'm not fussed either way and I agree that we're doing too many redirects like this.

If we do start going to an error page, we need to make a nice one at least.

jbrooksuk: @GrahamCampbell merging this now but we'll come back to it.

PR #711, opened 2015-06-12
Correct feeds

ldidry:
Small corrections to have valid feeds (validity checked from [W3C validator](https://validator.w3.org/feed/))

### DIFF
--- a/app/Http/Controllers/AtomController.php
+++ b/app/Http/Controllers/AtomController.php
@@ -30,7 +30,7 @@ class AtomController extends AbstractController
         $feed = Feed::make();
         $feed->title = Setting::get('app_name');
         $feed->description = trans('cachet.feed');
-        $feed->link = Setting::get('app_domain');
+        $feed->link = $this->canonicalizeUrl(Setting::get('app_domain'));
 
         $feed->setDateFormat('datetime');
 
@@ -60,9 +60,19 @@ class AtomController extends AbstractController
         $feed->add(
             $incident->name,
             Setting::get('app_name'),
-            Setting::get('app_domain'),
+            $this->canonicalizeUrl(Setting::get('app_domain')),
             $incident->created_at->toAtomString(),
             $incident->message
         );
+    }
+
+    /**
+     * Add a / at the end of an URL to in order to be W3C compliant
+     *
+     * @param string $url
+     */
+    private function canonicalizeUrl($url)
+    {
+        return preg_replace('/([^\/])$/', '$1/', $url);
     }
 }
--- a/app/Http/Controllers/AtomController.php
+++ b/app/Http/Controllers/AtomController.php
@@ -60,7 +60,7 @@ class AtomController extends AbstractController
         $feed->add(
             $incident->name,
             Setting::get('app_name'),
-            $this->canonicalizeUrl(Setting::get('app_domain')),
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/RssController.php
```diff
@@ -60,9 +60,19 @@ private function feedAddItem(&$feed, $incident)
...
             Setting::get('app_name'),
-            Setting::get('app_domain'),
+            $this->canonicalizeUrl(Setting::get('app_domain')).'#'.$incident->id,
             $incident->created_at->toRssString(),
             $incident->message
         );
     }
+
+    /**
+     * Add a / at the end of an URL to in order to be W3C compliant
```
GrahamCampbell: Missing fullstop.

app/Http/Controllers/AtomController.php
```diff
@@ -60,9 +60,19 @@ private function feedAddItem(&$feed, $incident)
...
             Setting::get('app_name'),
-            Setting::get('app_domain'),
+            $this->canonicalizeUrl(Setting::get('app_domain')).'#'.$incident->id,
             $incident->created_at->toAtomString(),
             $incident->message
         );
     }
+
+    /**
+     * Add a / at the end of an URL to in order to be W3C compliant
```
GrahamCampbell: Missing fullstop.

app/Http/Controllers/AtomController.php
```diff
@@ -60,9 +60,19 @@ private function feedAddItem(&$feed, $incident)
...
+     *
+     * @param string $url
+     */
```
GrahamCampbell: You must include a return annotation.

app/Http/Controllers/RssController.php
```diff
@@ -60,9 +60,19 @@ private function feedAddItem(&$feed, $incident)
...
+     *
+     * @param string $url
+     */
```
GrahamCampbell: You must include a return annotation.

app/Http/Controllers/AtomController.php
```diff
@@ -60,9 +60,19 @@ private function feedAddItem(&$feed, $incident)
...
+    private function canonicalizeUrl($url)
+    {
+        return preg_replace('/([^\/])$/', '$1/', $url);
```
GrahamCampbell: This code should not be duplicated between controllers.

ldidry: I thought so. How would you want to do it? An other class and these two classes inheriting from it?

GrahamCampbell: A trait?

ldidry: Where should I put it? Creating a directory Traits in `app/Http`? I'm not a PHP guy, I don't know the conventions.

GrahamCampbell: TBH, the best thing here would probably be to just make a utilities class and stick this as a static method.

GrahamCampbell: Actually, no...

GrahamCampbell: Use the macro method on laravel's Str class. :)

GrahamCampbell:
I'll just make the change. It'll be quicker. :)

GrahamCampbell:
Thank you.

ldidry:
Thank you @GrahamCampbell !

PR #713, opened 2015-06-12
Improvements to the exception handler

GrahamCampbell:
What do you make of this @cachethq/owners?

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -11,11 +11,11 @@
 
 namespace CachetHQ\Cachet\Exceptions;
 
-use CachetHQ\Cachet\Repositories\InvalidModelValidationException;
 use Exception;
-use GrahamCampbell\Exceptions\ExceptionHandler as ExceptionHandler;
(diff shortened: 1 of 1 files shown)


joecohens:
I would still like to have the json response for the API can we use the displayers?

jbrooksuk:
I agree with @joecohens, we could still use the JSON response too.

GrahamCampbell:
We are sending back a json response though?

jbrooksuk:
Oh yeah, merge ahead :)

GrahamCampbell:
We're sending back a response with that error information because it's an http exception, so we keep the exception message in the response.

GrahamCampbell:
Making a json response in that way for our api is bad. We don't even know if the user understands json.

GrahamCampbell:
This way we're manipulating the exception, then letting the exception handler return a response the user understands.

GrahamCampbell:
Though we do have an issue here because if they set their accept header to `*` they will get an html error page, though that's technically the fault of the client.

PR #714, opened 2015-06-13
Fix #712 Hide .bottom-menu-sidebar on small devices

### LINKED ISSUE #712: Make .bottom-menu-sidebar disappear on small devices
`.bottom-menu-sidebar` is hidden on tablets, but not on smartphones. This is annoying since it make it hard to hit the "save" buttons.

I see what I should do to hide it on smartphones, but I was wondering if I need to compile sass files myself and add it to my commit (seems to be no since `public/dist` is in the `.gitignore`) or if it was compiled by some continuous integration robot.

Once I know, I'll do a PR.

ldidry:


jbrooksuk:
Smashing, thanks!

PR #719, opened 2015-06-13
Use HTML version of incident message in atom feed

ldidry:
Atom (not RSS) feeds are able to handle html in entry content, so it
makes sense to put HTML version of incident in it.


jbrooksuk:
Nice idea!

GrahamCampbell:
Need to check if the message is escaped in the view too...

PR #720, opened 2015-06-13
Add piwik stats

ldidry:
Before considering merging this branch, could you tell me how to update lang files? This branch adds two more sentences in the settings.

Do I have to add the sentences manually in the `resources/lang/*/forms.php` files or is there a tool that can do it for me?


jbrooksuk:
@ldidry ideally we should be updating all language files, but I'm sometimes too lazy to do so. The translation functions do fallback to `en` by default, so I don't usually bother.

With that said @GrahamCampbell will be disappointed in me if I start advocating laziness, so if you could copy them across that'd be perfect.

Once you've done that, squash your commits into one and we will merge.

ldidry:
Ok.

Hum, what is that strange language 'en-UD' ?

ldidry:
Please note that I did not put the strings in en-UD.

jbrooksuk:
@ldidry the `en-UD` strings are used by CrowdIn for translations :)

resources/views/partials/analytics.blade.php
```diff
@@ -18,3 +18,19 @@
   _gs('{{ $gosquaredTracking }}');
 </script>
 @endif
+@if($piwikTracking = Setting::get('app_analytics_pi_url'))
```
jbrooksuk: You may want to be checking for both `app_analytics_pi_url` and `app_analytics_pi_siteid` to ensure that Piwik is properly configured.

ldidry: Since there is a default value (1) in the form (see <https://github.com/ldidry/Cachet/commit/78639072c6e921c78ea166464cedcb8cc321bbc1#diff-a45ca592db654261f0e7da8c28be27d2R74>), when you save your setup, you set `app_analytics_pi_siteid`, even if you don't set `app_analytics_pi_url`.

So I think we don't need to check `app_analytics_pi_siteid`, but I can add a check if you think it's relevant.

jbrooksuk: Ok, so as well as setting a default value within the dashboard, you should pass through a default value in the `Setting::get()` method:

```php
Setting::get('app_analytics_pi_siteid', 1)
```

This means that if they remove the default value, it'll still default.

ldidry: Done :smiley:

PR #726, opened 2015-06-16
Use the ::class syntax

jbrooksuk:


GrahamCampbell:
:+1: provided we don't start using it in the config files. I think that stuff looks so silly.

jbrooksuk:
Agreed!

PR #730, opened 2015-06-16
add Korean translation

amondnet:


jbrooksuk:
None of this is actually in Korean?

PR #738, opened 2015-06-17
Respond with the correct responses on error

GrahamCampbell:
These changes make sure we're letting our exception handler sort out the correct http response to send. Also note, our "error pages" are only partially translated and most of our errors didn't even use them.


jbrooksuk:
I'm a firm believer in an API always returning in English.

GrahamCampbell:
> I'm a firm believer in an API always returning in English.

Technically we must respect the language accept header, though I never can be arsed with that. :P

GrahamCampbell:
https://i.starbs.net/wLzz

GrahamCampbell:
oops - missed an import

GrahamCampbell:
I guess if someone tries to use the api with a non-english language accept header, we should return an http not acceptable response.

GrahamCampbell:
> 10.4.7 406 Not Acceptable
> 
> The resource identified by the request is only capable of generating response entities which have content characteristics not acceptable according to the accept headers sent in the request.
> 
> Unless it was a HEAD request, the response SHOULD include an entity containing a list of available entity characteristics and location(s) from which the user or user agent can choose the one most appropriate. The entity format is specified by the media type given in the Content-Type header field. Depending upon the format and the capabilities of the user agent, selection of the most appropriate choice MAY be performed automatically. However, this specification does not define any standard for such automatic selection.
> 
>       Note: HTTP/1.1 servers are allowed to return responses which are
>       not acceptable according to the accept headers sent in the
>       request. In some cases, this may even be preferable to sending a
>       406 response. User agents are encouraged to inspect the headers of
>       an incoming response to determine if it is acceptable.
> If the response could be unacceptable, a user agent SHOULD temporarily stop receipt of more data and query the user for a decision on further actions.

GrahamCampbell:
Though, this line in the spec is interesting: "In some cases, this may even be preferable to sending a
  406 response.". It implies we COULD ignore the accepted language and return English responses.

jbrooksuk:
We *should* ignore it IMHO :)

jbrooksuk:
@GrahamCampbell we need to merge this for tomorrow, right?

GrahamCampbell:
Once I've looked at this again. :)

GrahamCampbell:
Ok, this shizzle should actually work now. :)

GrahamCampbell:
Ideally we should be using the accept middleware to check the client accepts html too for all our other pages, but we can't do that till we've fixed the ajax crap on the setup page. :)

jbrooksuk:
Merge when you're ready :)

GrahamCampbell:
:)

jbrooksuk:
Awesome!

PR #749, opened 2015-06-19
Don't HTML escape custom stylesheet

ldidry:
If you HTML escape the custom stylesheet, you can't have quotes in it.
You need quotes if you want to add something with url('foo') (in a
background-image by exemple).


jbrooksuk:
:beers:

PR #750, opened 2015-06-19
Fix piwik missing slash problem

ldidry:
If there is no trailing slash in the Piwik URL, it won't load piwk.js. I added a slash in the view to fix it.


ldidry:
Sorry, I mixed branchs. I close this PR and will push a clean branch.

PR #751, opened 2015-06-19
Fix piwik missing slash problem

ldidry:
If there is no trailing slash in the Piwik URL, it won't load piwk.js or piwik.php. I added a slash in the view to fix it.

### DIFF
--- a/resources/views/partials/analytics.blade.php
+++ b/resources/views/partials/analytics.blade.php
@@ -18,7 +18,7 @@
   _gs('{{ $gosquaredTracking }}');
 </script>
 @endif
-@if($piwikTracking = Setting::get('app_analytics_pi_url'))
+@if($piwikTracking = rtrim(Setting::get('app_analytics_pi_url'), '/'))
 <!-- Piwik -->
 <script type="text/javascript">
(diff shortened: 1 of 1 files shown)


jbrooksuk:
What happens if the user supplies a `/` at the end of their URL?

jbrooksuk:
Maybe `rtrim($url, '/')`?

ldidry:
Sure, good idea!

GrahamCampbell:
Surely this is still wrong?

GrahamCampbell:
It should be done earlier on when the config is set in the first place. We need to validate it correctly.

ldidry:
Why do you say this is still wrong?

I'm not fond of modifying user's input at registering the config: it might be disturbing to see something different from what you typed earlier. But your advice will prevail, I don't really care.

jbrooksuk:
@willdurand modifying the user input here is important because if we were to use this value anywhere else, we'd need to keep sanitising it. Since this is a setting, we should do that before it goes into the database.

ldidry:
Ok.

jbrooksuk:
Merged and fixed.

PR #753, opened 2015-06-21
Fix floating of buttons on login screen & made border radius equal

fabianlaule:
Currently, there a design issues on the dashboard login screen in Cachet. If you resize your window to a smaller width e.g. for smartphones, the two buttons under the login fields aren't inline any more.

Also, the border radius isn't the same on both buttons (2px vs. 4px).

Before:
cachet-before

After:
cachet-after


TheoBearman:
Thanks for fixing this!

jbrooksuk:
Thanks! :)

PR #779, opened 2015-07-01
Add support for Vagrant/Homestead

chaseconey:
Given the [new changes to Laravel Homestead](http://laravel.com/docs/5.1/homestead#per-project-installation), I thought it might be appropriate to setup a per-project Vagrant environment for out-of-the-box development usage.

Overview of Changes:
* Added new requirement for dev-dependencies (laravel/homestead)
* Modified .env.example to default to mysql and use the built in Homestead mysql database and user information
* Added Homestead-related files
* Modified gitignore to account for vagrant
* Updated README information for Vagrant setup.

I know that there is already a docker setup for the project, but I figure another configuration option wouldn't be a bad thing. I was already using homestead to develop with, and thought I would go ahead and see if y'all might want to go ahead and just add the per-project support.

Let me know if you see a need for any changes.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3,10 +3,10 @@ APP_DEBUG=true
 APP_URL=http://localhost
 APP_KEY=SomeRandomString
 
-DB_DRIVER=sqlite
+DB_DRIVER=mysql
 DB_HOST=localhost
 DB_DATABASE=cachet
-DB_USERNAME=user
+DB_USERNAME=homestead
 DB_PASSWORD=secret
 
 CACHE_DRIVER=file
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,8 @@
     "require-dev": {
         "fzaninotto/faker": "^1.5",
         "graham-campbell/testbench-core": "^1.0",
-        "phpunit/phpunit": "^4.7.6"
+	"phpunit/phpunit": "^4.7.6",
+	"laravel/homestead": "dev-master"
     },
     "autoload": {
         "classmap": [
--- a/composer.json
+++ b/composer.json
@@ -39,8 +39,8 @@
     "require-dev": {
         "fzaninotto/faker": "^1.5",
         "graham-campbell/testbench-core": "^1.0",
-	"phpunit/phpunit": "^4.7.6",
-	"laravel/homestead": "dev-master"
+        "phpunit/phpunit": "^4.7.6",
+        "laravel/homestead": "^2.1"
     },
     "autoload": {
         "classmap": [
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "687f3b651fc7e102f31fd6cc34bd2020",
+    "hash": "52170701af5a0e6e40fbc59832487b3e",
     "packages": [
         {
             "name": "alt-three/segment",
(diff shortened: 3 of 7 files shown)


composer.json
```diff
@@ -39,7 +39,8 @@
     "require-dev": {
         "fzaninotto/faker": "^1.5",
         "graham-campbell/testbench-core": "^1.0",
-        "phpunit/phpunit": "^4.7.6"
+        "phpunit/phpunit": "^4.7.6",
+        "laravel/homestead": "dev-master"
```
GrahamCampbell: This isn't right. This is meant to be a global install so should not go in here. Also, never use dev-master anyway.

composer.lock
```diff
@@ -1209,7 +1209,7 @@
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/guzzle/guzzle/zipball/9c43893a8e580d8401e4210cebe038d850341012",
+                "url": "https://api.github.com/repos/guzzle/guzzle/zipball/66737d2024d49c06bd294f33d60eb369f18b20f2",
```
GrahamCampbell: ?

composer.json
chaseconey: Read through this: http://laravel.com/docs/5.1/homestead#per-project-installation

It is a new feature of Homestead.

I can change it to be a specific version though.

GrahamCampbell: Yes, but global installs are also supported, and that's what we want here.

composer.lock
chaseconey: I think it got updated as it is a dependency of homestead. Not sure how to get around that.

composer.json
chaseconey: I understand that is still an acceptable way to use it, but you can't really distribute anything with that setup. 

If you are very familiar with Laravel and the environment, sure you can set up the global homestead instance and add Cachet and reprovision, etc. Personally I would rather have everything necessary in the project so someone can just run `vagrant up` and not need worry about changing things outside of the project.

If this isn't something you are interested in, I understand. Go ahead and decline the PR and I will just continue using the global installation.

GrahamCampbell: I'm saying all you need to do is not have it in the composer.json, and everything still works...

.env.example
```diff
@@ -3,10 +3,10 @@ APP_DEBUG=true
 APP_URL=http://localhost
 APP_KEY=SomeRandomString
 
-DB_DRIVER=sqlite
+DB_DRIVER=mysql
```
GrahamCampbell: these defaults don't need changing

.env.example
```diff
@@ -3,10 +3,10 @@ APP_DEBUG=true
...
 DB_HOST=localhost
 DB_DATABASE=cachet
-DB_USERNAME=user
+DB_USERNAME=homestead
```
GrahamCampbell: these defaults don't need changing

chaseconey:
This setup will not work without having homestead installed as part of the project. The Vagrantfile and Homestead file require having homestead in the vendor folder.

As far as I understand, you can either:

* not include Homestead in your project and just configure it globally (using the global homestead command and whatnot)
* include homestead as a dependency and have all required configuration files within the project for distribution.

You can't use the global homestead command with project-level configuration items.

.env.example
chaseconey: The reason for changing the defaults is so that it is setup for development by default. In any production environment, these would get overwritten anyway.

Not sure if that makes sense.

We can change the defaults back, and then add more documentation explaining what to change - that would be the other option.

composer.lock
chaseconey: I think this is something that guzzle did one their end. When I am adding this dependency or not, it still updates that tarball url for guzzle. I think they changed the url on their end for that same reference number (notice how the reference hash for that block is the same as before).

chaseconey:
Any ideas how you would like to proceed with this one?

jbrooksuk:
Hmm... Homestead is a **big** dependency for those who don't use it (like myself). @joecohens do you still use Homestead? What do you think?

joecohens:
Yeah I still use it and I love it lol, if we are only using this for dev it would be a good idea.

GrahamCampbell:
I don't use homestead, lol.

jbrooksuk:
Ok, well it'd make development easier for others I guess.

Rebase your changes @chaseconey and we'll merge.

chaseconey:
Man, that composer.lock file cause me all sorts of trouble, but I think everything is in order now.

Definitely give the diff a once over again :).

Thanks

GrahamCampbell:
The composer.lock changes are still wrong.

chaseconey:
Hmm, not sure what I am missing. I have re-generated the lock file with the given composer file and it continually generates that exact same file.

I see the ClassPreloader is 1.4 vs the master 2.0, but I am not sure how it was locked to that version.

chaseconey:
Rebased this as well. Please push so I don't have to rebase again, that composer lock file sucks :)

PR #781, opened 2015-07-02
Dashboard - Active Menu Item CSS Fix

chaseconey:
Very minor fix for the sidebar menu.

**Current behavior:** When clicking on an item in the sidebar, the page reloads and does not indicate which menu item is active in the dashboard sidebar.

**Expected behavior:** When clicking on an item in the sidebar, the page should reload and the sidebar menu should show the currently selected item.

### DIFF
--- a/app/Http/helpers.php
+++ b/app/Http/helpers.php
@@ -34,7 +34,7 @@ if (!function_exists('set_active')) {
 
         $class = implode(' ', $classes);
 
-        return empty($classes) ? '' : "class=\"{$class}\"";
+        return empty($classes) ? '' : "class={$class}";
     }
 }
 


jbrooksuk:
Perfect, thanks!

chaseconey:
@jbrooksuk Hmmm, I just realized that this is actually not fixed for all use cases.

The actual problem is that it is stripping quotes when you use regular `{{ }}` syntax. The real fix would involve using the `{!! !!}` syntax for all uses for `set_active`.

The reason this appeared to work is that it was just echoing class=active into html and the browser is smart enough when there is only one class to actually fix it for you.

Would you like me to go ahead and re-open and fix in the way I described?

jbrooksuk:
Please :)

PR #782, opened 2015-07-02
Dashboard - Active Menu Item CSS Fix v2

chaseconey:
Very minor fix for the sidebar menu (v2).

Relates to #781.

The fix in #781 did fix the issue, but not in the proper way. This should remedy all uses of the `set_active` function.

For more details: #781

### DIFF
--- a/app/Http/helpers.php
+++ b/app/Http/helpers.php
@@ -31,9 +31,9 @@ if (!function_exists('set_active')) {
             $classes[] = $active;
         }
 
-        $class = implode(' ', $classes);
+        $class = e(implode(' ', $classes));
 
-	return empty($classes) ? '' : "class=\"{$class}\"";
+        return empty($classes) ? '' : "class=\"{$class}\"";
     }
 }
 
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Maybe we should `e($class)` to be on the safe side?

jbrooksuk:
Yeah, could be worth doing that @GrahamCampbell.

chaseconey:
Hmm, maybe I am misunderstanding what you are saying @GrahamCampbell. If I were to use the `e()` function, that would be [equivalent to using regular blade braces](http://laravel.com/docs/5.1/blade#displaying-data).

The same double quoting issue will arise.

Lavoaster:
From what I see, there is no user input ever reaching the $classes array param or $active param.

So surely escaping it would be pointless?

GrahamCampbell:
We escape loads of things that we don't think were user provided. That's a good default.

Lavoaster:
True, it is a good default; However, is it overzealous?

chaseconey:
Using `e` helper now. Should be good to go.

Issue #761, opened 2015-06-23
Email subscription not visible


svenmueller:
I enabled "Allow people to signup to email notifications?" in the settings view, but it's not showing up on the status page. Did i miss something?

GrahamCampbell:
Have you cached the config? If so, you need to clear the cache - normally done by just caching it again.

GrahamCampbell:
> in the settings view

Ahhh, missed that...

svenmueller:
any idea how to solve that?

jbrooksuk:
1. Enable the setting in the dashboard.
2. Set the `.env` settings for `MAIL_NAME` and `MAIL_ADDRESS`.
3. Clear the cache `php artisan config:clear`

Done :)

svenmueller:
would be nice to have some info in the README ;) thx will try it out!

jbrooksuk:
We do have documentation at https://docs.cachethq.io/v1.0/docs/subscribers :)

svenmueller:
thx! i just tried tried it...even after setting all required env variables and clearing existing cache configuration (even after reboot) it's not visible (i'm running cachet in the docker container)

jbrooksuk:
Is the Docker build up to date?

jbrooksuk:
We don't do much with Docker, it's left to contributors.

svenmueller:
I'm using the lastest available docker container. I enabled the function in the settings, cleared the cache, defined the required env variables. What else could be the reason for not showing up on the status page? Any idea?

rtrauntvein:
Hi @svenmueller, are you setting those variables via the `docker run -e MAIL_NAME=Email` ?

I am still trying to determine the behavior of the ENV variables in the Docker container version. It seems like there is some dependency on the `.env` file existing per the Cachet design, but it is not present in the current iteration of the Dockerfile and resulting image.  (Some discussion on this in #732)

svenmueller:
@rtrauntvein yes, i think this is the underlying cause (#732) the passed variable (via `-e`) are simply not used as expected

rtrauntvein:
Ok that confirms my suspicions.  

We will need to figure out how the Database variables work currently for the Docker version, and if this can be extended to the other variables.

Otherwise there may be some re-working required to make the docker version actually use a `.env` within the container.

jbrooksuk:
Is there any reason why Docker can't use the `.env` file?

rtrauntvein:
It could use an `.env` file. 

[...]

PR #788, opened 2015-07-07
Use the Cachet .env file inside of docker image

### LINKED ISSUE #732: Docker APP_KEY setting
It doesn't look like [this line](https://github.com/cachethq/Cachet/blob/master/Dockerfile#L27) in the Dockerfile is working as intended.  Nothing is being replaced it looks like.

Would we be able to just pass in an environment variable such as `-e APP_KEY=<key>` when running the container instead of hard coding this?

```
Step 5 : RUN sed -i "s/'key' => '\w.*/'key' => 'f20d3e5ae02125a94bd60203a4edfbde',/" config/app.php &&     grep key config/app.php
 ---> Running in 2bbff0e3eb37
    | This key is used by the Illuminate encrypter service and should be set
    'key' => env('APP_KEY', 'SomeRandomString'),
```

### LINKED ISSUE #786: CentOS 7 + Docker 1.6.2 doesn't work with the documentation
docker exec -it cachet_cachet_1 php artisan migrate --force

                                                                                                               
  [PDOException]                                                                                               
  SQLSTATE[HY000] [2002] Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)  
                                                                                                               

I have tried to build it several times. It stopped working very recently.

rtrauntvein:
Here is my rework of the docker setup to utilize the `.env` file. 

+ It should now support setting all documented Cachet `.env` settings during run time (fixes #732)
+ Working DB Settings in the container (fixes #786)
+ Slight optimization of image layers to speed up re-builds


jbrooksuk:
I can only assume that this works :wink: as I don't know anything about Docker. Perhaps @sammcj could chip in a little here to confirm?

rtrauntvein:
Should also fix #761 - the email variables should be available now.

rtrauntvein:
@jbrooksuk it's mostly just a bunch of ugly bash ;)

 Basically I am just using sed in `entrypoint.sh` to set the variables in the `.env` file at runtime, then starting up the nginx and php processes with supervisord as before.

PR #814, opened 2015-07-24
Add small note to README that a composer install needs to be executed before a vagrant up

andygrunwald:
I wanted to try Cachet and thankful you provide a Vagrant machine.
I cloned the repo and fired a vagrant up. An error appears:

```
$ vagrant up
There was an error loading a Vagrantfile. The file being loaded
and the error message are shown below. This is usually caused by
a syntax error.

Path: /Users/agrunwald/Development/Cachet.git/Vagrantfile
Line number: 11
Message: LoadError: cannot load such file -- /Users/agrunwald/Development/Cachet.git/vendor/laravel/homestead/scripts/homestead.rb
```

After a `composer install` this problem was solved.
But maybe a small note about this step is useful.
This Pull Request is about adding this small note :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Here is a list of things that Cachet is not or does not do:
 
 If you would like to utilize [laravel homestead](http://laravel.com/docs/5.1/homestead), we have a per-project installation available to use for development purposes.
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Ping @jbrooksuk. Looks ok to me.

jbrooksuk:
This is probably quite useful to those who use Docker. Thanks! :)

Issue #805, opened 2015-07-18
Memcached / Redis usability


Herve-M:
Hello,

As I know, Laravel provide memcached / redis for many things like : 
> Session driver
> Cache driver

At first see, the DotEnv don't have memcached / redis variables. 
It will be a nice feature ! 

HerveM

matejkramny:
look in config/database.php

Herve-M:
For session driver, it's this file https://github.com/cachethq/Cachet/blob/master/config/session.php#L28
And for cache, it's this file https://github.com/cachethq/Cachet/blob/master/config/cache.php

But yes, redis config info must be set in config/database.php as said in [Laravel doc](http://laravel.com/docs/5.1/redis)

I think I will do a PR for adding this to .env file etc..

jbrooksuk:
Thanks!

rtrauntvein:
If this gets added into the `.env` file I can also add a Redis container to the docker-compose setup.

jbrooksuk:
Done :)

PR #831, opened 2015-07-28
#805 for setup UI

n0mer:
make 'redis' and 'memcached' options from #805 available during setup

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,3 @@
-.idea
-*.iml
 .env
 node_modules
 phpunit.xml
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,3 +1,5 @@
+.idea
```
jbrooksuk: Please don't commit these files into this .gitignore, they should be globally ignored.

n0mer:
okay done, removed them

.gitignore
n0mer: done

PR #839, opened 2015-07-31
Do not prepend "Status" to the site name

gregkare:
Currently the title of the index is the site name set in the settings is
prepended with "Status". This is confusing because the RSS feeds show
the site name without it (and you probably want to write "Status" in the site
name yourself anyway).


gregkare:
Closing because I write prepend instead of append

PR #840, opened 2015-07-31
Do not append "Status" to the site name

gregkare:
Currently the title of the index is the site name set in the settings is
appended with "Status". This is confusing because the RSS feeds show
the site name without it (and you probably want to write "Status" in the site
name yourself anyway).


GrahamCampbell:
There are probably some other places where this needs removing.

gregkare:
You're right. That's the only one I've found so far (the dashboard's title and RSS feeds don't have it)

GrahamCampbell:
Ping @jbrooksuk.

jbrooksuk:
To be honest, I'm not sure why we even did it in the first place. It causes more confusion than anything.

GrahamCampbell:
Thanks @gregkare. I've removed the other occurrences in 2425682381eaf1482b01e74dcaf508f2aa445a5b.

PR #863, opened 2015-08-03
Fixed login throttling

### LINKED ISSUE #857: Refactor login code
The login code is pretty broken:

1. We validate, then login once. We can reduce the code.
2. Throttling isn't doing anything unless the login validation fails.

GrahamCampbell:
We should throttle ALL requests to the login method. It shouldn't matter if they were successful or not. 

---

Closes #857.


jbrooksuk:
Merge when ready :)

PR #877, opened 2015-08-06
Added GrahamCampbell/Core for easier installs

jbrooksuk:


GrahamCampbell:
I guess this is fine.

jbrooksuk:
Although... We seed with test data and we don't want that.

GrahamCampbell:
Yeh. We're missusing the seeder there.

GrahamCampbell:
The database seeder shouldn't be used for test data.

joecohens:
We could have a if statement to check if it's a demo env and just in that case run the test data.

jbrooksuk:
We now have a command for Seed :)

joecohens:
:)

PR #887, opened 2015-08-07
Added the ability to sort and order API results

### LINKED ISSUE #756: feature: better incident search api
Looking at the [api docs](https://docs.cachethq.io/v1.0/docs/get-incidents) and [the code](https://github.com/cachethq/Cachet/blob/master/app/Http/Controllers/Api/IncidentController.php#L33) for incidents, I can't seem to find a way to allow sorting incidents via `GET /api/v1/incidents` (they default to being sorted by id, ascending). Along with this, the incidents are paginated by `(total rows / num per page) = num pages`.

My use case is that I would like to be able to get the most recent incidents that I haven't yet processed with my poller (ie, those I haven't seen and need to post to our slack channel).

A few solutions:
* allow adding incident id range to the api for `/incidents` instead of page number. This would allow pagination by already seen ids.
* allow sorting incidents by `id DESC`. This would allow me to at least ask for the most recent incidents and work my way backwards.
* add configurable webhooks that allow my site to be triggered on new incident creation. This is the most flexible and would have a lot of benefit for other sites that want the data "in real time."

GrahamCampbell:
Closes #756.

### DIFF
--- a/app/Http/Controllers/Api/AbstractApiController.php
+++ b/app/Http/Controllers/Api/AbstractApiController.php
@@ -159,7 +159,15 @@ abstract class AbstractApiController extends BaseController
             ],
         ];
 
-        return $this->setMetaData($pagination)->setData(AutoPresenter::decorate($paginator->getCollection()))->respond();
+        $items = $paginator->getCollection();
+
+        if ($sortBy = $request->get('sort')) {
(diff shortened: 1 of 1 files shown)


jbrooksuk:
Awesome.

Issue #867, opened 2015-08-03
Avoid Redirect::back


GrahamCampbell:
It is inherently insecure, and should be avoided.

jbrooksuk:
This will require all routes to be named.

GrahamCampbell:
Either that or we just redirect "to".

jbrooksuk:
Named routes would be good. I've started this locally, but in the mean time `to` is good.

jbrooksuk:
Closed by b393557c8a45a45812fda3838227f0ba1275e0b0

PR #888, opened 2015-08-07
Cleaned up routes

jbrooksuk:
Basically naming most of the routes so that we can easily do #867.


jbrooksuk:
@joecohens can you check this please before merge. I think at least every `GET` route (not including the API) is now aliased.

GrahamCampbell:
This needs rebasing.

jbrooksuk:
Done.

joecohens:
@jbrooksuk It seems very good to me :) :+1:

jbrooksuk:
Yay! Now we can sort out #867

PR #893, opened 2015-08-08
[WIP] Application Tokens

jbrooksuk:
This is actually a little bit more complicated than I first thought...

We currently authenticate the user if they use the `X-Cachet-Token` header, except with API tokens there is no assigned user. So what do we do?


jbrooksuk:
Closing this for now, it won't be ready till 3.0.0 anyway.

PR #898, opened 2015-08-13
Completely remove the use of Redirect::back - closes #867

### LINKED ISSUE #867: Avoid Redirect::back

jbrooksuk:
So since Cachet is used as a learning tool by many people, the reason we've done this is because `Redirect::back` uses the `referer` header, which can be hijacked to redirect you somewhere completely different.

By directing users back to the exact route we want, we can avoid this.


GrahamCampbell:
Looks good. :)

jbrooksuk:
Let's merge!

joecohens:
Nice!

PR #902, opened 2015-08-13
Add the component name to incident emails, if applicable. Closes #794

### LINKED ISSUE #794: Send component name within subscriber emails
Hi,

When you create an incident and you have subscribers, they receive an e-mail with only the status and the message. Why not show the component and the name in the e-mail? :-)

jbrooksuk:
@joecohens can you test this please? I don't have mail configured locally at the moment :laughing:

### DIFF
--- a/app/Handlers/Events/SendIncidentEmailNotificationHandler.php
+++ b/app/Handlers/Events/SendIncidentEmailNotificationHandler.php
@@ -69,7 +69,7 @@ class SendIncidentEmailNotificationHandler
                 $mail = [
                     'email'           => $subscriber->email,
                     'subject'         => 'New incident reported.',
-                    'incident'        => $event->incident,
+                    'component_name'  => $event->incident->component->name,
                     'status'          => $data->humanStatus,
                     'htmlContent'     => $data->formattedMessage,
                     'textContent'     => $data->message,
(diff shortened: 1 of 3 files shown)


app/Handlers/Events/SendIncidentEmailNotificationHandler.php
```diff
@@ -69,6 +69,7 @@ public function handle(IncidentHasReportedEvent $event)
                 $mail = [
                     'email'           => $subscriber->email,
                     'subject'         => 'New incident reported.',
+                    'incident'        => $event->incident,
```
joecohens: @jbrooksuk I think this is not going to work since we send the mail to the queue and Laravel has to serialize it (problems). It would be better to decorate the component as well and push it into de view if exists and check if it's null or not. :)

joecohens: BTW I'm getting this error `Trying to get property of non-object (View: /home/vagrant/code/cachet/resources/views/emails/incidents/new-html.blade.php)` exactly on the if $incident->component

jbrooksuk: Hmm, good point @joecohens...

jbrooksuk: Should be fixed now.

jbrooksuk:
My last push didn't fix this issue, but I can't actually test it because I keep getting this:

```
[GuzzleHttp\Exception\ServerException]  
Server error: 500
```

When running `php artisan queue:listen`

n0mer:
what is in your log file?

jbrooksuk:
Ran out of time to mess with this, I'll fix it later if @joecohens doesn't beforehand :)

kersthien:
Hi @jbrooksuk!
in cachet  v 2.4 the templates seem to have a different structure. how can we cleanly reintegrate the components into the templates? thanks in advance for the help!

PR #924, opened 2015-08-22
Added casts properties to all models, re-ordered model properties

### LINKED ISSUE #916: Status changes when editing component
If the status of a component is not 'Operational' and you edit it then save, it will be set to operational unless you remember to change the loaded 'default'.

The 'Status' field in the component edit should reflect the current status, not simply always set to 'Operational' as it is nasssty if you change a detail field for a component and its status changes when you didn't mean it to do so.

jbrooksuk:
Should fix #916

### DIFF
--- a/app/Models/Component.php
+++ b/app/Models/Component.php
@@ -23,15 +23,36 @@ class Component extends Model implements HasPresenter
     use SoftDeletes, ValidatingTrait;
 
     /**
-     * The validation rules.
+     * List of attributes that have default values.
+     *
+     * @var mixed[]
+     */
+    protected $attributes = [
+        'order'       => 0,
+        'group_id'    => 0,
+        'description' => '',
+        'link'        => '',
+    ];
+
+    /**
+     * The attributes that should be casted to native types.
      *
      * @var string[]
      */
-    public $rules = [
-        'name'   => 'required|string',
-        'status' => 'integer|required',
-        'link'   => 'url',
+    protected $casts = [
+        'id'          => 'int',
+        'order'       => 'int',
+        'group_id'    => 'int',
+        'description' => 'string',
+        'link'        => 'string',
     ];
+
+    /**
+     * The attributes that should be mutated to dates.
+     *
+     * @var array
+     */
+    protected $dates = ['deleted_at'];
 
     /**
      * The fillable properties.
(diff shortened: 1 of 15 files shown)


app/Models/Component.php
```diff
@@ -23,17 +23,37 @@ class Component extends Model implements HasPresenter
...
+     * The attributes that should be casted to native types.
      *
      * @var string[]
      */
-    public $rules = [
-        'name'   => 'required|string',
-        'status' => 'integer|required',
-        'link'   => 'url',
+    protected $casts = [
+        'order'       => 'integer',
```
GrahamCampbell: int

app/Models/Component.php
```diff
@@ -23,17 +23,37 @@ class Component extends Model implements HasPresenter
...
+        'group_id'    => 'integer',
```
GrahamCampbell: int

app/Models/Incident.php
```diff
@@ -23,19 +23,30 @@ class Incident extends Model implements HasPresenter
...
-        'message'      => 'required',
+    protected $appends = ['human_status'];
+
+    /**
+     * The attributes that should be casted to native types.
+     *
+     * @var string[]
+     */
+    protected $casts = [
+        'visible'      => 'integer',
```
GrahamCampbell: int

app/Models/Metric.php
```diff
@@ -37,16 +37,16 @@ class Metric extends Model implements HasPresenter
...
      */
-    public $rules = [
-        'name'          => 'required',
-        'suffix'        => 'required',
+    protected $casts = [
+        'name'          => 'string',
         'display_chart' => 'boolean',
-        'default_value' => 'numeric',
-        'places'        => 'numeric|min:0|max:4',
+        'default_value' => 'integer',
```
GrahamCampbell: in all places, please use the correct types

app/Models/User.php
```diff
@@ -25,17 +25,27 @@ class User extends Model implements AuthenticatableContract, CanResetPasswordCon
...
-    public $rules = [
-        'username' => ['required', 'regex:/\A(?!.*[:;]-\))[ -~]+\z/'],
-        'email'    => 'required|email',
-        'password' => 'required',
+    protected $casts = [
+        'username'          => 'string',
+        'email'             => 'string',
+        'google_2fa_secret' => 'string',
+        'api_key'           => 'string',
+        'active'            => 'boolean',
```
GrahamCampbell: bool

app/Models/Metric.php
jbrooksuk: I didn't think Laravel supported short versions?

GrahamCampbell: it does

jbrooksuk: Since when? :)

GrahamCampbell:
Shouldn't this go to 1.2 since it's a bug fix?

jbrooksuk:
1.2.1 I guess...

jbrooksuk:
Rebased and tests do pass.

GrahamCampbell:
> Rebased and tests do pass.

Still failing?

jbrooksuk:
Hmm, leave it with me.

jbrooksuk:
This actually fixes the tests too, since they weren't comparing integers.

joecohens:
:+1:

PR #930, opened 2015-08-28
Route 'dashboard.components' not found

mbabker:
Updated to reference `dashboard.components.index` route instead.


jbrooksuk:
:beers:

PR #935, opened 2015-09-04
Fixes #934 - Scheduled items now drop into the right days

### LINKED ISSUE #934: Scheduled event displayed on date of creation, not date of actual event
In version 1.2 maintenance appears in event list on the date it was added, not on the date of event itself.
We scheduled the event two weeks in advance and today users need to scroll back in history to find today's maintenance event. A bit confusing...

jbrooksuk:
@GrahamCampbell this will need copying over to `master` :)


GrahamCampbell:
We don't need to copy to master, we only need to copy to 1.2 if it went to the master. Branch merges from 1.2 -> master are usually easy/

PR #939, opened 2015-09-10
add UTC time zone as an option on Application Setup page

andrewgoktepe:
Adds a UTC option to the drop-down list for time zone


jbrooksuk:
:+1:

PR #952, opened 2015-09-18
Code tidy

jbrooksuk:


app/Providers/ConsoleServiceProvider.php
```diff
@@ -23,13 +22,6 @@ class ConsoleServiceProvider extends ServiceProvider
      */
     public function register()
     {
-        $this->app->singleton('CachetHQ\Cachet\Console\Commands\FixPermissionsCommand', function ($app) {
```
GrahamCampbell: Right, so shouldn't we just delete this whole service provider then?

jbrooksuk: Probably, yeah.

PR #973, opened 2015-10-01
Alert improvements

billmn:
Replaced JS alert with SweetAlert (http://t4t5.github.io/sweetalert/)

schermata 2015-10-01 alle 23 29 16


jbrooksuk:
This is nice, thanks!

jbrooksuk:
I'll pull this down later and check it :)

resources/assets/js/app.js
```diff
@@ -351,3 +361,16 @@ $(function() {
             .addClass("active");
     }
 });
+
+function askConfirmation(callback) {
+    swal({
+        type: "warning",
+        title: "Confirm your action",
+        text: "Are you sure you want to do this ?",
```
wngravette: you'll have to zap that space before the ?

GrahamCampbell: good catch

billmn: Done.

jbrooksuk:
Perfect!

GrahamCampbell:
Feels like the yes and cancel buttons are the wrong way around.

jbrooksuk:
They're right according to `confirm()`?

GrahamCampbell:
Just looks wrong to me.

GrahamCampbell:
Maybe it's just me.

jbrooksuk:
I think it may be the colours.

GrahamCampbell:
Maybe.

billmn:
Colors and buttons text can be customized if you want. 

We maintain the standard JS confirm prompt?

PR #984, opened 2015-10-02
per_page attribute in API responses should be an integer

AntoineAugusti:
It was previously a string because of `Binput::get`. I don't know how to reflect this in the API documentation


jbrooksuk:
I'll update the docs later :) thanks!

PR #986, opened 2015-10-02
changed german translation to a more sensible one

manuelgu:
The original sentence sounded a bit weird


GrahamCampbell:
Thanks. I'll assume this is right since I don't speak German. :)

jbrooksuk:
That's what I'm going on too!

ApfelUser:
Well. I can confirm this is correct :smile:

Issue #974, opened 2015-10-01
Use translations on JS labels


billmn:
I noticed that JS doesn't use translated labels, like here: https://github.com/cachethq/Cachet/blob/eadedeecbf4c43256c883efb98e9a1dd2db5f906/resources/assets/js/app.js#L172

What about create a specific lang file that contains labels for JS?
This array can be converted into JSON and used in a JS variable.

Suggestions?

jbrooksuk:
Yeah, it's always annoyed us that we've never translated them. Alternatively we could use a data attribute with the translation string somewhere?

billmn:
@jbrooksuk I'm not sure I understand what you intend to do, can you give me an example?

An example of my proposal:

```html
    <script type="text/javascript">
        var Global = {};
        Global.locale = '{{ Setting::get('app_locale') }}';

        // Extract translations from a Lang file
        Global.transl = {!! json_encode(trans('auth')) !!};

        // Result: "These credentials do not match our records."
        console.log(Global.transl.failed);
    </script>
```

AntoineAugusti:
Could https://github.com/laracasts/PHP-Vars-To-Js-Transformer be useful?

jbrooksuk:
Ah, I see. That's a good idea @billmn.

jbrooksuk:
@AntoineAugusti I'm really sorry that I didn't reply! I think that yes, it'd be really useful!

jbrooksuk:
Thank you for your input on Cachet 2.x. We are shifting our attention and resources to Cachet 3.x and will no longer be supporting the 2.x version. If your feedback or issue is relevant to the 3.x series, we encourage you to engage with the new branch.

For more information on the Cachet rebuild and our plans for 3.x, you can read the announcement [here](https://github.com/CachetHQ/Cachet/discussions/4342).

We appreciate your understanding and look forward to your contributions to the new version.

Issue #832, opened 2015-07-28
Improve translations handling


jbrooksuk:
Whilst CrowdIn really helps out with translations, if we ever have to manually add a translation string, we have to add it to **every** file.

I get annoyed with this myself (and sometimes just don't do it) - so I can imagine developers/translators getting frustrated when adding a new feature.

jbrooksuk:
@cachethq/owners any ideas on how to do this?

AntoineAugusti:
This package seems interesting and provide useful commands to at least add new strings everywhere

jbrooksuk:
Yeah, we're going to need to modify the package though, so that it uses short-array syntax etc.

GrahamCampbell:
And fix the wrong indentation.

jbrooksuk:
No longer accepting PR's for langs except English. Only CrowdIn. I'll keep 'em in sync.

PR #987, opened 2015-10-03
[Proposal] Javascript translations

AntoineAugusti:
Proof of concept for https://github.com/cachethq/Cachet/issues/974. This is also related to https://github.com/cachethq/Cachet/issues/832

I've used https://github.com/PhiloNL/laravel-translate to add a new translation, because I was too damn lazy to open every file. The thing is that this tool messes up the formatting of the translation files... But StyleCI can automatically fix this.

So, before replacing every hardcoded notification in JavaScript with translation keys, I want to know what you think about this. Cc @jbrooksuk @GrahamCampbell 

*Note*: JS was not compiled and commits have not been squashed so that you will see which steps I've done


joecohens:
I like the idea! The only thing that bugs me a bit it's loading all does translations in the DOM directly, maybe we can code a command to publish the translation files to json when we have a change so we will have a /public/build/dist/js/langs/dashboad.json in that way we can import them individually per page.

AntoineAugusti:
@joecohens My ideas were: creating a specific file for JS translations or filtering keys when getting translations values, so that the DOM will not be too filled up as you say.

joecohens:
I know right now we only have like 4 or 5 javascript messages to translate, but I think it might be a good idea to give javascript access to the whole php translation files. We could also load them dynamically depending on the configured locale like `/public/build/dist/js/langs/{es}/dashboad.json`.

GrahamCampbell:
Please:

1. Fix the conflicts
2. Revert all changes to the formatting of the lang files

jbrooksuk:
We do need a better way of managing translations... It's a nightmare.

ApfelUser:
Indeed. It would be extremely helpful if there would be some automated commit bot from crowdin. This should be actually also doable, as they have an [API](https://crowdin.com/page/api) and also some [API librarys](https://github.com/crowdin). There is also an issue on their [uservoice](https://crowdin.uservoice.com/forums/31787-collaborative-translation-tool/suggestions/5435777-integrate-with-source-control) forum for this.

jbrooksuk:
I've also spoken directly to them about this.

AntoineAugusti:
I am happy to work on this feature, but I don't want to manage translation by hands. What do you suggest?

I'd say:
- use https://github.com/PhiloNL/laravel-translate
- apply patches from StyleCI on PRs

But we will loose comments in language files if we do this.

GrahamCampbell:
> apply patches from StyleCI on PRs

StyleCI doesn't fix the extra alignment yet though.

PR #989, opened 2015-10-03
Some improvements on Settings controller

billmn:
- Redirect back instead of always redirect to setup page
- Use URL generation by route instead of specify it manually
- More expressive image check


jbrooksuk:
- We actually avoid doing `Redirect::back` purposefully because it's insecure. A MITM attack could redirect you somewhere else.

billmn:
What about this alternative?

jbrooksuk:
Still open to attack. Hmm, maybe we could store a session value with the last route when in the dashboard? @GrahamCampbell?

billmn:
This require to reflash session on reload. Anyway I think is not the right thing to redirect always to setup section

billmn:
@jbrooksuk I've changed the code to use session like your suggestion. Take a look ;-)

app/Http/Controllers/Dashboard/SettingsController.php
```diff
@@ -68,6 +68,8 @@ public function showSetupView()
     {
         $this->subMenu['setup']['active'] = true;
 
+        session()->flash('redirect_to', $this->subMenu['setup']['url']);
```
jbrooksuk: @GrahamCampbell what do you think, should we be using the Session facade here?

GrahamCampbell: meh, doesn't really matter - if you want to, go for it ;)

GrahamCampbell: nice to be consistent with what ever we use else where

billmn: @GrahamCampbell make sense. I'll use facade.

billmn:
Now with facade

jbrooksuk:
Squash into one commit and we'll get it merged! :)

GrahamCampbell:
> Squash into one commit and we'll get it merged! :)

I can do that. :)

GrahamCampbell:
Squashed and merged. Thank you @billmn.

GrahamCampbell:
Missed the `#` in the commit so this PR didn't auto-close. Oops...

PR #990, opened 2015-10-03
Command Tests

### LINKED ISSUE #942: Add Command/Event Tests Like StyleCI
This stuff is brilliant for checking for typos and missing assignments, and will also check the handlers can be instantiated, but doesn't test actual handler execution.

I'll get this done asap.

jbrooksuk:
This will close #942.

### DIFF
--- a/app/Commands/Component/AddComponentCommand.php
+++ b/app/Commands/Component/AddComponentCommand.php
@@ -11,7 +11,7 @@
 
 namespace CachetHQ\Cachet\Commands\Component;
 
-class AddComponentCommand
+final class AddComponentCommand
 {
     /**
      * The component name.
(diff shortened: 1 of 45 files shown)


GrahamCampbell:
Shall I make an Alt Three TestBench pacage so we don't have to duplicate the bas stuff?

jbrooksuk:
That'd be super useful since both StyleCI and Cachet use the exact same abstract code.

jbrooksuk:
Merging this now then and we'll just refactor later on.

GrahamCampbell:
Sure.

GrahamCampbell:
Maybe now is a good time to tag v2.0.0-beta2?

PR #995, opened 2015-10-05
[Proposal] Added background color to sortable list

billmn:
Now sortable list doesn't have a background color and during re-order seems like this:
no-background

I think that with this little change, items are more visible during reorder and can be useful to see "move" cursor on drag-handle hover.


jbrooksuk:
I don't remember much about this, but is it possible to wrap the dragged item in a border?

GrahamCampbell:
> I don't remember much about this, but is it possible to wrap the dragged item in a border?

:+1:

billmn:
Sure, I'll add it.

Grey background (like sidebar) is ok for you? It can also be white. 

The important thing is that have a background color and not be transparent to increase readability

jbrooksuk:
> Grey background (like sidebar) is ok for you? It can also be white.

Yeah, that's absolutely fine.

billmn:
Done ;)

jbrooksuk:
Sexy. Thanks for this!

PR #1008, opened 2015-10-06
Use class name resolution via ::class

billmn:


GrahamCampbell:
No point.

app/Console/Kernel.php
```diff
@@ -22,7 +22,7 @@ class Kernel extends ConsoleKernel
      * @var array
      */
     protected $commands = [
-        'CachetHQ\Cachet\Console\Commands\DemoSeederCommand',
+        \CachetHQ\Cachet\Console\Commands\DemoSeederCommand::class,
```
GrahamCampbell: maybe this one can stay, but it would make sense to import it

billmn:
The sense of this PR is to use a single "method" to specify class names.

Some parts of Cachet already use this (like models) and can also be useful to keep configuration files updated with Laravel.

GrahamCampbell:
It's not consistent though. We NEVER reference classes using a leading slash. We always import them, and we're not going to import like 50 classes to use ::class.

GrahamCampbell:
This is only useful when referencing our own code at best.

config/app.php
```diff
@@ -122,56 +122,56 @@
...
-        'Illuminate\Pagination\PaginationServiceProvider',
-        'Illuminate\Pipeline\PipelineServiceProvider',
-        'Illuminate\Queue\QueueServiceProvider',
-        'Illuminate\Redis\RedisServiceProvider',
-        'Illuminate\Auth\Passwords\PasswordResetServiceProvider',
-        'Illuminate\Session\SessionServiceProvider',
-        'Illuminate\Translation\TranslationServiceProvider',
-        'Illuminate\Validation\ValidationServiceProvider',
-        'Illuminate\View\ViewServiceProvider',
+        Illuminate\Foundation\Providers\ArtisanServiceProvider::class,
```
GrahamCampbell: we're never going to do this in config files - i think it's a stupid idea

billmn:
Ok make sense, thank you

app/Console/Kernel.php
GrahamCampbell: https://github.com/cachethq/Cachet/commit/814fcd844c30aa2bfc1f3bb96812a3f678a49450

config/app.php
billmn: ok I got it

PR #1012, opened 2015-10-07
[WIP] Started working on API timezone support

jbrooksuk:
I feel like I'm missing something pretty obvious here.

### DIFF
--- a/app/Http/Middleware/Timezone.php
+++ b/app/Http/Middleware/Timezone.php
@@ -1,0 +1,35 @@
+<?php
+
+/*
+ * This file is part of Cachet.
+ *
+ * (c) Alt Three Services Limited
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace CachetHQ\Cachet\Http\Middleware;
+
+use Closure;
+
+class Timezone
+{
+    /**
+     * Handle an incoming request.
+     *
+     * @param \Illuminate\Http\Request $request
+     * @param \Closure                 $next
+     * @param string                   $type
+     *
+     * @return mixed
+     */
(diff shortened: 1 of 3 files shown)


app/Http/Middleware/Timezone.php
```diff
@@ -0,0 +1,35 @@
...
+     * @param \Closure                 $next
+     * @param string                   $type
+     *
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+        if ($tz = $request->header('Time-Zone')) {
+            app('config')->set('app.timezone', $tz);
+        }
```
billmn: This could be <code>Setting::set('app_timezone', $tz);</code> ?

billmn:
I've tested your code locally and yes, you need to:

1. Use <code>Setting::set('app_timezone', $tz)</code>
2. Make this change to apply timezone on attributes when there is a pagination (https://github.com/cachethq/Cachet/pull/1017)

jbrooksuk:
Point 1 is now done. Not sure how best to do point 2.

jbrooksuk:
@billmn with the paginator stuff, do you think we should loop the results, then modify any dates:

- `deleted_at`
- `created_at`
- `updated_at`
- `scheduled_at`

And convert the timezone?

billmn:
I have tested <code>/api/v1/incidents</code> and already works:

schermata 2015-10-08 alle 09 36 29
schermata 2015-10-08 alle 09 36 44

jbrooksuk:
Nice! So actually, this is done? I'm always dubious about timezone code so I want to make sure this is right.

jbrooksuk:
I need to add a test for the `Timezone` middleware really. CC @GrahamCampbell

PR #1015, opened 2015-10-07
Autopresenter Updates

GrahamCampbell:


GrahamCampbell:
NB, the facade changes are just because there's little point in injecting - it just means we write more code.

PR #1019, opened 2015-10-07
Fix problems with the incident add/edit form

mrbase:
Removing the `incident` array from the form and the js selector.

Fixes:
* Missing `component_status` dropdown when adding incident with component set.
* Failing updates due to controller code not using the `incident` form array


GrahamCampbell:
Thanks. Merging this, but pinging @jbrooksuk for additional review. :)

jbrooksuk:
I've not tried it out, but it looks fine! :+1:

jbrooksuk:
Just tried it out, worked nicely!

PR #1024, opened 2015-10-08
Restored Status Page localization

billmn:
Has been removed here: https://github.com/cachethq/Cachet/commit/c2c815ab14a7664fb7e56e4319377a175fd878cf#diff-d02b216c1d1bdf67feaa85180f7b2a84L30

But I think it's been removed by mistake


jbrooksuk:
Whoops. Good spot!

PR #1029, opened 2015-10-11
Use warning rather than error

jbrooksuk:
image

Red is quite serious, maybe we could make it yellow? Also it matches the favicon when there is an issue.

### DIFF
--- a/app/Composers/StatusPageComposer.php
+++ b/app/Composers/StatusPageComposer.php
@@ -30,7 +30,7 @@ class StatusPageComposer
     {
         // Default data
         $withData = [
-            'systemStatus'  => 'danger',
+            'systemStatus'  => 'warning',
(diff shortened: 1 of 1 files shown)


jbrooksuk:
Ping @cachethq/owners.

GrahamCampbell:
Maybe `info` instead?

GrahamCampbell:
I think yellow looks strange.

jbrooksuk:
Will try `info`.

jbrooksuk:
image

Doesn't look bad.

GrahamCampbell:
Yeh, I prefer that. :)

jbrooksuk:
Me too, I'll update and merge. Feels better than a red block.

joecohens:
:+1:

PR #1032, opened 2015-10-12
Test setting timezone on the api for metric points

joecohens:
/cc @jbrooksuk

### DIFF
--- a/app/Handlers/Commands/Metric/AddMetricPointCommandHandler.php
+++ b/app/Handlers/Commands/Metric/AddMetricPointCommandHandler.php
@@ -36,7 +36,7 @@ class AddMetricPointCommandHandler
         ];
 
         if ($createdAt) {
-            $data['created_at'] = Carbon::createFromFormat('U', $createdAt)->format('Y-m-d H:i:s');
+            $data['created_at'] = Carbon::createFromFormat('U', $createdAt, config('cachet.timezone'))->format('Y-m-d H:i:s');
         }
 
         $metricPoint = MetricPoint::create($data);
--- a/app/Handlers/Commands/Metric/AddMetricPointCommandHandler.php
+++ b/app/Handlers/Commands/Metric/AddMetricPointCommandHandler.php
@@ -15,9 +15,29 @@ use CachetHQ\Cachet\Commands\Metric\AddMetricPointCommand;
 use CachetHQ\Cachet\Events\Metric\MetricPointWasAddedEvent;
 use CachetHQ\Cachet\Models\MetricPoint;
 use Carbon\Carbon;
+use Illuminate\Contracts\Config\Repository as Config;
 
 class AddMetricPointCommandHandler
 {
+    /**
+     * The illuminate config repository instance.
+     *
+     * @return \Illuminate\Contracts\Config\Repository
+     */
+    protected $config;
+
+    /**
+     * Create a new add metric point command handler instance.
+     *
+     * @param \Illuminate\Contracts\Config\Repository $config;
+     *
+     * @return void
+     */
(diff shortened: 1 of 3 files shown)


app/Handlers/Commands/Metric/AddMetricPointCommandHandler.php
```diff
@@ -36,7 +36,7 @@ public function handle(AddMetricPointCommand $command)
         ];
 
         if ($createdAt) {
-            $data['created_at'] = Carbon::createFromFormat('U', $createdAt)->format('Y-m-d H:i:s');
+            $data['created_at'] = Carbon::createFromFormat('U', $createdAt, config('cachet.timezone'))->format('Y-m-d H:i:s');
```
GrahamCampbell: handlers shouldn't ever be calling these global functions

GrahamCampbell: that stuff is ok in controllers, but i always feel handers need to be framework agnostic, especially when we come to actually test the handlers

GrahamCampbell: i guess it would be "ok" to inject the config contract, but best would be to inject the actual config value, by registering a custom handler binding

joecohens: Yeah that's true I need to inject the config repository

GrahamCampbell: are there any other places where we've been naughty calling global horrors?

joecohens: I think so

GrahamCampbell: i think we can get away with global functions in controllers, views, and middleware, but we shouldn't use them anywhere else at all

GrahamCampbell: basically, use it in the stuff that we'd never want to separate, or unit test

do you agree?

joecohens: Totally

jbrooksuk: Agreed

GrahamCampbell:
Any movement here?

joecohens:
@cachethq/owners Injected the config repository

jbrooksuk:
:+1:

GrahamCampbell:
This wasn;t quite what I meant.

GrahamCampbell:
I meant actually inject the config values, rather than the config repository,

GrahamCampbell:
https://github.com/AltThree/StyleCI/blob/master/app/Providers/AppServiceProvider.php#L68

GrahamCampbell:
Example binding where we inject the config values rather than the config repo.

joecohens:
Oh! Ok ok but that means we need to register an instance of the handler in a service provider, don't we?

GrahamCampbell:
Yeh, we'll have to add a binding.

GrahamCampbell:
I really hate code that has config reading in it. Feels really dirty.

GrahamCampbell:
I think if we are reading the config, do it at the service provider level, then behond that, you can continue writing code that's not bound to the config.

jbrooksuk:
@joecohens once you've got @GrahamCampbell's changes done, just go ahead and merge! :)

GrahamCampbell:
Got a better way to sort this. Will PR in a sec.

Issue #1034, opened 2015-10-14
Language and Date


Lumiahna:
Hi,
I'm having an issue with my status page.
I set the language into French, but the date seems to be display in English ( day and month are in English instead of French) .
I didn't find a file to edit and to translate that,
How can I do that ?
Thx for your awesome project (: .

Conf: Debian 7 / Apache

jbrooksuk:
![image](https://cloud.githubusercontent.com/assets/246103/10485037/5b846a4a-7280-11e5-878c-0e6b69d598c0.png)

Go to your Settings on the dashboard and change the date format.

Lumiahna:
Hi,
I did it 
Set Timezone: Paris
Date Format: l j F Y
Incident Timestamp Format: l j F Y H:i:s
Site language: English
But it seems to be display in English, I mean: day ( monday, tuesday, ...) and months .
Check it on http://status.2fy.eu
thx

jbrooksuk:
Oh I see. Lemme investigate further.

jbrooksuk:
@joecohens looking at this, it'd require rewriting most of the date stuff and make use of `Carbon::setLocale` and also `formatLocalized`... :see_no_evil:

joecohens:
Mmm this is odd. We can't use Carbon set locale because Carbon use the Datetime locale that means people will have to install the langs on their own OS. That's why we are using https://github.com/jenssegers/date package so we won't force people to install langs on their OS. Let me check what's happening here.

Lumiahna:
My vps is in French, and I misstype Site language: , ofc It was French instead of English.
But day/month are in English instead of French ...

PR #1037, opened 2015-10-14
Set date lang on the localized middleware

joecohens:
Fixes https://github.com/cachethq/Cachet/issues/1034

/cc @jbrooksuk


jbrooksuk:
That's way easier than what I had in mind :laughing:

jbrooksuk:
Hmm, this still isn't translating dates?

image

PR #1039, opened 2015-10-15
Add `composer test` script

MaxGfeller:
Composer supports defining custom scripts. Major PHP projects started adapting this and are now all defining their own `test` scripts. Here are some examples:

- https://github.com/composer/composer/blob/master/composer.json#L63-L65
- https://github.com/Seldaek/monolog/blob/master/composer.json#L60-L63

The advantage of this is pretty obvious; you don't have to have a global phpunit/whatever setup and can just run `composer test` for each library that you want to start working with.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -81,6 +81,9 @@
         "post-create-project-cmd": [
             "php -r \"copy('.env.example', '.env');\"",
             "php artisan key:generate"
+        ],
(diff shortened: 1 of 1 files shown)


jbrooksuk:
Doesn't hurt, thanks! :)

Issue #1030, opened 2015-10-11
Dashboard forms


jbrooksuk:
The forms in the dashboard could do with a bit of TLC, making them consistent, cleaner and easier to use.

mmollick:
I'm gonna give this a crack, from my initial comb through the dashboard I noticed a few things:

* The Application Setup page could use some compartmentalization, I'm thinking this page could be split up visually into subsections; Analytics, Localization, and Application settings.
* The optional indicator for fields could me moved next to the form label, perhaps as a bootstrap `.label`.
* Checkboxes could be condensed down into one line and make use of bootstrap's `.checkbox` class.
* The status selection on the Add an incident page could be represented as a select field (This wouldn't allow for the icons however), not sure it's appropriate for this field though...

If anyone has any additional suggestions I'm all ears

jbrooksuk:
That'd be awesome @mmollick!

jbrooksuk:
Closed by 974fd5015d98cacdc9b75c37384b0dfc9f783fd0

PR #1048, opened 2015-10-21
Cleaning up dashboard forms

mmollick:
This is some general TLC for the dashboard forms #1030
* The existing 'Application Setup' page has been split into three pages; Application Setup, Analytics, and Localization.
* Checkboxes are now all inline with their labels.
* None of the checkboxes have an optional indicator; they're inherently optional anyways.
* Optional indicators are now displayed to the right of the labels.


jbrooksuk:
Awesome, thank you! :heart:

PR #1070, opened 2015-10-28
Improved and fixed the report incident command handler

GrahamCampbell:
Hopefully this works. I have no idea. Let's see what travis makes of it all.

### DIFF
--- a/app/Handlers/Commands/Incident/ReportIncidentCommandHandler.php
+++ b/app/Handlers/Commands/Incident/ReportIncidentCommandHandler.php
@@ -29,23 +29,28 @@ class ReportIncidentCommandHandler
      */
     public function handle(ReportIncidentCommand $command)
     {
-        $incident = Incident::create([
-            'name'         => $command->name,
-            'status'       => $command->status,
-            'message'      => $command->message,
-            'visible'      => $command->visible,
-            'component_id' => $command->component_id,
-        ]);
+        $data = [
+            'name'    => $command->name,
+            'status'  => $command->status,
+            'message' => $command->message,
+            'visible' => $command->visible,
+        ];
+
+        // Link with the component.
+        if ($command->component_id) {
+            $data['component_id'] = $command->component_id;
+        }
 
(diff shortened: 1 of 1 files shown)


app/Handlers/Commands/Incident/ReportIncidentCommandHandler.php
```diff
@@ -29,24 +29,29 @@ class ReportIncidentCommandHandler
...
+        ];
+
+        // Link with the component.
+        if ($command->component_id) {
+            $data['component_id'] = $command->component_id;
+        }
 
         // The incident occurred at a different time.
         if ($command->incident_date) {
             $incidentDate = Date::createFromFormat('d/m/Y H:i', $command->incident_date, config('cachet.timezone'))->setTimezone(Config::get('app.timezone'));
```
GrahamCampbell: also, we shouldn't ever be using the config helper here

GrahamCampbell: all "foundation" global functions should be banned unless they're in controllers

GrahamCampbell: our handler classes should be framework agnostic

jbrooksuk: So pass it through to the command?

GrahamCampbell: no

GrahamCampbell: we need to bind the handler to the ioc and inject the config value at that point

GrahamCampbell: we shouldn't abuse commands to pass config through

GrahamCampbell:
Travis passing. Whoop whoop.

jbrooksuk:
:tada:

PR #1101, opened 2015-11-04
Display the current version in the settings sidebar. Closes #1061

### LINKED ISSUE #1061: Version of existing instance
Is there a way to find out the version number of the currently running Cachet instance?

jbrooksuk:
We don't really have anywhere else to put the version for now, so this will do, plus it's out of the way.

image

I've chosen to create a `VERSION` file in the root because this makes it easier for people to open the repo and immediately see the value. The defined version is a constant because this will make it easier to check versions in the future.

Ping @cachethq/owners


TheoBearman:
Personally, I would move this to the bottom of the dashboard navigation bar.

jbrooksuk:
> Personally, I would move this to the bottom of the dashboard navigation bar.

I thought about that, but we always have issues with shallow screens. It'd probably move there once we have a redesigned dashboard.

TheoBearman:
Then what about the footer on the status page? I just think it looks too out of place on the settings page.

jbrooksuk:
> Then what about the footer on the status page?

Nope, I wouldn't want to publicly expose the version of Cachet like that.

TheoBearman:
In that case, I guess where you have put it is the best place then.

jbrooksuk:
> In that case, I guess where you have put it is the best place then.

Well, it's the best we can do for now.

VERSION
```diff
@@ -0,0 +1 @@
+2.0.0-RC1
```
GrahamCampbell: this is wrong

GrahamCampbell: should be `2.0.0-dev`

GrahamCampbell: we need to update it when we actually do a tag, then again straight after to `-dev`

app/Providers/AppServiceProvider.php
```diff
@@ -28,6 +28,8 @@ public function boot(Dispatcher $dispatcher)
             return Dispatcher::simpleMapping($command, 'CachetHQ\Cachet', 'CachetHQ\Cachet\Handlers');
         });
 
+        define('CACHET_VERSION', trim(file_get_contents($this->app->basePath().'/VERSION')));
```
GrahamCampbell: this will fail in some cases

GrahamCampbell: for example, if we run the tests more than once

GrahamCampbell: oh wow, yeh, already failing

GrahamCampbell: best option here is to set this constant in the early bootstrap process

GrahamCampbell: https://github.com/cachethq/Cachet/blob/master/bootstrap/autoload.php

GrahamCampbell: move this code into there

jbrooksuk:
Thanks for finishing this @GrahamCampbell :)

GrahamCampbell:
No problem. :)

joecohens:
:heart_eyes:

jbrooksuk:
We just need to remember to update this file now.

PR #1107, opened 2015-11-07
Tweaked three command handlers

GrahamCampbell:


jbrooksuk:
I feel like the code which checks if we should notify should at least contain the condition now?

PR #1118, opened 2015-11-10
translate lang/zh-CN resources completely

phecho:
translate lang/zh-CN resources for dashboard and forms completely


jbrooksuk:
Thank you :beers:

PR #1123, opened 2015-11-10
added sub_title localization for SettingController,updated translatio…

phecho:
added sub_title localization for SettingController
fixed translations of zh-CN for human friendly


jbrooksuk:
Thanks buddy 🍻

PR #1128, opened 2015-11-11
Updated Chinese(zh-CN & zh-TW) forms translations

phecho:


phecho:
@GrahamCampbell 

OK，I will pay attention to it next time.

PR #1132, opened 2015-11-12
Incident API can now use incident templates with Twig templating

jbrooksuk:
Incident templates are way more useful now as they can be used when creating an incident via the API, with Twig templating!

The API is yet to be finalised. I like the `template` parameter, but not sure about `vars` yet. Thoughts?

If we decide to merge this, then we'll need to do the following:

- [ ] Documentation
- [x] Change the template dashboard page to show that the content can be Twig templates
- [x] Code Mirror should be used in the template dashboard pages to ease editing of templates
- [x] Edit Incident API should use the templating features too

Ping @cachethq/owners

### DIFF
--- a/app/Commands/Incident/ReportIncidentCommand.php
+++ b/app/Commands/Incident/ReportIncidentCommand.php
@@ -70,6 +70,20 @@ final class ReportIncidentCommand
     public $incident_date;
 
     /**
+     * A given incident template.
+     *
+     * @var string|null
+     */
+    public $template;
+
+    /**
+     * Variables for the incident template.
+     *
+     * @var string[]|null
+     */
+    public $template_vars;
+
+    /**
      * The validation rules.
      *
      * @var string[]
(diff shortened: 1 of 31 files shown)


jbrooksuk:
We'd also need to add an API for Incident Tmplates too.

jbrooksuk:
@cachethq/owners what do we think about the `vars` param?

joecohens:
Should we set the vars?

jbrooksuk:
Nope, it's all template driven. Variables can be anything. Maybe we could add default variables for things?

jbrooksuk:
Tests fail because PHP 7 is broken on Travis.

phecho:
Missing argument 9 for CachetHQ\Cachet\Commands\Incident\ReportIncidentCommand::__construct(), called in Cachet/app/Http/Controllers/Dashboard/IncidentController.php on line 121 and defined

PR #1138, opened 2015-11-16
Add tips in the situation of no items

phecho:
For the Index page of Metrics,Schedule and Incident Template.


GrahamCampbell:
:+1

PR #1139, opened 2015-11-16
Adjust the orders of php compiler for travis

phecho:
PHP 7 is up to twice as fast as PHP 5.x. 
So php unit test under PHP7 cost less time.We can get the response immediately if the travis unit test unpassed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,11 @@
 language: php
 
 php:
+  - 7.0
+  - hhvm
   - 5.5.9
   - 5.5
   - 5.6
-  - 7.0
-  - hhvm
 
 sudo: false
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
:-1: for this as is.

GrahamCampbell:
We should maybe just change the order to be:

* 7.0
* 5.6
* 5.5
* 5.5.9
* hhvm

jbrooksuk:
@phecho if you can change it to @GrahamCampbell's suggestion, then yeah!

phecho:
done.

GrahamCampbell:
Thanks.

GrahamCampbell:
Reverting this because it actually causes the total build time to increase.

PR #1147, opened 2015-11-18
Fixes #1145

### LINKED ISSUE #1145: Improve setup wizard setup2
If user set the timezone and locale in .env, setup wizard on step 2 will select them automatically.

phecho:


jbrooksuk:
@joecohens?

phecho:
There is no need to support i18n in setup wizard pages if we can't set the default locale in the .env file.

phecho:
Hi @joecohens ,
For one thing, I'm not setting the timezone in config/app.php, value of timezone is still UTC, only travis will be affected if I changed the default value of timezone in config/app.php.

For another thing, I just give users a choice to change the default value, Users can choose to use the default value.

joecohens:
@phecho I'll say this is ok only for the APP_LOCALE to default to a lang, still we can do the APP_TIMEZONE just without modifying config/app.php `'timezone' => env('APP_TIMEZONE', 'UTC'),` we have another value for this on config/cachet.php if you want to default the value since the beginning

PR #1152, opened 2015-11-19
Fixed app banner can not be removed in the theme setting page

phecho:


phecho:
I only added one line bellow in the template file

```html
<input type="hidden" name="remove_banner" value="0">
```

travis-ci unpassed, It's not because of the cause of my fix.

phecho:
@jbrooksuk 

Merge this pull-request please.

PR #1157, opened 2015-11-20
Give a right tip when a schedule be deleted successfully.

phecho:


joecohens:
Thanks @phecho :)

@cachethq/owners I guess we have to decide if we want success messages on all delete actions, since we only do it on ScheduleController and TeamController

phecho:
@joecohens :+1:

jbrooksuk:
We probably should have success messages where appropriate.

PR #1162, opened 2015-11-20
Drop the HQ on js

joecohens:
:stuck_out_tongue: 

@cachethq/owners


jbrooksuk:
"Drop the HQ. Just Cachet. It's cleaner."

PR #1166, opened 2015-11-21
Fixes #1147, also see #1145

phecho:
## Changes

* SetupController
```PHP
public function getIndex()
    {
        // If we've copied the .env.example file, then we should try and reset it.
        if (getenv('APP_KEY') === 'SomeRandomString') {
            $this->keyGenerate();
        }

        return View::make('setup')
            ->withPageTitle(trans('setup.setup'))
            ->withCacheDrivers($this->cacheDrivers)
            ->withAppTimezone(Config::get('cachet.timezone'))
            ->withAppLocale(getenv('APP_LOCALE'))
            ->withAppUrl(Request::root());
    }
```
* setup.blade.php
```HTML
<option value="{{ $timezone }}" @if(Input::old('settins.app_timezone', $app_timezone) == $timezone) selected @endif>
......
<option value="{{ $lang }}" @if(Input::old('settins.app_locale', $app_locale) == $lang) selected @endif>
                                    {{ $name }}

```

## Why
If we set the value of APP_LOCAL in .env,  change the default value of timezone in config/cachet.php,
setup wizard pages will automatically select the default value.

### DIFF
--- a/app/Http/Controllers/SetupController.php
+++ b/app/Http/Controllers/SetupController.php
@@ -65,6 +65,8 @@ class SetupController extends Controller
         return View::make('setup')
             ->withPageTitle(trans('setup.setup'))
             ->withCacheDrivers($this->cacheDrivers)
+            ->withAppTimezone(Config::get('cachet.timezone'))
+            ->withAppLocale(getenv('APP_LOCALE'))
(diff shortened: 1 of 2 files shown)


jbrooksuk:
I've since done this now :)

PR #1189, opened 2015-11-28
Update check system

jbrooksuk:


jbrooksuk:
Ping @cachethq/owners

joecohens:
Very nice James! :heart_eyes:

app/Http/Controllers/Dashboard/ApiController.php
```diff
@@ -89,4 +110,18 @@ public function getIncidentTemplate()
 
         throw new ModelNotFoundException("Incident template for $templateSlug could not be found.");
     }
+
+    /**
+     * Checks if Cachet is up to date.
+     *
+     * @return \Illuminate\Http\JsonResponse
+     */
+    public function checkVersion()
```
joecohens: We can use method injection here `checkVersion(Release $release)` instead :)

jbrooksuk:
Thanks Joe!

app/Http/Controllers/Dashboard/ApiController.php
jbrooksuk: I'll update to use this.

app/GitHub/Release.php
```diff
@@ -0,0 +1,70 @@
...
+use GuzzleHttp\Client;
+use Illuminate\Cache\CacheManager;
+use Illuminate\Config\Repository;
+
+class Release
+{
+    /**
+     * Cache instance.
+     *
+     * @var \Illuminate\Cache\CacheManager
```
GrahamCampbell: Don't inject a cache manager. Inject the cache repository contract.

app/Http/Controllers/Dashboard/ApiController.php
GrahamCampbell: Or, just `app(Release::class)->...`. That would be just fine too. I think we do that in StyleCI now eveywhere rather than method injection.

joecohens: Why we where not injecting? I forgot lol

jbrooksuk: Overhead I presume?

app/GitHub/Release.php
jbrooksuk: Does that mean we should also be using the config repository contract?

jbrooksuk: Forget the above :)

app/Http/Controllers/Dashboard/ApiController.php
GrahamCampbell: Because there's less code to write by not being parsed to inject. ;P

GrahamCampbell: Adding phpdoc is a ball ache, lol.

GrahamCampbell: Also, for performance reasons, we can resolve as late as possible, so we might bail out before we need to resolve an object.

jbrooksuk: I was only complaining about phpdoc like 2 days a go ;)

jbrooksuk:
We good to merge this then? :)

PR #1191, opened 2015-11-30
Fixed wrong var

phecho:
In the file of CachetHQ\Cachet\Commands\User\InviteTeamMemberCommand.php, line 21
from
```php
public $email;
```

to
```php
public $emails;
php

### DIFF
--- a/app/Commands/User/InviteTeamMemberCommand.php
+++ b/app/Commands/User/InviteTeamMemberCommand.php
@@ -18,7 +18,7 @@ final class InviteTeamMemberCommand
      *
      * @var string
      */
-    public $email;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by GrahamCampbell** [comment]
Thanks. I'll merge this into 2.0.

File: app/Commands/User/InviteTeamMemberCommand.php [hunk 1]
```diff
@@ -18,7 +18,7 @@
      *
      * @var string
```
GrahamCampbell: this phpdoc is also wrong

GrahamCampbell:
Merged.

PR #1193, opened 2015-12-01
Merge Atom & Rss Controllers into Feed Controller

phecho:
RssController and AtomController have a lot of same processing logic. 
They should be merged into one controller named FeedController.


app/Http/Controllers/FeedController.php
```diff
@@ -0,0 +1,120 @@
...
+use Roumen\Feed\Facades\Feed;
+
+class FeedController extends Controller
+{
+    /**
+     * Instance of Feed.
+     *
+     * @var Roumen\Feed\Facades\Feed
+     */
+    private $feed;
```
jbrooksuk: We usually use `protected` variables here.

phecho: OK, I will fix it immediately.

jbrooksuk: Also the phpdoc namespace needs to start with `\`.

phecho: My bad. I have fixed them.

phecho: @jbrooksuk 

Only the php5.5 test is not passed, not because of the change.

app/Http/Controllers/FeedController.php
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * Generates a Rss feed of all incidents.
+     *
+     * @param \CachetHQ\Cachet\Models\ComponentGroup|null $group
+     *
+     * @return \Illuminate\Http\Response
+     */
+    public function rssAction(ComponentGroup $group = null)
+    {
+        $this->isRss = true;
```
GrahamCampbell: don't set this property please

GrahamCampbell: pass it through to the feedAction method

app/Http/Controllers/FeedController.php
```diff
@@ -0,0 +1,120 @@
...
+        $this->feed->lang = Setting::get('app_locale');
```
GrahamCampbell: same here

app/Http/Controllers/FeedController.php
```diff
@@ -0,0 +1,120 @@
...
+use CachetHQ\Cachet\Models\Incident;
+use GrahamCampbell\Markdown\Facades\Markdown;
+use Illuminate\Routing\Controller;
+use Illuminate\Support\Str;
+use Roumen\Feed\Facades\Feed;
+
+class FeedController extends Controller
+{
+    /**
+     * Instance of Feed.
```
GrahamCampbell: this is't

GrahamCampbell: instance of feed, but it's actually a facade...

phecho: Feed::make() return value is actually a instance.
```php
 /**
     * Returns new instance of Feed class
     *
     * @return Feed
     */
    public function make()
    {
        return new Feed();
    }
```

phecho:
Hi @GrahamCampbell 
I have fixed them you mentioned above.

GrahamCampbell:
Thanks. :)

PR #1211, opened 2015-12-07
Moved stuff to foundation

GrahamCampbell:
Now that the testbench package is common to cachet, it has to mimic changes to StyleCI.


jbrooksuk:
:+1:

Issue #1192, opened 2015-11-30
Improved RSS and Atom indent


dereckson:
Currently, RSS and Atom indentation is rather incoherent, as shown at https://demo.cachethq.io/rss.

It would be tidier if some love is given to the streams output code.

jbrooksuk:
We're using https://github.com/RoumenDamianoff/laravel-feed to generate the feed. Can you open an issue there please.

phecho:
modify vendor/roumen/feed/src/views/rss.blade.php link the code below if you care the indentation.
@dereckson 
```xml
{!! '<'.'?'.'xml version="1.0" encoding="UTF-8" ?>' !!}
<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom" xmlns:content="http://purl.org/rss/1.0/modules/content/" xmlns:media="http://search.yahoo.com/mrss/"<?php foreach($namespaces as $n) echo " ".$n; ?>>
    <channel>
        <title><![CDATA[{{ $channel['title'] }}]]></title>
        <link>{{ Request::url() }}</link>
        <description>{{ $channel['description'] }}</description>
        <atom:link href="{{ $channel['link'] }}" rel="self"></atom:link>
@if (!empty($channel['logo']))
        <image>
            <url>{{ $channel['logo'] }}</url>
            <title>{{ $channel['title'] }}</title>
            <link>{{ Request::url() }}</link>
        </image>
@endif
        <language>{{ $channel['lang'] }}</language>
        <lastBuildDate>{{ $channel['pubdate'] }}</lastBuildDate>
@foreach($items as $item)
        <item>
            <title>{!! $item['title'] !!}</title>
            <link>{{ $item['link'] }}</link>
            <guid isPermaLink="true">{{ $item['link'] }}</guid>
            <description><![CDATA[{!! $item['description'] !!}]]></description>
@if (!empty($item['content']))
            <content:encoded><![CDATA[{!! $item['content'] !!}]]></content:encoded>
@endif
            <dc:creator xmlns:dc="http://purl.org/dc/elements/1.1/">{{ $item['author'] }}</dc:creator>
            <pubDate>{{ $item['pubdate'] }}</pubDate>
@if (!empty($item['enclosure']))
            <enclosure
@foreach ($item['enclosure'] as $k => $v)
            {!! $k.'="'.$v.'" ' !!}
@endforeach
            />
@endif
@if (!empty($item['media:content']))
            <media:content
[...]

PR #1223, opened 2015-12-15
Upgrade roumen/feed to 2.9.8 for #1192

phecho:


GrahamCampbell:
I'll do this on the 2.0 branch. Thanks though. :)

composer.lock
```diff
@@ -538,7 +538,7 @@
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/doctrine/cache/zipball/47c7128262da274f590ae6f86eb137a7a64e82af",
+                "url": "http://packagist.phpcomposer.com/files/doctrine/cache/47c7128262da274f590ae6f86eb137a7a64e82af.zip",
```
GrahamCampbell: all these changes here weren't want we wanted too

GrahamCampbell: also, i'd certainly recommend against this on security grounds anyway - you don't want your dependences intercepted on download and replaced with injected bad code

phecho: @GrahamCampbell 
OK, Thank you.

PR #1230, opened 2015-12-16
User can login with username or email

phecho:


resources/lang/en/forms.php
```diff
@@ -27,6 +27,7 @@
 
     // Login form fields
     'login' => [
+        'login'         => 'Username or Email',
```
GrahamCampbell: missing entry from all language files

phecho: @GrahamCampbell 
Thanks.
I have Added them.

PR #1231, opened 2015-12-16
User can login with username or email

phecho:


GrahamCampbell:
If we do want to allow this, then this looks ok to me. :+1:

phecho:
@GrahamCampbell 
I suggest that this feature is enabled for two reasons:
1,  Both of username and email are unique index.
2,  If username can't be logged in, it is easy to forget, there is no need to exist.

GrahamCampbell:
Well if @jbrooksuk and @joecohens give it a :+1:, then it'll get merged in. ;)

jbrooksuk:
Works! :)

joecohens:
:+1:

joecohens:
lol

PR #1241, opened 2015-12-19
Added missing translation

phecho:


GrahamCampbell:
We NEVER translate exception messages, ever.

GrahamCampbell:
If you want to catch the AlreadySubscribedException and translate at that point in the controller, we'd accept a PR for that.

PR #1248, opened 2015-12-21
Fix up php docs

thehereward:


joecohens:
Thanks! Can you revert the return void deletion on all of them please?

GrahamCampbell:
Please only send the param change. The other changes were wrong.

GrahamCampbell:
Also, please don't include my commits in your PRs.

thehereward:
Sorry for including your commit - honest mistake.

I have reverted the @return changes in #1250. I didn't realise using '@return void' on constructors was particularly common practice - we've always omitted the @return tag on our constructors and it does not appear to be suggested in the docs. Is there a particular reason for using them or is it just a personal preference?

GrahamCampbell:
We always included return tags so we can distinguish between methods that return nothing/garbage, and methods for which we've forgotten to include docs on. ;)

thehereward:
Okay, that makes some sense. :-)

What do you think about replacing `@return void` with `@return self` for constructors?

This would keep the distinction between documented and undocumented methods and be more descriptive of what the constructors are actually doing. `@return self` appears to be the default behaviour according to the [documentation](http://phpdoc.org/docs/latest/references/phpdoc/tags/return.html) anyway and it would stop someone else assuming it was incorrect and going around 'correcting' it (again)!

GrahamCampbell:
> What do you think about replacing @return void with @return self for constructors?

Return self is totally incorrect though. That documentation link is not official, and it totally incorrect.

As you can see, the fact that a constructor returns self is just bull shit: https://3v4l.org/IGGs3.

PR #1256, opened 2015-12-23
Added previous helper

phecho:


phecho:
User may enter the edit page for incident from statuspage

image

The cancel button shoud be back to where it come from.

GrahamCampbell:
Why are all those parameters needed? We never intend to actually use it like that.

GrahamCampbell:
Surely this would do:

```php
    /**
     * Create a new back url.
     *
     * @param string $route
     *
     * @return string
     */
    function previous($route)
    {
        $url = app('url');

        $previous = $url->previous()

        if ($url->full() === $previous) {
            return $url->route($route);
        }

        return $previous;
    }
```

phecho:
@GrahamCampbell 
:+1:

GrahamCampbell:
No. What I wrote was fine. Don't write code that we don't use.

phecho:
@GrahamCampbell 

I think ** $route !== null ** is needed.

If use call previous(), obviously the route is null. then the statement of  ** $previous === $url->full() ** will not be executed, and the this statement takes more time.

Right?

GrahamCampbell:
> I think $route !== null is needed.

No, it's not. Look at my code. I said that we don't allow passing `null`.

phecho:
OK, thank you. I will change them.

GrahamCampbell:
> The cancel button shoud be back to where it come from.

Honestly, I'm not convinced it actually should.

GrahamCampbell:
I think I'm actually :-1: here.

phecho:
@jbrooksuk 
What about you opnion?
Is it needed? if you give a :-1:  too. I will close this PR.

jbrooksuk:
I'd much rather pressing back goes back to the dashboard.

PR #1260, opened 2015-12-23
Pull in changes from Laravel

GrahamCampbell:
Some of the language changes are a bit dodgy.


GrahamCampbell:
Can we actually remove the password stuff that laravel ships with? Do we actually use it?

jbrooksuk:
We don't use it :)

PR #1264, opened 2015-12-23
Only generate code coverage on php 5.6

GrahamCampbell:
An alternative attempt to speed up builds. I've already updated our scrutinizer config to expect 1 build rather than 3 now.


GrahamCampbell:
Oh, oh, oh! And now that we have no nested logic, we can disable coverage on PRs. Yes!

jbrooksuk:
:D

joecohens:
:clap:

PR #1268, opened 2015-12-23
Fixes #1267

### LINKED ISSUE #1267: Clean up TimestampsTrait
1.  scheduled_at is reduplicated. It is defined in both of TimestampsTrait and IncidentPresenter. Remove it from TimestampsTrait.
2.  move verified_at from  TimestampsTrait to it's own Presenter (SubscriberPresenter)

phecho:


phecho:
@GrahamCampbell @jbrooksuk 
How about these changes?
BTW, Travis is so fast now.

app/Presenters/SubscriberPresenter.php
```diff
@@ -12,12 +12,24 @@
...
 
     /**
+     * Present formatted date time.
+     *
+     * @return string
+     */
+    public function verified_at()
+    {
+        return (new Date($this->wrappedObject->verified_at))
+            ->setTimezone($this->setting->get('app_timezone'))->toDateTimeString();
```
GrahamCampbell: Please don't do this manual code.

GrahamCampbell: Please use https://github.com/cachethq/Cachet/blob/master/app/Dates/DateFactory.php.

GrahamCampbell: Oh, I see it was already like that. Hmm. I'll change this in another PR then.

PR #1271, opened 2015-12-23
Minor controller cleanup

GrahamCampbell:


jbrooksuk:
What's the reason for using `app` over injection?

GrahamCampbell:
Just faster to write. We don't have to write any phpdoc. ;)

jbrooksuk:
> We don't have to write any phpdoc. ;)

Sold.

PR #1273, opened 2015-12-23
Don't use filters, use middleware. Closes #1269

### LINKED ISSUE #1269: Stop using filters
Just noticed the setup controller still uses a filter. We need to switch totally over to middleware.

jbrooksuk:


GrahamCampbell:
Are there any other places where we're using filters, or is that it. That's the only one I noticed, but I wasn't looking hard.

jbrooksuk:
That's all I've fixed, but I'm pretty sure that'd be all.

GrahamCampbell:
:+1:

PR #1276, opened 2015-12-23
Improved validation of component_id field

jbrooksuk:


app/Commands/Incident/ReportIncidentCommand.php
```diff
@@ -93,7 +93,7 @@
         'status'           => 'required|int|min:0|max:4',
         'message'          => 'string',
         'visible'          => 'bool',
-        'component_id'     => 'int',
+        'component_id'     => 'int|exists:components|required_with:component_status',
```
GrahamCampbell: > exists:components

What does that do?

jbrooksuk: Checks that the component actually exists.

GrahamCampbell:
Please rebase this. ;P

jbrooksuk:
Rebased.

app/Commands/Incident/ReportIncidentCommand.php
GrahamCampbell: I thought we agreed not to make database queries in the validator?

jbrooksuk: Good point, we did.

GrahamCampbell: Just done a quick search and `exists:` doesn't match anything in our codebase.

GrahamCampbell:
I've added issue closing data to the description. Is that correct?

GrahamCampbell:
Rather than passing through the ID, shouldn't we pass through an actual instance of the model?

GrahamCampbell:
That way we already know it exists, right?

jbrooksuk:
> I've added issue closing data to the description. Is that correct?

No, I don't know WTF is going on in that issue. This is just extra validation.

> Rather than passing through the ID, shouldn't we pass through an actual instance of the model?

Yeah, that's the better way of doing it.

GrahamCampbell:
Ok, I've removed the closes docs.

PR #1280, opened 2015-12-24
Make MetricPoint events consistent.

thehereward:
I noticed the properties were inconsistent during the documentation changes.
I've kept the original properties as well, for backwards compatibility although have marked these as deprecated.

### DIFF
--- a/app/Events/Metric/MetricPointWasAddedEvent.php
+++ b/app/Events/Metric/MetricPointWasAddedEvent.php
@@ -20,17 +20,17 @@ class MetricPointWasAddedEvent implements MetricEventInterface
      *
      * @var \CachetHQ\Cachet\Models\MetricPoint
      */
-    public $metric;
+    public $metricPoint;
 
     /**
      * Create a new metric point was added event instance.
      *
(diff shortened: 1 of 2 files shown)


app/Events/Metric/MetricPointWasAddedEvent.php
```diff
@@ -20,17 +20,27 @@ class MetricPointWasAddedEvent implements MetricEventInterface
...
      * @var \CachetHQ\Cachet\Models\MetricPoint
      */
+    public $metricPoint;
+
+    /**
+     * The metric point that was added.
+     *
+     * @var \CachetHQ\Cachet\Models\MetricPoint
+     *
+     * @deprecated Use $metricPoint instead.
```
GrahamCampbell: don't deprecate, just delete

GrahamCampbell:
Please rebase.

GrahamCampbell:
> for backwards compatibility

We have no need to care about that since we're an application.

GrahamCampbell:
Please correct the commit message.

GrahamCampbell:
Merged thanks.

PR #1294, opened 2015-12-24
Config Improvements

### LINKED ISSUE #1170: Cleanup use of "Setting"

### LINKED ISSUE #1300: writeEnv function broken
Laravel 5.2 doesn't load the env anymore when config is cached.

GrahamCampbell:
:christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree:
### Native configuration integration
:christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree: :christmas_tree:

# IT IS HERE

The new rules are that we NEVER interact with the setting repository in our code. That's there so it we can load it into the native configuration system once. I've also setup automatic config caching. This means we can load it, cache the result, then never hit the db or mess around with config at runtime on future executions.

---

Closes #1170. Closes #1300.


jbrooksuk:
YEAH!

GrahamCampbell:
It seem I did mess this up a bit. I can see why the first test is erroring, but why the second is failing, I'm not sure.

GrahamCampbell:
Hmm, I'm not convived that test was ever correct actually. Looks to like like our usage of the different timezone settings was incorrect. We were inconsistant about reading Setting app_timezone, or Config cachet.timezone. I cleaned this up so that we only read cachet.timezone, and now eveything breaks.

What is the actual wanted behaviour here. Ping @joecohens. Have I messed this up?

joecohens:
Actually the tests wasn't ok before. If we change the Timezone via the API Header the response should be also on that timezone. 

So what we need to actually test it's that even if you are creating a metric point on a different timezone, we are always converting it back to UTC to store it and when we get it back it's has to be on that Header timezone. Does it make sense?

This is fixed on the last commit.

/cc @CachetHQ/owners

jbrooksuk:
Makes sense.

jbrooksuk:
Rebased!

GrahamCampbell:
Not quite ready for merging though. Still has a few issues.

GrahamCampbell:
I'll sort this later today.

jbrooksuk:
:+1:

GrahamCampbell:
Ping @CachetHQ/owners.

jbrooksuk:
:boom:

GrahamCampbell:
Did you check it all worked?

GrahamCampbell:
I don't have a local install, so I no idea. :D

jbrooksuk:
Tests continued to pass and everything looked fine :)

PR #1297, opened 2015-12-25
Added const for user level

phecho:


app/Models/User.php
```diff
@@ -24,6 +24,8 @@ class User extends Model implements AuthenticatableContract, CanResetPasswordCon
 {
     use Authenticatable, CanResetPassword, ValidatingTrait;
 
+    const LEVEL_ADMIN = 1;
+    const LEVEL_USER = 2;
```
GrahamCampbell: Please add phpdoc, and a newline after this, and it looks good.

GrahamCampbell:
A bit like in here:

image

jbrooksuk:
:gift:

GrahamCampbell:
Thank you very much. ;)

PR #1298, opened 2015-12-25
Keep DB_CONNECTION consistent with Laravel 5

phecho:


GrahamCampbell:
Thanks, but I'm actually :-1: here. This is a large breaking change, and I don't think laravel named it right compared to the other driver env variables.

phecho:
@GrahamCampbell 
I also think that the naming of laravel is not very appropriate. So I'm saying that we need to keep it consistent with Laravel 5. To do this to help users to eliminate ambiguity.

jbrooksuk:
:-1: for me. This is a change I'd like to avoid forever, unless we had some migration stuff going on for upgrades.

PR #1301, opened 2015-12-25
Improved human_status of ComponentPresenter

phecho:
Why are there two different implementations of the same functionality?

compare human_status in IncidentPresenter and ComponentPresenter.


jbrooksuk:
They do different things.

phecho:
@jbrooksuk 

I didn't see any other differences except for the parameters (cachet.incidents.status and cachet.component.status).

PR #1303, opened 2015-12-25
Test that the cachet:* commands work

jbrooksuk:
This tests that `cachet:seed` and `cachet:metrics` work.


GrahamCampbell:
Maybe make separate test methods for them?

jbrooksuk:
I thought that, but figured they're overkill? I can do if you think it's a good idea though?

GrahamCampbell:
I think it would be useful to keep that atomic, so we can see exactly what's broken if we mess up.

jbrooksuk:
Sure, I'll do that.

jbrooksuk:
Done!

jbrooksuk:
Closing this, I'll just do it again.

PR #1312, opened 2015-12-27
Fixed typo

phecho:


jbrooksuk:
Probably should be "Invalid email/username or password combination"

phecho:
I don't think it is necessary. User should know what the username stands for.
just like github:
image

what about  "Invalid login or password combination" ?

resources/views/auth/login.blade.php
```diff
@@ -31,7 +31,7 @@
 
                     <div class="form-group">
                         <label class="sr-only">{{ trans('forms.login.login') }}</label>
-                        <input autocomplete="off" class="form-control login-input" placeholder="{{ trans('forms.login.login') }}" required="required" name="login" type="text" autofocus>
+                        <input autocomplete="off" class="form-control login-input" placeholder="{{ trans('forms.login.login') }}" required="required" name="login" type="text" value="{{Binput::old('login')}}" autofocus>
```
GrahamCampbell: cs

jbrooksuk:
Ok, I can go with that :)

resources/views/auth/login.blade.php
jbrooksuk: Fix and we can merge :)

jbrooksuk:
Cheers!

GrahamCampbell:
@jbrooksuk The language files are now out of sync.

jbrooksuk:
I need to upload them to CrowdIn.

PR #1322, opened 2016-01-03
Fixes missing titles

jbrooksuk:


app/Presenters/ComponentGroupPresenter.php
```diff
@@ -36,8 +36,8 @@ public function lowest_status()
      */
     public function lowest_human_status()
     {
-        if ($enabledComponents = $this->wrappedObject->enabled_components()->orderBy('status', 'desc')->first()) {
-            return $enabledComponents->human_status;
+        if ($component = $this->wrappedObject->enabled_components_lowest->first()) {
```
GrahamCampbell: no we not need a function call here rather than property access?

GrahamCampbell: otherwise you're falling `first` on a collection, right?

jbrooksuk: Does the same thing I guess, it's just waiting for the results?

GrahamCampbell: Well, first on the query is much faster because of the `limit 1`

jbrooksuk: Ok :)

PR #1327, opened 2016-01-04
We need to use the raw value of app_about

### LINKED ISSUE #1326: About this page - Save issue
When I put something in "About this page" like: 

    Test

It's converted to

    <p>Test</p>

That is okay. But if I add something like:

    <p>Test</p>
     a

Everything will be deleted if I save the changes

Maybe related with #1325

jbrooksuk:
Fixes #1326


GrahamCampbell:
We need to avoid using Setting:: in views.

jbrooksuk:
> We need to avoid using Setting:: in views.

I was being lazy, but then I've just realised I don't need to be since the settings pages were tidied up. Hang on.

jbrooksuk:
Fixed up.

PR #1329, opened 2016-01-04
Wrong array key in MetrixController

ZengineChris:
Fixed a wrong key identifier in createMetricAction method.


jbrooksuk:
Thank you :+1:

PR #1340, opened 2016-01-05
Ensure incident field names are the same everywhere

### LINKED ISSUE #1339: Incident templates not imported
On the "Add an Incident" form selecting an Incident Template from the dropdown doesn't update the Name and Message fields.

This was previously fixed in #1064 but reverted to fix #1102

agilmore:
Closes #1339.


jbrooksuk:
I'd rather that we didn't use the array index for `incident[]`, instead just name the fields by themselves.

PR #1343, opened 2016-01-06
Ensure incident field names are the same everywhere

agilmore:
Changes the JS to not use incident[]


GrahamCampbell:
Please send again but without compiling the js. You appear to have compiled it slightly differently to us. The diff shows 2 extra lines in the compiled file.

PR #1360, opened 2016-01-10
Per-component subscriptions

### LINKED ISSUE #734: Subscribe to specific components
I have some cool ideas for V2 :

- Might be cool to subscribe to specific components only for some users
- Multiple client dashboards (separated)?

If this is not the place for ideas, please let me know haha!

jbrooksuk:
Aside from tests, I think the core system works for this, just need a UI now to manage preferences.

Closes #734

### DIFF
--- a/app/Bus/Commands/Subscriber/SubscribeSubscriberCommand.php
+++ b/app/Bus/Commands/Subscriber/SubscribeSubscriberCommand.php
@@ -11,6 +11,11 @@
 
 namespace CachetHQ\Cachet\Bus\Commands\Subscriber;
 
+/**
+ * This is the subscribe subscriber command.
+ *
+ * @author James Brooks <james@alt-three.com>
+ */
 final class SubscribeSubscriberCommand
 {
     /**
@@ -28,6 +33,13 @@ final class SubscribeSubscriberCommand
     public $verified;
 
     /**
+     * The subscriptions that we want to add.
+     *
+     * @var array|null
+     */
+    public $subscriptions;
+
+    /**
      * The validation rules.
      *
      * @var array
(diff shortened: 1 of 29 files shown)


jbrooksuk:
For now, rather than a UI for managing preferences, how about an unsubscribe from component link in the email itself? Ping @joecohens

joecohens:
Yeah we need a way to unsubscribe from a single component as well

jbrooksuk:
@joecohens do you believe this to be ready and finished?

joecohens:
I think we are just missing the link to unsubscribe to a specific component. We could attach it to the current one.

jbrooksuk:
Are you ok adding this in?

joecohens:
Adding it now :)

jbrooksuk:
Seems to work fine Joe! Squash and merge?

jbrooksuk:
Squashed. Waiting on tests then we can merge!

joecohens:
:raised_hands:

txsteven:
Currently on 2.1.0 so forgive me if this is addressed in 2.1.1 But If I subscribe to only component A. And I create an incident on Component B. The user who has only subscribed to A still gets the notification. Is this working as designed or a bug fixed in 2.1.1?  

Respectfully,

jbrooksuk:
Ping @joecohens, this doesn't sound right?

joecohens:
Mmm I think we didn't filter subscriptions on the incidents, we only did when you update the a component. But I'm not sure if we should, because they are different subscriptions, one is site wide (all incidents) and the other one it's just for component updates not it's incidents. @jbrooksuk we need to review the functionality.

https://github.com/CachetHQ/Cachet/blob/master/app/Bus/Handlers/Events/Component/SendComponentUpdateEmailNotificationHandler.php#L59

https://github.com/CachetHQ/Cachet/blob/master/app/Bus/Handlers/Events/Incident/SendIncidentEmailNotificationHandler.php#L68

PR #1396, opened 2016-01-25
Site title shows Cachet is we're showing support. Closes #1394

### LINKED ISSUE #1394: Remove the "| Cachet" when show support is off
Minor annoyance to some people who don't want to show that they're using Cachet. Understandable.

jbrooksuk:


joecohens:
:+1:

PR #1402, opened 2016-01-29
Allow component groups to be collapsed by default

### LINKED ISSUE #1398: Add option for component group to by minimized by default
At the moment, component groups have the option to be minimised but by default you see all the components.

I have a lot of components in a single group, and I'd like to be able to make that group minimised by default, so a user has to open it to see the status of their component. A tickbox or something on the component groups page would be perfect.

jbrooksuk:
Closes #1398


joecohens:
:+1:

PR #1420, opened 2016-01-30
Disable xdebug when not running as PHP 5.6

jbrooksuk:
Although @GrahamCampbell hasn't had much luck with this, we've at least tried to make Travis more efficient.


GrahamCampbell:
This makes essentially zero difference to the performance. Most of the time is waiting on http requests and git clones.

PR #1430, opened 2016-02-01
Added Component badges/shields

### LINKED ISSUE #1313: Display a Badge for each Component
Support an endpoint for each component such that you can do things like:

 - Operational
 - Performance Issues
 - Partial Outage
 - Major Outage

I'm using http://shields.io/ as the example (which we could use as the implementation). 

I expect we could do something like:
 1. Go to https://demo.cachethq.io/api/components/1/badge
 2. Redirects with 302 Found (or 307 Temporary Redirect, your choice) to shields.io

We could also replace the text on the UI with these badges:
screen shot 2015-12-28 at 9 44 48 am

__Note:__ I would love to contribute this change.  Do you have recommendations for where this should go?  What should the messages be (status/name/etc.)?

jbrooksuk:
image

---

Closes #1313.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,9 +4,70 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "ba38e29b36a482c9350a62373ef12a5b",
-    "content-hash": "a28b0a82f75e6f39655f7bb1def163e4",
+    "hash": "9e2445f92e1fd3c9598bb0001ea765f7",
+    "content-hash": "afe2e5eb227fb6b808aadafb3b51bd3a",
     "packages": [
+        {
+            "name": "alt-three/badger",
+            "version": "v1.0.7",
+            "source": {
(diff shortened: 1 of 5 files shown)


jbrooksuk:
CC @CachetHQ/owners can we get this in?

jbrooksuk:
We're serving with the right content-type now.

joecohens:
:+1:

composer.lock
```diff
@@ -2021,16 +2082,16 @@
         },
         {
             "name": "mccool/laravel-auto-presenter",
-            "version": "4.2.1",
+            "version": "v4.2.1",
```
GrahamCampbell: @jbrooksuk Please clear your composer cache. Composer has f***ed up here again.

joecohens: This is odd

GrahamCampbell: Not really. It was my fault. I messed the name and retagged, but composer doesn't bother with that and uses it's stale cache.

jbrooksuk: Ah ok :)

PR #1446, opened 2016-02-02
Twigbridge fixes

### LINKED ISSUE #1437: Twig doesn't work
i use twig in my incident templates but it doesn't work. It output only the twig code
e.g.`{% if {{ app.request.get('trackDomain') }} == 'test.webtrekk.com'|url_encode %}
 Das ist ein Test
{% endif %}`

I'm on HEAD e33f794. i tested it also with deee3b7. It doesn't work. Is there anything i have to do? Config? Shortcut before the twig code start?

### LINKED ISSUE #1445: Broken twig bridge config
image

That debug config clearly won't work. You can't read the config before it's been filled.

jbrooksuk:
Fixes #1437, fixes #1445

### DIFF
--- a/app/Bus/Handlers/Commands/Incident/ReportIncidentCommandHandler.php
+++ b/app/Bus/Handlers/Commands/Incident/ReportIncidentCommandHandler.php
@@ -50,16 +50,17 @@ class ReportIncidentCommandHandler
      */
     public function handle(ReportIncidentCommand $command)
     {
-        if ($command->template) {
-            $command->message = $this->parseIncidentTemplate($command->template, $command->template_vars);
-        }
-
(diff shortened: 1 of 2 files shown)


jbrooksuk:
Ballsed up the commit message, but whatever.

Citrullin:
I'm on ab2ab5d . And the twig template doesn't work. Our Story we want to build with CachetHQ: We have different tracking servers etc. for our customers. With this template i want to filter the status updates .

jbrooksuk:
If you're still using the Twig code from #1437 then it's invalid. I had to make some changes for it to work.

Citrullin:
Ahh, okay. :)

Issue #1363, opened 2016-01-12
Broken metric y-axis scaling


GrahamCampbell:
The y-axis should start at 0, not the min value.

![image](https://cloud.githubusercontent.com/assets/2829600/12273218/8be5131a-b95c-11e5-8b07-5eb56b9d8dd2.png)

jbrooksuk:
I'm unsure what's up here @GrahamCampbell. Is it because the first point in the second chart is not at 0?

jbrooksuk:
Oddly it looks ok here:

![image](https://cloud.githubusercontent.com/assets/246103/12458686/f7847966-bfa2-11e5-8af2-bd841cc2a1ef.png)

GrahamCampbell:
That's because there are zero values there.

GrahamCampbell:
Add more points so everything is much greater than zero.

PR #1469, opened 2016-02-09
Make the Y axis start at zero

clbn:
Because tiny changes in, say, response time (210ms -> 219ms -> 208ms) on the non-zero Y scale (200ms to 220ms) might look like really big scary peaks or dips even if they are actually not.


GrahamCampbell:
Thanks @clbn. This also fixes https://github.com/CachetHQ/Cachet/issues/1363.

jbrooksuk:
Ah! Awesome :)

PR #1473, opened 2016-02-10
[WIP] Show the component status within a label

jbrooksuk:
Looks like this:

image

A custom stylesheet can be applied to remove the colours though.

### DIFF
--- a/app/Presenters/ComponentPresenter.php
+++ b/app/Presenters/ComponentPresenter.php
@@ -35,6 +35,21 @@ class ComponentPresenter extends BasePresenter implements Arrayable
     }
 
     /**
+     * Return the color for the status label.
+     *
+     * @return string
+     */
+    public function label_color()
+    {
+        switch ($this->wrappedObject->status) {
+            case 1: return 'success';
+            case 2: return 'info';
+            case 3: return 'warning';
+            case 4: return 'danger';
+        }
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Moved the milestone. I assume you clicked the wrong one by mistake?

jbrooksuk:
I did.

GrahamCampbell:
Let me know when you're ready to start merging "features", and I'll spin off a 2.1 branch first.

GrahamCampbell:
Would be useful to make this configurable?

TheoBearman:
I think so.

jbrooksuk:
Would it? You can just re-style them.

app/Presenters/ComponentPresenter.php
```diff
@@ -35,6 +35,21 @@ public function status_color()
...
 
     /**
+     * Return the color for the status label.
+     *
+     * @return string
+     */
+    public function label_color()
+    {
+        switch ($this->wrappedObject->status) {
+            case 1: return 'success';
```
GrahamCampbell: cs

GrahamCampbell: we don't yet have a fixer for this on styleci, but currently, this isn't PSR-2

PR #1477, opened 2016-02-10
Tweaked the subscriber model

GrahamCampbell:


app/Models/Subscriber.php
```diff
@@ -49,6 +49,13 @@ class Subscriber extends Model implements HasPresenter
     ];
 
     /**
+     * The relations to eager load on every query.
+     *
+     * @var string[]
+     */
+    protected $with = ['subscriptions'];
```
GrahamCampbell: This is unneeded though, for now, I think. Should I remove it and just backport the typo fix?

jbrooksuk: Sure.

Issue #1486, opened 2016-02-12
v2.1.0 installer says [object Object] just before complete setup once.


sawanoboly:
duplicate from https://github.com/CachetHQ/Cachet/issues/1324#issuecomment-183189686

![setup](https://cloud.githubusercontent.com/assets/272388/13000008/b424e766-d198-11e5-8484-866ff22276b2.jpg)

Reproduce.

- install v2.1.0 below steps
    - composer install --no-dev -o
    - php artisan key:generate
    - php artisan app:install
- I through the setup wizard.

When I click `complete setup` again after occur it. It completes without problem.

GrahamCampbell:
@joecohens This doesn't fix the frontend bug though?

joecohens:
We were getting the object object since the request didn't have any kind of response or a 500 I don't remember. We could code to catch if the error message thrown by the server it's not "readable" and show a generic error maybe?

GrahamCampbell:
Are you sure just didn't forget to update the setup page js for the new error format from the exception package?

sawanoboly:
I entered a weak password to see this message.

joecohens:
I'll double check :)

ihatehandles:
I can also confirm running into [object object] in more than one installation, but I failed to consistently reproduce the error message consistently when [I reported](https://github.com/CachetHQ/Cachet/issues/1460) some other setup troubles.

PR #1491, opened 2016-02-12
Fix path to write the env

joecohens:
PING @GrahamCampbell 

Closes https://github.com/CachetHQ/Cachet/issues/1486

### DIFF
--- a/app/Http/Controllers/SetupController.php
+++ b/app/Http/Controllers/SetupController.php
@@ -222,12 +222,13 @@ class SetupController extends Controller
     protected function writeEnv($key, $value)
     {
         $path = app()->environmentFile();
+        $basePath = base_path($path);
 
         try {
             (new Dotenv(app()->environmentPath(), $path))->load();
 
-            file_put_contents($path, str_replace(
-                env(strtoupper($key)), $value, file_get_contents($path)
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/SetupController.php
```diff
@@ -222,12 +222,13 @@ public function postStep3()
     protected function writeEnv($key, $value)
     {
         $path = app()->environmentFile();
+        $basePath = base_path($path);
 
         try {
             (new Dotenv(app()->environmentPath(), $path))->load();
```
GrahamCampbell: Why does the path work here but not else where?

joecohens: No

joecohens: Not really sure :( but we had an error on writing the file

GrahamCampbell: Oh, wait, i see. Don't call base_path. One sec...

GrahamCampbell:
Also, we can't assume `base_path` is where the `.env` file is located because laravel supports changing that location.

GrahamCampbell:
I've got a slightly different fix on the way. :)

PR #1512, opened 2016-02-17
Load metrics via AJAX

### LINKED ISSUE #819: Landing page times out due to graphs
There is a bug with the graphs. I've got a program logging uptime to cachet (every second). Cachet displays about 10 hours worth of data, so at any time it loads up to 36000 data points per graph from the database.

When I request the main page with two graphs, containing 72000 data points, the page loads for so long that nginx halts the requests rendering the 504 gateway timeout..

I wouldn't want to know what would happen if this were public-facing. Its obvious that the data aggregation shouldn't be happening when someone requests the page, that should already be done by a cronjob regularly..

I have absolutely no experience with laravel therefore call upon the maintainers to consider this issue. I am happy to help anyone get set up with the monitoring tools

jbrooksuk:
Fixes #819

---

Ping @CachetHQ/owners

### DIFF
--- a/app/Composers/MetricsComposer.php
+++ b/app/Composers/MetricsComposer.php
@@ -12,29 +12,11 @@
 namespace CachetHQ\Cachet\Composers;
 
 use CachetHQ\Cachet\Models\Metric;
-use CachetHQ\Cachet\Repositories\Metric\MetricRepository;
 use Illuminate\Contracts\View\View;
 use Illuminate\Support\Facades\Config;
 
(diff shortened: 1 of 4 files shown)


GrahamCampbell:
Is this good to go then?

jbrooksuk:
Everything seems to be working in testing.

There are some tweaks I want to the metric API, then we can share the public one.

joecohens:
:beers:

PR #1525, opened 2016-02-18
Change metric_points.value to DECIMAL(15,3)

### LINKED ISSUE #1508: Limitation in DB in metric_points
i am using Cachet for a status page showing some metrics. Yesterday i saw, that a normally growing metric is still standing at the same level (10.000.000 to be exact).
i am wondering why. First i search for the problem in my code. But then i found out, that the problem is your Database.

metric_points -> value is decimal(10,3) and so 9.999.999,99 is the highest value Cachet can handle. Can you please change this to a higher value (for example decimal(15,3) - or better, why not using float (double)?

jbrooksuk:
Closes #1508

---

Putting this into 2.2.0 as there is a migration.

### DIFF
--- a/database/migrations/2016_02_18_085210_AlterTableMetricPointsChangeValueColumn.php
+++ b/database/migrations/2016_02_18_085210_AlterTableMetricPointsChangeValueColumn.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of Cachet.
+ *
+ * (c) Alt Three Services Limited
+ *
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Are we all sure this will correctly migrate existing data?

jbrooksuk:
I could do with somebody checking on PgSQL, but it was fine for my demo data in MySQL.

jbrooksuk:
Needs more verification.

PR #1538, opened 2016-02-21
Remove JamesForks/badger dependency

### LINKED ISSUE #1537: [BLOCKER] Drop dependency on JamesForks/poser
We can't ship 2.2 while still needing that.

jbrooksuk:
Closes #1537

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "a469e0ab4ecee06801479b07e8604ec3",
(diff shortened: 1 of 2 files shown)


composer.lock
```diff
@@ -2082,16 +2082,16 @@
         },
         {
             "name": "mccool/laravel-auto-presenter",
-            "version": "4.2.1",
+            "version": "v4.2.1",
```
GrahamCampbell: @jbrooksuk Your composer cache or vendor folder still need deleting,.

jbrooksuk: I've done it numerous times.

jbrooksuk: Doing it again.

GrahamCampbell: DId you delete the lock file as well?

PR #1548, opened 2016-02-23
Updated alt-three/testbench

jbrooksuk:


composer.json
```diff
@@ -42,7 +42,7 @@
         "roumen/feed": "^2.10"
     },
     "require-dev": {
-        "alt-three/testbench": "^1.1",
+        "alt-three/testbench": "^1.3",
```
GrahamCampbell: No need to bump this. Cachet doesn't require 1.3, only 1.1 or greater.

composer.lock
```diff
@@ -2082,16 +2082,16 @@
         },
         {
             "name": "mccool/laravel-auto-presenter",
-            "version": "4.2.1",
+            "version": "v4.2.1",
```
GrahamCampbell: Your composer cache is still wrong.

GrahamCampbell:
Please both delete the vendor folder, the composer cache, and the lock file. You're missing one of those steps because composer is still resolving the wrong versions.

composer.lock
jbrooksuk: I'm telling you, I've removed `vendor`, `composer.lock`, ran `composer clear-cache`. Numerous times.

jbrooksuk:
I've done everything.

PR #1549, opened 2016-02-23
Make the model factories more dynamic

jbrooksuk:
The model factories have a bit more variety to them. Tests should be more reliable.

### DIFF
--- a/database/factories/ModelFactory.php
+++ b/database/factories/ModelFactory.php
@@ -25,7 +25,7 @@ $factory->define(Component::class, function ($faker) {
         'name'        => $faker->sentence(),
         'description' => $faker->paragraph(),
         'link'        => $faker->url(),
-        'status'      => 1,
+        'status'      => random_int(1, 4),
         'order'       => 0,
     ];
 });
@@ -34,7 +34,7 @@ $factory->define(ComponentGroup::class, function ($faker) {
     return [
         'name'      => $faker->words(2, true),
         'order'     => 0,
-        'collapsed' => false,
+        'collapsed' => $faker->boolean(),
     ];
 });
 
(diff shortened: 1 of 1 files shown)


database/factories/ModelFactory.php
```diff
@@ -34,7 +34,7 @@
     return [
         'name'      => $faker->words(2, true),
         'order'     => 0,
-        'collapsed' => false,
+        'collapsed' => (bool) rand(0, 1),
```
joecohens: I think there is a $faker->boolean() do we prefer it here?

jbrooksuk: Ah, is there? We'd be better using that.

jbrooksuk:
@joecohens fixed :)

database/factories/ModelFactory.php
```diff
@@ -25,7 +25,7 @@
         'name'        => $faker->sentence(),
         'description' => $faker->paragraph(),
         'link'        => $faker->url(),
-        'status'      => 1,
+        'status'      => rand(1, 4),
```
GrahamCampbell: better to use random_int

GrahamCampbell: that should be used always instead of rand or mt_rand

GrahamCampbell: it also actually has better performance than mt_rand

database/factories/ModelFactory.php
```diff
@@ -61,14 +61,14 @@
...
-        'display_chart' => 1,
+        'calc_type'     => $faker->boolean(),
+        'display_chart' => $faker->boolean(),
     ];
 });
 
 $factory->define(MetricPoint::class, function ($faker) {
     return [
-        'metric_id' => 1,
+        'metric_id' => factory('CachetHQ\\Cachet\\Models\\Metric')->create()->id,
```
GrahamCampbell: probably best to use ::class notation

database/factories/ModelFactory.php
jbrooksuk: :+1:

PR #1554, opened 2016-02-24
Tweaks

jbrooksuk:
image


GrahamCampbell:
Feels like there's not enough padding before the heading.

jbrooksuk:
I'm going to keep messing around with this anyway and I know @joecohens has some things he wants to do.

PR #1559, opened 2016-02-25
[WIP] Core event install listener

jbrooksuk:
`app:install` still won't fire an event (just seen Graham closed the PR I made). What's another way of doing this? Hooking into `extrastuff`?


jbrooksuk:
We won't need this. It'll be part of the Core package.

PR #1565, opened 2016-02-26
Move the CONTRIBUTING.md file into .github

jbrooksuk:


GrahamCampbell:
Why would we want to do this?

jbrooksuk:
Because that's where it belongs now.

GrahamCampbell:
It doesn't have to go there. Why would we want to make it hard to find? We wouldn;'t move the readme, so why move that file?

jbrooksuk:
Revert then :)

PR #1570, opened 2016-02-26
Completed Hungarian translation

tomcsanyid:


jbrooksuk:
Unfortunately we can't accept this translation here. You need to submit it to [CrowdIn](https://translate.cachethq.io) please.

PR #1579, opened 2016-03-01
Implement very basic API searching

### LINKED ISSUE #1348: Add some search parameters when performing a GET on /incidents or /components
It would be great to be able to get incidents or components using for example, name or url parameters.
Currently, I need to parse all result (i.e. process all paginated results) and then search on it. 
What a waste of CPU ;)

jbrooksuk:
Closes #1348

---

@joecohens maybe something to re-visit so we can add things like between dates etc.

### DIFF
--- a/app/Http/Controllers/Api/ComponentController.php
+++ b/app/Http/Controllers/Api/ComponentController.php
@@ -37,6 +37,8 @@ class ComponentController extends AbstractApiController
             $components = Component::enabled();
         }
 
+        $components->search(Binput::except(['sort', 'order', 'per_page']));
+
         if ($sortBy = Binput::get('sort')) {
             $direction = Binput::has('order') && Binput::get('order') == 'desc';
 
--- a/app/Models/Traits/SearchableTrait.php
+++ b/app/Models/Traits/SearchableTrait.php
@@ -1,0 +1,43 @@
+<?php
+
+/*
+ * This file is part of Cachet.
+ *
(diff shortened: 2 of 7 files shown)


app/Http/Controllers/Api/ComponentController.php
```diff
@@ -24,6 +24,8 @@
 
 class ComponentController extends AbstractApiController
 {
+    use SearchableTrait;
```
joecohens: Maybe we could do it on the model as we did with the sorting?

jbrooksuk: Yeah, okay :)

jbrooksuk:
@joecohens implemented on the model. Will merge once complete.

joecohens:
AWESOME!!! :beers:

app/Models/Traits/SearchableTrait.php
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @author James Brooks <james@alt-three.com>
+ */
+trait SearchableTrait
+{
+    /**
+     * Adds a sort scope.
+     *
+     * @param \Illuminate\Database\Eloquent\Builder $query
+     * @param array                                 $column
```
GrahamCampbell: oops

jbrooksuk: Whoopsy. I'll fix.

PR #1580, opened 2016-03-01
Scrolls the component list

### LINKED ISSUE #1500: The List Becomes Too Long
Hello , 

As of now we have only one option of having components under components group in vertical way .. But if i have around 50 components in each component group it becomes too big so can we also have an option or change it to horizontal tab . Just a suggestion .. Want to know why you have kept it in vertical way ..

jbrooksuk:
Closes #1500

---

Probably really ugly on Windows, but we can't do much about it other than use Lionbars or something?

### DIFF
--- a/resources/assets/sass/_status-page.scss
+++ b/resources/assets/sass/_status-page.scss
@@ -305,6 +305,11 @@ body.status-page {
                 margin-top: 5px;
             }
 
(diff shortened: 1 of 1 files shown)


joecohens:
Seems like a nice solution for now :smile:

Issue #1440, opened 2016-02-02
Stop using facades in bad places


GrahamCampbell:
It's particularly naughty to be using facades outside of controllers and views, in places where DI is happening, and then randomly a facade just appears, lol

jbrooksuk:
So... `CurrentUserComposer` for example?

GrahamCampbell:
Oh, nah, they're ok too, I was thinking like metrics and repositories.

GrahamCampbell:
Also, hacks like this need to stop:

![image](https://cloud.githubusercontent.com/assets/2829600/12763127/b3be70ae-c9ea-11e5-8f0e-55d76779abd8.png)

Infact, I'd go as far to say we should never be interfacing with even the dates libraries like that. I'm going to move stuff over to our date factory.

jbrooksuk:
I think that this is more or less done, we'll just remove them as we go.

PR #1591, opened 2016-03-07
Facade removal

jbrooksuk:


app/Http/Middleware/SetupAlreadyCompleted.php
```diff
@@ -12,13 +12,41 @@
 namespace CachetHQ\Cachet\Http\Middleware;
 
 use Closure;
+use Illuminate\Config\Repository;
```
GrahamCampbell: should be using the contract

GrahamCampbell:
This isn't really necessary imo.

GrahamCampbell:
When I say I want to avoid using the config facade, what i really mean is accessing the config repository in any way, inline like this.

jbrooksuk:
So then we can close #1440?

GrahamCampbell:
No. There's still some stuff that I don't like. Just injecting the dependencies doesn't fix that, it just adds more code.

PR #1609, opened 2016-03-13
Default to 2

jbrooksuk:
@GrahamCampbell no idea what the issue is, but this needs doing anyway.


GrahamCampbell:
Does making it tinyInteger instead of integer work?

jbrooksuk:
Not on MySQL at least, no.

PR #1617, opened 2016-03-15
[WIP] Upgrade to Chart.js 2.0-beta2

### LINKED ISSUE #1614: Metrics chart shows 3-4 tooltips on hovering data points in last hour
### Expected behaviour

Show one tooltip of the data point that is hovered.

### Actual behaviour

When hovering a data point in the last-hour resolution, 3-4 tooltips are shown:

screenshot from 2016-03-15 08-45-25

This is the same in both latest Firefox and Chrome stable releases.

### Steps to reproduce

Record data for every minute, show graph, switch to "last hour" resolution, hover a data point.

jbrooksuk:
Seeing whether this will fix #1614

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -4,7 +4,7 @@
         "animate-sass": "~0.6",
         "autosize": "~3.0.15",
         "bootstrap-sass": "~3.3",
-        "chartjs": "~1.0.2",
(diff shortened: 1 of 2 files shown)


jbrooksuk:
Tooltips no longer work, so forget this.

PR #1625, opened 2016-03-17
Upgrade to Chart.js 2.0-beta2

fabian:
Picked up https://github.com/CachetHQ/Cachet/pull/1617 and fixed the tooltip problem.

Will also fix https://github.com/CachetHQ/Cachet/issues/1614.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -4,7 +4,7 @@
         "animate-sass": "~0.6",
         "autosize": "~3.0.15",
         "bootstrap-sass": "~3.3",
-        "chartjs": "~1.0.2",
+        "chartjs": "~2.0.0-beta2",
         "eonasdan-bootstrap-datetimepicker": "~3.1",
         "humane-js": "~3.2",
         "ionicons": "~2.0",
--- a/resources/views/partials/metrics.blade.php
+++ b/resources/views/partials/metrics.blade.php
@@ -35,22 +35,9 @@
 </ul>
 <script>
 (function () {
-    Chart.defaults.global.pointHitDetectionRadius = 1;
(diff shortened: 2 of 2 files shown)


jbrooksuk:
Awesome, thank you!

GrahamCampbell:
This has broken the animations.

GrahamCampbell:
Also, the tooltips are really hard to get to show up unless you make a real effort with the mouse.

fabian:
Okay, will see if that can be fixed.

fabian:
I was able to improve the tooltips by increasing the hit radius, will send a PR.

@GrahamCampbell which animations do you refer to?

GrahamCampbell:
what we had before?

fabian:
I can see an animation in the rendering of the chart, the lines "grow" from the bottom. However that animation works for me both before and after this PR.

Before: 71d6e692eb4aa2a4e838fcef2008006c0f839e69

After: 
ffa1fdbf1aa319ed66cc89e5c3e463217de53e6d

If you are referring to another animation please describe the element the animation is happening on and what it does.

GrahamCampbell:
https://demo.cachethq.io/.

jbrooksuk:
The demo just seems laggy?

fabian:
Actually I just noticed it seems to be a browser problem. It works for me in Safari but not in Chrome and Firefox.

Even weirder when I set `responsiveAnimationDuration: 1000` it works in all browsers. The default value is 0 so it never should have worked in Safari. Probably a bug in Chart.js. Will send a PR with the new `responsiveAnimationDuration` value.

jbrooksuk:
:+1:

PR #1648, opened 2016-03-22
Navbar

jbrooksuk:
image

Really basic for now. I don't want to implement too many CSS changes before v2.2 is out. 

This only shows for single pages.

---

@joecohens maybe you want to do something more with this?


joecohens:
:+1: Nice!!!

PR #1652, opened 2016-03-23
[WIP] Scheduled maintenance refactoring

### LINKED ISSUE #495: Scheduled maintenance should have the ability to affect components
When creating a new scheduled status, the user should have the ability to assign one or more components that will be affected.

### LINKED ISSUE #715: Improving Scheduled Maintenance
I do not like how the scheduled maintenance is handled at present. Can I suggest a new way of handling it, much more like a timeline of events.

For example:

* I add scheduled maintenance to my Cachet install with a brief description. This is prior to the scheduled maintenance window.
* I then add updates as the maintenance is being carried out. For example, updating the issue when the maintenance has started and updating the issue as each section of the maintenance is completed.
* Updating the issue when the maintenance is completed with a time.

Here is an example of what I mean (hopefully it could be made to look a bit prettier):

dws

In many ways this suggestion is very much like #530 but for maintenance as well as incidents.

I love Cachet but I think maintenance and issues are too closely linked at the moment. This is shown most clearly when Cachet says that `Some Systems Are Experiencing Issues` when maintenance begins. What if I'm just adding a feature as part of the maintenance which does not impact the performance of the website/infrastructure at all?

Thanks!

### LINKED ISSUE #754: Display of Maintenances
It would be nice to be able to choose when to display maintenances. Currently, all maintenances are displayed even those planned in 3 or 4 weeks/months.

So it would be nice to have a setting like:

Display maintenances : 
    - All 
    - At scheduled date and time
    - X minutes/days/months before scheduled date and time

Thanks!

### LINKED ISSUE #1576: [API] cannot schedule a maintenance through the API

- [ ] You're running the [latest release](https://github.com/CachetHQ/Cachet/releases/latest) version of Cachet.
- [ ] Ensure that you're running at least PHP 5.5.9, you can check this by running `php -v`
- [ ] You've ran `rm -rf bootstrap/cache/*` from the root of your Cachet installation.

- [x] Running online version

### Expected behaviour

create a scheduled maintenance at the time specified

### Actual behaviour

creates an incident now

### Steps to reproduce

**send**
{
    "name": "test maintenance",
    "status": "0",
    "message": "test message",
    "scheduled_at": "2016-03-03 20:00:00"
}

**receive**
{
  "data": {
    "name": "test maintenance",
    "status": "0",
    "message": "test message",
    "visible": 1,
    "updated_at": "2016-02-29 21:05:58",
    "created_at": "2016-02-29 21:05:58",
    "id": 21,
    "notify": true,
    "human_status": "Scheduled",
    "component": null,
    "scheduled_at": "2016-02-29 21:05:58"
  }
}

### LINKED ISSUE #1607: Treat Maintenance and Issues as different event types
### Feature Request/Description

**TLDR: Separate API requests for maintenance and incident events. Treat them as two separate event classes.**

This is perhaps a bit general and might be something there are differing opinions on, but to me it is unintuitive that maintenance would be considered a type of incident. I think there should be two distinct URLs that return incidents and maintenance events separately (and perhaps if necessary or for backwards compatibility, a 3rd URL that returns both event types as [site]/api/v1/incidents does currently.)

My reasoning is this.. Maintenance isn't necessarily a "bad thing." There is a big difference between scheduled downtime and unexpected service interruption. If I told the leads in my org that there were 10 incidents this month, they might panic. However, if I told them that we had 9 scheduled maintenance activities and one brief outage incident, that sounds a lot less dire. From a reporting perspective, it makes sense to separate these two categories of data.

I'd love to hear any thoughts (including/especially disagreement) on this. I combed through the issue backlog and I do see a lot of enhancements coming up for scheduled maintenance, etc. but I didn't see this particular issue addressed. Apologies if I just missed it!
### Actual behaviour

Incidents and maintenance are essentially treated as one object type.
### Steps to reproduce

API calls, etc.

### LINKED ISSUE #1629: Incidents timeline should not show "scheduled at" once it fixed.
When a schedule Incident is fixed - on its timeline avoiding "scheduled at" is good.

resources/views/partials/incident.blade.php
Workaround check ($incident->humanStatus != 'Fixed') got me the fix.

```php
<strong>{{ $incident->name }}</strong>{{ $incident->isScheduled && $incident->humanStatus != 'Fixed' ? trans("cachet.incidents.scheduled_at", ["timestamp" => $incident->scheduled_at_diff]) : null }}
```

It would be best if it can be shown as "completed on" instead.

jbrooksuk:
# DO NOT RUN THIS BRANCH, IT'LL IRREVERSIBLY CORRUPT YOUR DATA.

Closes #754, closes #715, closes #495, closes #1576, closes #1629, closes #1607.

---

- [x] Separate schedules from core incidents.
- [x] Allow schedules to be have a start and end date.
- [ ] Single page for schedules.
- [x] Redesign scheduled maintenance view within main status page.
- [x] Add API for schedules


GrahamCampbell:

sirkkalap:
Hi brave CachetHQ, 

Is there now an API for scheduled maintenance? I can't find it in the API docs.

Should I just set "status" -> "0" and "scheduled_at" -> ""2017-03-20 02:00:00"" and get a future scheduled maintenance?

PR #1702, opened 2016-04-17
fix hidden (overlayed) tooltip

berlic:
moved component's icons tooltip to the right

before:

after:

### DIFF
--- a/resources/views/partials/component.blade.php
+++ b/resources/views/partials/component.blade.php
@@ -6,11 +6,11 @@
     @endif
 
     @if($component->description)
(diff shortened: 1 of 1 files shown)


jbrooksuk:
This doesn't actually properly fix the tooltip, it just moves it so that it's less hidden.

The CSS itself is what needs fixing here I'm afraid.

berlic:
I've replaced the commit with a proper fix, I think.
Changed [tooltip container](http://getbootstrap.com/javascript/#tooltips-options) to 'body'.

Looks like this:

jbrooksuk:
This is perfect! Could you rebase this please.

berlic:
Rebased

jbrooksuk:
Thank you!

PR #1705, opened 2016-04-19
Use Google Fonts subsets for localizations

rkgrep:
Before/after screenshot below

screenshot from 2016-04-19 11 20 10

### DIFF
--- a/app/Composers/AppComposer.php
+++ b/app/Composers/AppComposer.php
@@ -50,5 +50,7 @@ class AppComposer
         } else {
             $view->withSiteTitle(Config::get('setting.app_name'));
         }
+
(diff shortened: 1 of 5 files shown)


app/Composers/AppComposer.php
```diff
@@ -50,5 +51,7 @@ public function compose(View $view)
         } else {
             $view->withSiteTitle(Config::get('setting.app_name'));
         }
+
+        $view->withFontSubset(Config::get('langs.'.App::getLocale().'.subset', 'latin'));
```
jbrooksuk: Since the class is already using the `Config` facade, you could just call `Config::get('app.locale')`.

rkgrep: Ok!

jbrooksuk:
Awesome, thank you for this!

Issue #1205, opened 2015-12-06
Ditch catch QueryException Hax


GrahamCampbell:

jbrooksuk:
Is the plan here to throw specific exceptions?

GrahamCampbell:
Yes.

PR #1712, opened 2016-04-21
Tags API

### LINKED ISSUE #1700: No API for Component Tags
Component tags don't have an API.

jbrooksuk:
Closes #1700.

- [x] Commands for creating tags.
- [x] Commands for creating relations between tags and components

---

This adds two new API's:

1. Tags - for actually managing tags themselves.
2. Component Tags - for adding and deleting tags on a component. This is API differs from the others a little, as we don't use model bindings. This will be explained in the documentation.

It turns out that we were never using the `component_tag` table, so that's been fixed.

- It's now renamed to `component_tags`.
- It contains the usual timestamp columns.

Aside from that:

- The `Component` model now has many `tags` through the `component_tags` table.
- We're now starting to use "Bus Exceptions", which aims to make a start on #1205.
- The dashboard remains the same. In the future the tags field will be a taggable list of some sort for cleaner UX.

All that is left is to implement the new commands into the dashboard code.

### DIFF
--- a/app/Http/Controllers/Api/ComponentTagController.php
+++ b/app/Http/Controllers/Api/ComponentTagController.php
@@ -1,0 +1,117 @@
+<?php
+
+/*
+ * This file is part of Cachet.
+ *
+ * (c) Alt Three Services Limited
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace CachetHQ\Cachet\Http\Controllers\Api;
+
+use CachetHQ\Cachet\Bus\Commands\ComponentTag\AddComponentTagCommand;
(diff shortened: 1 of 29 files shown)


app/Http/Controllers/Api/ComponentTagController.php
```diff
@@ -0,0 +1,117 @@
...
+        try {
+            dispatch(new DeleteComponentTagCommand(
+                $component,
+                $tag
+            ));
+        } catch (TagDoesNotExistOnComponentException $e) {
+            throw new BadRequestHttpException();
+        }
+
+        return $this->noContent();
```
GrahamCampbell: is that 204, or 202?

jbrooksuk:
I'll open a new issue to add the dashboard code. This should be done once we have a new dashboard anyway.

PR #1715, opened 2016-04-22
Use the application/vnd.api+json content-type header

jbrooksuk:
Ping @CachetHQ/owners.


GrahamCampbell:
Nah. I don't like this. GitHub don't do that, right?

jbrooksuk:
According to http://jsonapi.org/recommendations/ it's the right thing to do?

GrahamCampbell:
That spec doesn't seem widely adopted though, and technically, it itself doesn't comply with the HTTP spec.

GrahamCampbell:
GitHub sets the content type to be application/json on their API, then has a special X-GitHub- header for the specific types/

jbrooksuk:
That's enough for me to say don't worry.

PR #1717, opened 2016-04-22
[WIP] Upgrade paths

jbrooksuk:


app/Console/Commands/UpgradeCommand.php
```diff
@@ -0,0 +1,88 @@
...
+
+        $this->upgrader = $upgrader;
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
```
jbrooksuk: This needs a way of forcing commands to run too, so that `app:update` will always run the update without prompting for confirmation.

GrahamCampbell:

PR #1753, opened 2016-05-05
Allow configuring of automatic localization

### LINKED ISSUE #1747: Disabling automatic localization
### My setup

- *What version of Cachet?* **2.2.2**
- *What database driver? MySQL? Postgres? SQLite?* **SQLite**
- *What version of PHP?* **5.5.9**

### Expected behaviour

The public facing interface (the status page itself) is not automatically translated based on the users' locale, unless I've explicitly enabled that feature.

### Actual behaviour

The public facing interface is always translated into the users' locale, even though I've specified another language as "site language" within the localization settings.

### Motivation

Don't get me wrong; I thing automatic localization is a great feature, in some cases. What I'm looking for is some way to disable it on a per-site basis. Like in our case, where we have all our communication in English, it looks quite strange for some of our Swedish customers when parts of the website is in Swedish while the actual information is in English.

Perhaps there is a feature to do this, but I simply haven't found it? :-) Or is it a bug that the language set in the settings isn't respected?

**Anyhow; many thanks for the great work put into Cachet!**

jbrooksuk:
Closes #1747

---

The setting is off by default.

### DIFF
--- a/app/Http/Middleware/Localize.php
+++ b/app/Http/Middleware/Localize.php
@@ -55,6 +55,11 @@ class Localize
      */
     public function handle(Request $request, Closure $next)
     {
+        // Early exit optimization.
+        if (!$this->config->get('setting.automatic_localization')) {
+            return $next($request);
+        }
+
         $supportedLanguages = $request->getLanguages();
         $userLanguage = $this->config->get('app.locale');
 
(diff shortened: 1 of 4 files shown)


jbrooksuk:
Rebased. Merge this first.

app/Http/Middleware/Localize.php
```diff
@@ -55,6 +55,11 @@ public function __construct(Repository $config)
      */
     public function handle(Request $request, Closure $next)
     {
+        // Early exit optimization.
```
GrahamCampbell: not exactly an optimization - it's a behaviour change

jbrooksuk: Details :trollface:

ManneW:
Awesome! Thanks again for this great project! :-)

PR #1754, opened 2016-05-05
Fix translations in the cachet.php file

jbrooksuk:
@joecohens when get some time, could you go through the Spanish translations again and see if anything else is in German?


joecohens:
Lol for sure!

PR #1761, opened 2016-05-06
When removing a component, clean up the subscriptions

### LINKED ISSUE #1739: Deleted stuffs cleanup wrt subscribers
> BTW, what happens if someone subscribes to something, and then it gets deleted later?

jbrooksuk:
Closes #1739 

---

I did think about an edge-case where a subscriber may no longer be subscribed to anything. Should we make them global? I don't think so, so I left that the same.


jbrooksuk:
@GrahamCampbell can you merge once this has passed please? I might be able to, but I'm heading out now.

joecohens:
👍

PR #1771, opened 2016-05-08
Specify components to subscribe to via API

### LINKED ISSUE #1685: Feature request: Add parameter to subscribers API to specify component
For those like me who are adding subscribers via de API it would be great to have the option to subscribe someone to a specific component (e.a. a clients server). Otherwise they will be subscribed to all updates.

https://docs.cachethq.io/docs/subscribers

jbrooksuk:
Closes #1685 

---

When subscribing a customer, you can supply a `components` parameter which is an array of component id's to subscribe to.

### DIFF
--- a/app/Bus/Commands/Subscriber/SubscribeSubscriberCommand.php
+++ b/app/Bus/Commands/Subscriber/SubscribeSubscriberCommand.php
@@ -33,6 +33,13 @@ final class SubscribeSubscriberCommand
     public $verified;
 
     /**
+     * The list of subscriptions to set the subscriber up with.
+     *
+     * @var array|null
+     */
+    public $subscriptions;
+
+    /**
      * The validation rules.
      *
      * @var array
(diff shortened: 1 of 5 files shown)


app/Bus/Commands/Subscriber/SubscribeSubscriberCommand.php
```diff
@@ -44,14 +51,16 @@
...
-     * @param string $email
-     * @param bool   $verified
+     * @param string     $email
+     * @param bool       $verified
+     * @param array|null $subscriptions
      *
      * @return void
      */
-    public function __construct($email, $verified = false)
+    public function __construct($email, $verified = false, $subscriptions = null)
```
GrahamCampbell: can put a typehint on this

jbrooksuk: I'll fix this in a couple of other places once it's merged then :)

PR #1794, opened 2016-05-20
Move packages to  devDependencies

vinkla:
Since we build the asset files with gulp we can move them to devDependencies instead.


jbrooksuk:
Can you do this on the `2.3` branch please? :)

vinkla:
Done #1795

PR #1820, opened 2016-05-25
Cleaned up the release class and provider

GrahamCampbell:


app/Foundation/Providers/AppServiceProvider.php
```diff
@@ -62,10 +64,25 @@ public function register()
...
+     *
+     * @return void
+     */
+    protected function registerRelease()
+    {
+        $this->app->singleton(Release::class, function ($app) {
+            $cache = $app['cache.store'];
+            $token = $app['config']->get('services.github.token')
+
+            return new Release($cache, $config);
```
jbrooksuk: Need to be sending `$token` as the second parameter.

Issue #1009, opened 2015-10-06
Customize email templates


peterlaws:
Sorry if this has been suggested already.

Currently, the email that subscribers get are very basic. Will there be a way to edit this, what it states, layout, etc?

eg.

Subject

> New Incident Reported.

Body

> New incident has been reported on Name Status Page.
> Investigating
> This is a test incident
> Name Status Page is powered by Cachet.
> Unsuscribe from email updates.
> Name Status Page

I also noticed that the email subject has "New Incident Reported" which is very ambiguous. Maybe include the incident title to the subject, as I do not see the title of the incident mentioned anywhere else?

wngravette:
Changing the layout/appearance of the emails would simply be a matter of editing the relevant views.

jbrooksuk:
@wngravette feel free to submit a PR with your suggested changes :)

jbrooksuk:
The email system isn't the best, we're completely aware :)

wngravette:
No problem, I may just even be able to whip something up within the next few days. Cheers.

jbrooksuk:
Thanks @wngravette, I'm moving this to 2.1.0 but if you're able to do it before we ship, that'd be awesome.

peterlaws:
> Changing the layout/appearance of the emails would simply be a matter of editing the relevant views.

I assume the email subjects are hardcoded, right?

wngravette:
Indeed, at app/Handlers/Events/ in the relevant files.

jbrooksuk:
Renamed this issue.

AntoineAugusti:
@peterlaws Are you going to tackle this? If not, what do we need to do precisely?

jbrooksuk:
@AntoineAugusti:

1. Store email contents in a table.
2. Use Twig or something to allow customization, unless we want to stick with Blade? Twig is a good templating language for things like this.
3. We'll need a panel in the dashboard that allows editing of templates.
4. Change the email handling code to first grab the content, compile and then send.

AntoineAugusti:
@jbrooksuk This seems a lot of work. Are you sure a lot of people are going to want to edit the basic email template?

jbrooksuk:
@AntoineAugusti it is, which is why we've not done it ourselves yet :)

srikanth88infy:
May I know if there are any updates regarding this feature? I would love to know how can I tweak a basic template. Can anyone point me how can I achieve this?

jbrooksuk:
[...]

PR #1841, opened 2016-05-26
[WIP] Added email subject to incident model and form

bmsauer:
I added an extra field for email subject lines for incidents.  It stores it in the incident table.  This is somewhat related to https://github.com/cachethq/Cachet/issues/1009 .


.gitignore
```diff
@@ -2,3 +2,4 @@
 node_modules
 phpunit.xml
 vendor
+*~
```
GrahamCampbell: please erert that change

resources/views/dashboard/incidents/add.blade.php
```diff
@@ -115,6 +115,10 @@
                                 {{ trans('forms.incidents.notify_subscribers') }}
                             </label>
                         </div>
+			<div class="form-group">
```
GrahamCampbell: Please use spaces for indentation.

GrahamCampbell:
Please fix the failing travis tests.

database/migrations/2016_05_26_131224_add_incident_email_subject_column.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('incidents', function ($table) {
+            $table->string('email_subject');
```
jbrooksuk: This column should be added before the timestamps fields, but after `message`.

jbrooksuk:
What is the use case for this?

jbrooksuk:
In my opinion, we should make the subjects standardised, such as "New incident, `$title`".

.gitignore
bmsauer: This is to ignore out emacs autosave files.

database/migrations/2016_05_26_131224_add_incident_email_subject_column.php
bmsauer: I'm not sure I understand what you mean, can you explain more?

bmsauer:
I'm working on doing the tests now.  As for the use case, we just want to be specific for our customers, as some incidents within components still only affect a subset of the subscribers.

.gitignore
GrahamCampbell: Please ad this to your global git ignore file, and revert this change.

GrahamCampbell:
Thanks, but this isn't something we want to add to Cachet right now. Do check out the recent changes we've just merged though. ;)

PR #1847, opened 2016-05-28
Update gitignore

splitice:
update gitignore to ignupdate gitignore to ignore PHPStorm / IntelliJ IDE files & composer.phar & composer.lock

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 node_modules
 phpunit.xml
 vendor
+.idea
+.composer.phar
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,4 @@ node_modules
 phpunit.xml
 vendor
 .idea
-.composer.phar
+*.phar
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ phpunit.xml
 vendor
 .idea
 *.phar
+composer.lock


jbrooksuk:
Please don't modify the `.gitignore` file. Anything that is generated by an IDE should be ignored by your user-based `.gitignore` file but not on a per-project basis.

The `composer.lock` file is needed for software like Cachet because we need to lock the dependencies down to `x` so that when Person A installs Cachet, Person B-Infinite will have the exact same installation. This is crucial. You can ignore this file for packages, but not software.

splitice:
Thats what a well formed composer.json is for. composer.lock is build metadata which is not part of the project. The state of dependencies should be controlled through how you're versioning them (either manually or as part of your automated build process) and not arbitrarily by the last developer to update them and commit the lock file.

If you are concerned about your dependencies changing between composer updates then you have a lack of confidence in your versioning scheme. Versions (1.0, 1.1, 1.2, etc) should be immutable and you should avoid "dev-" and "X.*" wildcards outside of initial feature development.

Committing the lock file is a regression for your dependency management system as the dependency version has now gone back to being implicity defined.

Anyway do as you like. I'll keep it in my fork.

jbrooksuk:
> Thats what a well formed composer.json is for. composer.lock is build metadata which is not part of the project. The state of dependencies should be controlled through how you're versioning them (either manually or as part of your automated build process) and not arbitrarily by the last developer to update them and commit the lock file.

You're not quite right there. The `composer.lock` file literally locks dependencies at a given version. If we want to control the version of a dependency that someone uses we need to lock them. Not committing the `composer.lock` file would potentially result in people having different versions of every dependency.

Actually, Composer's very own documentation suggests [committing this file](https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file).

As I said, you should never commit the lock file for a package that someone else is installing, this would be very bad. But for software, you absolutely should.

splitice:
That statement reads word for word correct to me. Anyway, VCS best practices of higher importance to most organizations than the documentation of a specific piece of software.

composer.json IS build metadata it is the state of the dependencies for a particular system (exact versions). composer.json is a contract, a list of versions that are supported. While you may be providing software, ultimately its the composer.json file that matters, not the .lock file. If the composer.json allows for the installation of a version of a package that is not supported (either technically, or from lack of testing) then its a fault in the project's composer.json description. Not due to a lack of lock file.

If you are using the lock file as versioning perhaps you need to look about better defining requirements in composer.json?

Furthermore, many organizations (including the one this work is conducted for) have quality checks that check for files generated in the build process like this being committed.

Regardless, it doesnt matter me what you do. It really doesnt affect me, I viewed the commit of this file as a mistake, so I PR'ed to fix it - and some other common exclusions for an IDE with a large market share.

I'll leave it here.

GrahamCampbell:
We are currently using the best practices. You need to use them too, and use a global gitignore file.

PR #1851, opened 2016-05-28
Fix crap login throttling

### LINKED ISSUE #1849: Improve throttling
`Illuminate\Routing\Middleware\ThrottleRequests` is a bit crap. I'm going to do a new major version of my throttling package that uses Laravel's native stuff though. We can switch back to that.

GrahamCampbell:
Uses https://github.com/AltThree/Throttle, which still uses Laravel's rate limiter under the hood.

---

Closes #1849.

### DIFF
--- a/app/Foundation/Exceptions/Displayers/ThrottleDisplayer.php
+++ b/app/Foundation/Exceptions/Displayers/ThrottleDisplayer.php
@@ -1,0 +1,88 @@
+<?php
+
+/*
+ * This file is part of Cachet.
+ *
+ * (c) Alt Three Services Limited
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace CachetHQ\Cachet\Foundation\Exceptions\Displayers;
+
+use Exception;
+use GrahamCampbell\Exceptions\Displayers\DisplayerInterface;
+use Illuminate\Http\Request;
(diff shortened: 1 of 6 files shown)


app/Foundation/Exceptions/Displayers/ThrottleDisplayer.php
```diff
@@ -0,0 +1,88 @@
...
+ */
+
+namespace CachetHQ\Cachet\Foundation\Exceptions\Displayers;
+
+use Exception;
+use GrahamCampbell\Exceptions\Displayers\DisplayerInterface;
+use Illuminate\Http\Request;
+use Symfony\Component\HttpKernel\Exception\TooManyRequestsHttpException;
+
+class ThrottleDisplayer implements DisplayerInterface
```
GrahamCampbell: This class basically says, if throttling occurred on an auth route, go back to the login page with a translated rate limit flash message.

resources/lang/en/forms.php
```diff
@@ -34,6 +34,7 @@
         'invalid'       => 'Invalid username or password',
         'invalid-token' => 'Invalid token',
         'cookies'       => 'You must enable cookies to login.',
+        'rate-limit'    => 'Rate limit exceeded.',
```
GrahamCampbell: @jbrooksuk Note the addition of this translation.

PR #1869, opened 2016-06-01
Updated login page

GrahamCampbell:


resources/views/auth/login.blade.php
```diff
@@ -3,67 +3,46 @@
 @section('bodyClass', 'login')
 
 @section('content')
-    <div class="container">
-        <div class="row">
-            <div class="form-bg">
-                <div class="logo">
-                    <svg viewBox="0 0 1150 296" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:sketch="http://www.bohemiancoding.com/sketch/ns">
```
jbrooksuk: I'm glad this has gone!

jbrooksuk:
Merging this so all the images are local. I'm not too fussed. If a company want to remove their logo, they can ask and it'll be gone in a minor patch version.

We'll update this image anyway.

GrahamCampbell:
We have to include every companies license in here by law.

jbrooksuk:
Can we not do that in the LICENSE file?

GrahamCampbell:
No. Some we actually are not allowed to put in here anyway if they have trademarks.

PR #1882, opened 2016-06-02
Added Team Member API

### LINKED ISSUE #1822: User management from API
Hello,

Since the LDAP auth approach was scrubbed is there any possibility that we as an additional feature
could add the ability to manage users from the API?
Having a large organization and being able to integrate it with other systems is crucial or most enterprises.

Best regards

jbrooksuk:
Closes #1822

### DIFF
--- a/app/Bus/Commands/User/UpdateTeamMemberCommand.php
+++ b/app/Bus/Commands/User/UpdateTeamMemberCommand.php
@@ -11,8 +11,22 @@
 
 namespace CachetHQ\Cachet\Bus\Commands\User;
 
-final class AddTeamMemberCommand
+use CachetHQ\Cachet\Models\User;
+
+/**
+ * This is the update team member command.
+ *
+ * @author James Brooks <james@alt-three.com>
+ */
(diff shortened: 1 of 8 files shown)


app/Bus/Commands/User/AddTeamMemberCommand.php
```diff
@@ -48,8 +48,9 @@
      */
     public $rules = [
         'name'     => 'required|string',
-        'password' => 'string',
-        'level'    => 'int',
+        'email'    => 'required|email',
+        'password' => 'required|string',
+        'level'    => 'int|min:1|max:2',
```
GrahamCampbell: In all our rules on L5.3, if they're not required, we need to add `nullable` to them.

GrahamCampbell: That's something I didn't upgrade, but need to do.

jbrooksuk: I'll do that as a separate PR later.

GrahamCampbell:

PR #1901, opened 2016-06-08
[WIP] Add user management to API

tobias-urdin:
WIP to add user management to API as issue #1822 suggests

Trying to get this fixed in order to get Cachet up and running in our enterprise. First time using Laravel, please give input on what can be better. Haven't implemented the PUT to update users, a bus command is needed.

### DIFF
--- a/app/Http/Controllers/Api/AbstractApiController.php
+++ b/app/Http/Controllers/Api/AbstractApiController.php
@@ -175,6 +175,16 @@ abstract class AbstractApiController extends Controller
     }
 
     /**
+     * Respond with a 401 unauthorized.
+     *
+     * @return \Illuminate\Http\JsonResponse
+     */
+    protected function unauthorized()
+    {
+        return $this->setStatusCode(401)->respond();
+    }
+
+    /**
      * Build the response.
      *
      * @return \Illuminate\Http\Response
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Api/AbstractApiController.php
```diff
@@ -175,6 +175,16 @@ protected function noContent()
     }
 
     /**
+     * Respond with a 401 unauthorized.
+     *
+     * @return \Illuminate\Http\JsonResponse
+     */
+    protected function unauthorized()
+    {
+        return $this->setStatusCode(401)->respond();
```
GrahamCampbell: This does not look correct. You should throw an exception.

app/Http/Controllers/Api/AbstractApiController.php
```diff
@@ -175,6 +175,16 @@ protected function noContent()
     }
 
     /**
+     * Respond with a 401 unauthorized.
+     *
+     * @return \Illuminate\Http\JsonResponse
+     */
+    protected function unauthorized()
```
GrahamCampbell: Infact, we don't need this method here.

GrahamCampbell:
Please target the develop branch. 3.0 now lives there.

Issue #1936, opened 2016-07-05
API endpoint to get System Status


odannyc:
Can we get an endpoint that just gets the entire system status message.
For example, if all systems are operational the message the endpoint should return

```json
{
  "message": "All systems are operational"
}
```

jbrooksuk:
Yeah, I don't see why not :+1:

adammbalogh:
Nice idea!

Could I create a PR for this feature?

jbrooksuk:
I'd love for you to! Could you base it on the 2.3 branch please? That'll see it gets into the v2.4.0 release.

adammbalogh:
#1943

jbrooksuk:
Closed by https://github.com/CachetHQ/Cachet/pull/1968

Daniel15:
Is there an ETA for getting this into a released version of Cachet?

PR #1942, opened 2016-07-06
Add the /api/v1/status endpoint

odannyc:
To get the system status all you need to do now is:
`/api/v1/status`

And it'll return:

```json
{
  "data": {
    "system_status": "success",
    "system_message": "System operational",
    "favicon": "favicon"
  }
}
```

### DIFF
--- a/app/Http/Controllers/Api/GeneralController.php
+++ b/app/Http/Controllers/Api/GeneralController.php
@@ -12,6 +12,8 @@
 namespace CachetHQ\Cachet\Http\Controllers\Api;
 
 use CachetHQ\Cachet\Integrations\Releases;
+use CachetHQ\Cachet\Models\Component;
+use CachetHQ\Cachet\Models\Incident;
 
 /**
  * This is the general api controller.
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Api/GeneralController.php
```diff
@@ -44,4 +46,48 @@ public function version()
...
+        } elseif (Component::enabled()->notStatus(1)->count() === 0) {
+            // If all our components are ok, do we have any non-fixed incidents?
+            $incidents = Incident::notScheduled()->orderBy('created_at', 'desc')->get();
+            $incidentCount = $incidents->count();
+
+            if ($incidentCount === 0 || ($incidentCount >= 1 && (int) $incidents->first()->status === 4)) {
+                $withData = [
+                    'system_status'  => 'success',
+                    'system_message' => trans_choice('cachet.service.good', $totalComponents),
+                    'favicon'        => 'favicon',
```
GrahamCampbell: i'm a bit confused why we need that?

app/Http/Controllers/Api/GeneralController.php
```diff
@@ -44,4 +46,48 @@ public function version()
...
+
+    /**
+     * Show the general system status.
+     *
+     * @return \Illuminate\Http\JsonResponse
+     */
+    public function status()
+    {
+        $totalComponents = Component::enabled()->count();
+        $majorOutages = Component::enabled()->status(4)->count();
```
GrahamCampbell: Is this code taken from the other controller? If so, we should probably reuse it, not copy it.

app/Http/Controllers/Api/GeneralController.php
odannyc: It's the same check that the main status page uses:
`app/composers/StatusPageComposer.php`

GrahamCampbell: Yeh, but from the API it doesn't have an actual use case?

odannyc: Yeah, on the page: https://social.ringplus.net/categories/community there's a section that checks a cachet status page and just displays the same exact message the cachet status page displays, on the backend of that site it gets all components and loops through them to see if any of them are not functional which then displays the message from there. It'll be better if I can just use this API endpoint to get the message directly. Also check the Issue #1936

jbrooksuk:
The favicon can probably be dropped?

GrahamCampbell:
> The favicon can probably be dropped?
 
See the earlier comments. ;)

GrahamCampbell:
Replaced by https://github.com/CachetHQ/Cachet/pull/1945 I think?

PR #1945, opened 2016-07-07
Add the /status endpoint

### LINKED ISSUE #1936: API endpoint to get System Status

adammbalogh:
Closes #1936

---

The API endpoint is ```/status```for getting all kinds of status messages. Right now it looks like this:
```json
{
  "system_status": "danger",
  "system_message": "The service experiencing a major outage"
}
```

(It still needs unit tests.)

### DIFF
--- a/app/Foundation/ViewObjects/SystemStatusViewObject.php
+++ b/app/Foundation/ViewObjects/SystemStatusViewObject.php
@@ -1,0 +1,60 @@
+<?php
+
+/*
+ * This file is part of Cachet.
+ *
+ * (c) Alt Three Services Limited
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace CachetHQ\Cachet\Foundation\ViewObjects;
+
+use CachetHQ\Cachet\Models\Component;
+use CachetHQ\Cachet\Foundation\Queries\SystemStatusQuery;
+
+class SystemStatusViewObject
+{
+    /**
+     * @var SystemStatusQuery
+     */
+    private $systemStatusQuery;
+
+    /**
+     * @param SystemStatusQuery $systemStatusQuery
+     */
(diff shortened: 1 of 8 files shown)


GrahamCampbell:
Please don't put things in the foundation namespace.

app/Foundation/ViewObjects/SystemStatusViewObject.php
```diff
@@ -0,0 +1,60 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace CachetHQ\Cachet\Foundation\ViewObjects;
+
+use CachetHQ\Cachet\Models\Component;
+use CachetHQ\Cachet\Foundation\Queries\SystemStatusQuery;
+
+class SystemStatusViewObject
```
GrahamCampbell: View object isn't really right. It's not used in the context of views exclusively?

adammbalogh: Yeah, it's not quite right. I'm thinking about how we should name this object.

GrahamCampbell: We don't really need an object tbh. Just pass about an array.

adammbalogh:
About foundation namespace. Sorry about that, I'm just learning the project with this PR. How we should structure my stuff, do you have an idea?

GrahamCampbell:
I'll have to get back to you later. At work right now.

app/Foundation/ViewObjects/SystemStatusViewObject.php
adammbalogh: That can work but how about a transformer object maybe?

jbrooksuk:
Replaced by #1968 :) much love to you for this @adammbalogh.

PR #1947, opened 2016-07-08
Set Date locale

### LINKED ISSUE #1727: Dates not being localized
I've set the locale to norwegian in `.env` as well as under settings in the dashboard, however, dates are still being formatted in English.

If I `php artisan tinker` everything seems to work as you'd expect, dates are translated:

```
>>> Jenssegers\Date\Date::getLocale();
=> "no"
>>> $df = app('CachetHQ\Cachet\Dates\DateFactory');
=> CachetHQ\Cachet\Dates\DateFactory {#888}
>>> $df->make('2016-04-27 12:00:00')->format('l j. F Y');
=> "onsdag 27. april 2016"
```

However, if I `var_dump($dt->getLocale())` inside `IncindentPresenter::create_at_formatted`, it shows the locale is "en".

jbrooksuk:
Fixes #1727 

---

I think this is all that we need to do? By default we're overriding the `Date` locale, but then the `Localize` middleware is free to change it again to the visitors locale, if the setting is enabled.

@CachetHQ/owners agree?

### DIFF
--- a/app/Foundation/Providers/ConfigServiceProvider.php
+++ b/app/Foundation/Providers/ConfigServiceProvider.php
@@ -16,6 +16,7 @@ use CachetHQ\Cachet\Settings\Cache;
 use CachetHQ\Cachet\Settings\Repository;
 use Exception;
 use Illuminate\Support\ServiceProvider;
+use Jenssegers\Date\Date;
 
 /**
  * This is the config service provider class.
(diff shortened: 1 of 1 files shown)


joecohens:
Yeah I think at this point we are resetting it. If you take a look at the DateServiceProvider it takes the current locale at that point. But we then switch it again at the ConfigServiceProvider

jbrooksuk:
So you agree that this should fix it too?

joecohens:
Yeah I this seems like the right fix

Issue #1916, opened 2016-06-20
Internal server error at setup page


xedro00:
I just followed the installation guide and installed it but when I visit my site it takes me to

www.mysite.com/setup and shows an Internal server error this is my log below.

http://pastebin.com/stTTKA0L

hope I can find a solution to this!

jbrooksuk:
Did you run `php artisan app:install`?

xedro00:
Yep! but no luck still the same.
http://puu.sh/pzDhh/96b96ae335.png

GrahamCampbell:
Can you show us the output of `php artisan app:install` please? Are you running the latest 2.3 RC release?

jbrooksuk:
The permissions are correct on `./bootstrap` and `./storage`?

xedro00:
php artisan app:install
Clearing settings cache...
Settings cache cleared!
Backing up database...
Dumping database and uploading...

Successfully dumped mysql, compressed with gzip and store it to local at /home/bitsnap/backend/database/backups/2016-06-21 02.22.11
Backup completed!
Application key [base64:RRIWRZNE/KkajUMu3HBU70ZMBRfIYm3k+VBOuCPoXPg=] set successfully.
Configuration cache cleared!
Configuration cached successfully!
Route cache cleared!
Routes cached successfully!
Copied Directory [/vendor/roumen/feed/src/views] To [/resources/views/vendor/feed]
Publishing complete for tag []!
Nothing to migrate.
Clearing cache...
Application cache cleared!
Cache cleared!


Permission is correct for those folders.

jbrooksuk:
Can you send us the full output of the most recent log file please?

xedro00:
Here you go!

https://www.cyph0.com/paste/paste2321.log

GrahamCampbell:
Looks like some of our files are missing from the public folder. Did you install them correctly?

xedro00:
yes  I have uploaded every files!

xedro00:
do I need to set permission for any folder in public?

jbrooksuk:
Have you run `gulp`? (Don't do so, if you haven't).

xedro00:
Nope! i have not run that!

xedro00:
any update on this?

jbrooksuk:
Try installing again as you seem to be missing the `./public/build` files, which **definitely** ship with Cachet, so it looks like your installation has gone crazy.

xedro00:
[...]

PR #1953, opened 2016-07-11
fix setup env writing

joecohens:
This will fix https://github.com/CachetHQ/Cachet/issues/1916#issuecomment-231407142

/cc @jbrooksuk


GrahamCampbell:
Sweet.

xedro00:
do I need to do anything else after editing the following code? coz still it shows same error.

PR #1957, opened 2016-07-12
Update the font-stack to be -apple-system font

jbrooksuk:
This only affects macOS systems, but makes it feel more native.

image


jbrooksuk:
FWIW, this looks way better on Safari.

joecohens:
👍 nice addition!

PR #1959, opened 2016-07-12
remove executable permission from the files don't need it

PeterDaveHello:


jbrooksuk:
This is something that I need to be more careful with.

PeterDaveHello:
Should do it by gulp, currently the built files seem to be executable.

PR #1971, opened 2016-07-12
[WIP] Metric repository perf

### LINKED ISSUE #1900: Non-optimal app layer iteration in MetricRepository
- [X] You're running the [latest release](https://github.com/CachetHQ/Cachet/releases/latest) version of Cachet.
- [X] Ensure that you're running at least PHP 5.5.9, you can check this by running `php -v`
- [X] You've ran `rm -rf bootstrap/cache/*` from the root of your Cachet installation.

### Your setup
- _What version of Cachet?_ v2.3.0-RC5
- _What database driver? MySQL?_ 5.5.47
- _What version of PHP?_ 5.6
### Expected behaviour

Iterating over the metric points should be quite performant.
### Actual behaviour

I've noticed in the MetricRepository you are iterating over each minute/hour in the app layer - which will potentially generate between 12 - 60 queries of an average return time of 4-8 seconds each - depending on the data size. 

Subsequently, we're seeing our MySql process jump to 100% CPU time once the Metric Points table gets close to 40,000 records

Wouldn't it be more performant to iterate this in the query itself? ie.MySql example

`
SELECT AVG(mp.value * mp.counter) AS value
FROM metrics m
  INNER JOIN metric_points mp ON m.id = mp.metric_id
WHERE m.id = 1 AND DATE_SUB(curdate(), INTERVAL 12 HOUR) <= mp.created_at
GROUP BY HOUR(mp.created_at)
`

jbrooksuk:
Closes #1900.

---

Refactoring metrics. This is probably going to end up bigger than I intended it to be, but it'll make it easier to deal with.


GrahamCampbell:
Needs retargeting at the 2.4 branch, if we still want this in 2.4.

jbrooksuk:
We do (if I can get it all done).

GrahamCampbell:
I'll leave this one, for now, and we can retarget it at a later point if we think we can get it in.

PR #1976, opened 2016-07-15
Don't use settings caching in cli

### LINKED ISSUE #1935: Only allow the web server to write to the bootstrap/cachet folder
People keep screwing up the file permissions by running artisan commands as root. As an attempt to mitigate this issue, we can only allow the web server to write to our setting cache

GrahamCampbell:
This should avoid the file permission issues RE people running commands as root by mistake.

---

What do you make of this @jbrooksuk and @joecohens please?

---

Closes #1935.


jbrooksuk:
:shipit:

PR #1993, opened 2016-07-20
Added option to skip subscriber verification

### LINKED ISSUE #1990: Skip subscriber verification
Hi Team,

I am using the CACHE ( 2.3.0-RC4) version. & am trying to skip the CONFIRM SUBSCRIPTION when an user subscribes to the system 

I would really appreciate if you can provide a solution to my issue?

Let me know about your thoughts.

Thanks,
Karthik

jbrooksuk:
Closes #1990 

---

Ping @joecohens


app/Bus/Handlers/Events/Subscriber/SendSubscriberVerificationEmailHandler.php
```diff
@@ -25,15 +27,23 @@ class SendSubscriberVerificationEmailHandler
     protected $mailer;
 
     /**
+     * The illuminate config instance.
+     *
+     * @var \Illuminate\Contracts\Config\Repository
+     */
+    protected $config;
```
GrahamCampbell: We should never need to read config from inside handlers.

joecohens:
I think we can do this at the controller, just use the config here https://github.com/CachetHQ/Cachet/blob/2.3/app/Http/Controllers/SubscribeController.php#L61 what do you think?

GrahamCampbell:
Also, this implementation was broken with the API before. :P

GrahamCampbell:
This is not pretty, but I think a bit better now. What do you think?

GrahamCampbell:
@jbrooksuk or @joecohens merge if you're happy with this. :)

PR #2015, opened 2016-07-28
Set the step size to the metrics places

### LINKED ISSUE #2012: Y-axis scale precision not consistent with decimal value of metrics
### Expected behaviour
In case of multiple metrics too, Y-axis scale precision must be set to metric's decimal value respectively.

### Actual behaviour
Y scale precision doesn't get reflected unless the **last** metric's decimal value is set > 0

### Steps to reproduce

Reproducible on [Cachet Demo](https://demo.cachethq.io/)

1. Create a new metric with decimal value > 0  ( . . say '2')
2. After step 1, create another metric with decimal value = 0
3. Visit the status page. Both the Y-axes will show non-decimal scale (Not expected)
4. Now, update decimal value of the last metric created to '2'.
5. Visit the status page. Now both the Y-axes will show decimal scale (Expected)

Tried this with different permutations as well.

jbrooksuk:
Fixes #2012 

---

I think this fixes the issue. Y-axis steps are now set per-metric.

### DIFF
--- a/resources/views/partials/metrics.blade.php
+++ b/resources/views/partials/metrics.blade.php
@@ -25,7 +25,7 @@
         </div>
         <div class="row">
             <div class="col-xs-12">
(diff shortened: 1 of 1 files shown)


joecohens:
👍

GrahamCampbell:
:bomb:

PR #2026, opened 2016-08-03
Use class annotation to register event handlers

jbrooksuk:
@CachetHQ/owners this is more a proof of concept than anything else... It works though!

### DIFF
--- a/app/Annotations/HandlerAnnotation.php
+++ b/app/Annotations/HandlerAnnotation.php
@@ -1,0 +1,29 @@
+<?php
+
+/*
+ * This file is part of Cachet.
+ *
+ * (c) Alt Three Services Limited
+ *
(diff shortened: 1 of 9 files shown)


GrahamCampbell:
I think we should instead add this the other way around, and allow the event listeners to define what they listen to.

GrahamCampbell:
We definitely don't want to do this at runtime I think. We'll need to pre-compute this.

jbrooksuk:
> We definitely don't want to do this at runtime I think. We'll need to pre-compute this.

Fo sho. It's slow.

jbrooksuk:
I'm closing this because the implementation is both slow and incorrect.

ConnorVG:
And the idea sucks :trollface:

PR #2028, opened 2016-08-04
Beacons

jbrooksuk:
Ping @CachetHQ/owners.


tests/Bus/Jobs/System/SendBeaconJobTest.php
```diff
@@ -0,0 +1,48 @@
...
+ * @author James Brooks <james@alt-three.com>
+ */
+class SendBeaconJobTest extends AbstractTestCase
+{
+    use JobTrait;
+
+    /**
+     * @before
+     */
+    public function setEventExpectations()
```
GrahamCampbell: No need for this if there's nothing to expect.

jbrooksuk: I'll clean these up once merged then, I think I do it elsewhere at the moment.

PR #2037, opened 2016-08-07
Send component update email on status changes

### LINKED ISSUE #1989: Warn subscribers only when component state changes
Before submitting your issue, please make sure that you've checked the checkboxes below.

- [x] I am running the [2.3.6](https://github.com/CachetHQ/Cachet/releases/latest) version of Cachet.
- [x] I am running at least PHP 5.5.9. *You can check this by running `php -v`.*
- [x] I have ran `rm -rf bootstrap/cache/*`.

### Expected behaviour

I have an external component which uses the API to issue various components status updates, I think that when an API call is sent which actually does -not- change the status, subscribers should not be warned as nothing changed. I understand that I can refactor my stuff, but I still think there should be at least an optional way to disable email updates when component status is not updated, instead of whenever an API PUT is made.

### Actual behaviour

Emails are sent to subscribers whenever a PUT is made to the component even if the status is the same.

### Steps to reproduce

From an initial operational status, update to Major Outage (get an email), update to Operational (get an email), update to Operational again (get another email).

jbrooksuk:
Fixes #1989

---

Ping @CachetHQ/owners. To me, this makes *way* more sense.

### DIFF
--- a/app/Bus/Events/Component/ComponentStatusWasUpdatedEvent.php
+++ b/app/Bus/Events/Component/ComponentStatusWasUpdatedEvent.php
@@ -1,0 +1,59 @@
+<?php
+
+/*
+ * This file is part of Cachet.
+ *
+ * (c) Alt Three Services Limited
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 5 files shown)


joecohens:
👍

PR #2038, opened 2016-08-07
Stop using facades in presenters

jbrooksuk:
Part of #1440.

### DIFF
--- a/app/Presenters/ComponentPresenter.php
+++ b/app/Presenters/ComponentPresenter.php
@@ -12,14 +12,36 @@
 namespace CachetHQ\Cachet\Presenters;
 
 use CachetHQ\Cachet\Dates\DateFactory;
+use CachetHQ\Cachet\Models\Component;
 use CachetHQ\Cachet\Presenters\Traits\TimestampsTrait;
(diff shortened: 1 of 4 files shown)


app/Presenters/ComponentPresenter.php
```diff
@@ -12,16 +12,38 @@
...
+     *
+     * @param \CachetHQ\Cachet\Models\Component       $resource
+     * @param \Illuminate\Contracts\Config\Repository $config
+     *
+     * @return void
+     */
+    public function __construct(Component $resource, Repository $config)
+    {
+        $this->wrappedObject = $resource;
+        $this->config = $config;
```
GrahamCampbell: Should stop using config in here completely tbh. It's not the use of facades that bothers me as such, it's the fact they hide usage like this.

GrahamCampbell: I'd prefer if as few objects as possible knew config existed. Perhaps add a method to the date factory?

jbrooksuk: Makes sense.

PR #2040, opened 2016-08-08
Metric repository perf

### LINKED ISSUE #1900: Non-optimal app layer iteration in MetricRepository

jbrooksuk:
Closes #1900.

---

Refactoring metrics. With this, I intend to fix the performance of metrics across all three database drivers.

Note that this is probably going to end up bigger than I intended it to be, but it'll make it easier to deal with.

Todo:

- [x] Refactor all metric repositories to query all metric points between a given time span.
- [x] Stop using static table names as the tables may be prefixed.

### DIFF
--- a/app/Foundation/Providers/RepositoryServiceProvider.php
+++ b/app/Foundation/Providers/RepositoryServiceProvider.php
@@ -13,9 +13,9 @@ namespace CachetHQ\Cachet\Foundation\Providers;
 
 use CachetHQ\Cachet\Dates\DateFactory;
 use CachetHQ\Cachet\Repositories\Metric\MetricRepository;
-use CachetHQ\Cachet\Repositories\Metric\MySqlRepository as MetricMySqlRepository;
-use CachetHQ\Cachet\Repositories\Metric\PgSqlRepository as MetricPgSqlRepository;
-use CachetHQ\Cachet\Repositories\Metric\SqliteRepository as MetricSqliteRepository;
+use CachetHQ\Cachet\Repositories\Metric\MySqlRepository;
+use CachetHQ\Cachet\Repositories\Metric\PgSqlRepository;
+use CachetHQ\Cachet\Repositories\Metric\SqliteRepository;
 use Illuminate\Contracts\Config\Repository as ConfigRepository;
 use Illuminate\Contracts\Container\Container;
 use Illuminate\Support\ServiceProvider;
@@ -46,14 +46,11 @@ class RepositoryServiceProvider extends ServiceProvider
     {
         $this->app->singleton(MetricRepository::class, function (Container $app) {
             $config = $app->make(ConfigRepository::class);
-            $driver = $config->get('database.default');
 
-            if ($driver == 'mysql') {
-                $repository = new MetricMySqlRepository($config);
-            } elseif ($driver == 'pgsql') {
-                $repository = new MetricPgSqlRepository($config);
-            } elseif ($driver == 'sqlite') {
-                $repository = new MetricSqliteRepository($config);
+            switch ($config->get('database.default')) {
+                case 'mysql': $repository = new MySqlRepository($config); break;
+                case 'pgsql': $repository = new PgSqlRepository($config); break;
+                case 'sqlite': $repository = new SqliteRepository($config); break;
             }
 
             $dates = $app->make(DateFactory::class);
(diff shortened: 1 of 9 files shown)


app/Foundation/Providers/RepositoryServiceProvider.php
```diff
@@ -13,9 +13,9 @@
 
 use CachetHQ\Cachet\Dates\DateFactory;
 use CachetHQ\Cachet\Repositories\Metric\MetricRepository;
-use CachetHQ\Cachet\Repositories\Metric\MySqlRepository as MetricMySqlRepository;
-use CachetHQ\Cachet\Repositories\Metric\PgSqlRepository as MetricPgSqlRepository;
-use CachetHQ\Cachet\Repositories\Metric\SqliteRepository as MetricSqliteRepository;
+use CachetHQ\Cachet\Repositories\Metric\MySql as MetricMySqlRepository;
```
GrahamCampbell: I don't like these renames. ;(

jbrooksuk: What would you suggest?

GrahamCampbell: Keeping it how it was. :P

jbrooksuk: Hah, okay then. I'll rename them once it's finished.

app/Repositories/Metric/MySql.php
```diff
@@ -0,0 +1,80 @@
...
+     *
+     * @param \CachetHQ\Cachet\Models\Metric $metric
+     * @param int                            $minutes
+     *
+     * @return \Illuminate\Support\Collection
+     */
+    public function getPointsSinceMinutes(Metric $metric, $minutes)
+    {
+        $queryType = $this->getQueryType($metric);
+        $points = Collection::make(DB::select("SELECT DATE_FORMAT(metric_points.`created_at`, '%H:%i') AS `key`, {$queryType} FROM {$this->getTableName()} INNER JOIN metric_points ON metrics.id = metric_points.metric_id WHERE metrics.id = :metricId AND metric_points.`created_at` >= DATE_SUB(NOW(), INTERVAL :minutes MINUTE) GROUP BY HOUR(metric_points.`created_at`), MINUTE(metric_points.`created_at`) ORDER BY metric_points.`created_at`", [
```
ConnorVG: These are absolutely disgusting... :scream:

GrahamCampbell: :trollface:

jbrooksuk: Laravel 5.3 stops us having to wrap `DB::select` in a collection at least.

jbrooksuk:
The issue we have with all of the metrics is that each DB driver queries differently and Eloquent isn't flexible enough to allow us to keep the same query structure — unless I'm missing something?

I really want to get this into 2.4, so I'll try and finish this at a later date.

jbrooksuk:
Rebased.

jbrooksuk:
This is finally done!

ConnorVG:
😱 👍

PR #2050, opened 2016-08-10
[WIP] Send errors to our beacon

GrahamCampbell:


config/bugsnag.php
```diff
@@ -0,0 +1,115 @@
...
+    | Enable this if you'd like us to set the current user logged in via
+    | Laravel's authentication system.
+    |
+    | If you'd like to add your own user resolver, you can call the
+    | Bugsnag::registerUserResolver method from the boot method of your app
+    | service provider.
+    |
+    */
+
+    'user' => true,
```
jbrooksuk: I don't think we should do this. Since users from several status pages would be grouped into one on Bugsnag.

GrahamCampbell: Not if we transform it on our end.

GrahamCampbell: That said, maybe user information is private, and we shouldn't send it to us?

jbrooksuk: I don't think we should be sending it to us. Beacon is anonymous on a whole, with the exception being the support contact email.

PR #2084, opened 2016-08-26
Use updated status in component status update event

### LINKED ISSUE #2061: Wrong component status in status update email
Before submitting your issue, please make sure that you've checked the checkboxes below.

- [x ] I am running the [2.4](https://github.com/CachetHQ/Cachet/releases/latest) version of Cachet.
- [x ] I am running at least PHP 5.5.9. *You can check this by running `php -v`.*
- [x ] I have ran `rm -rf bootstrap/cache/*`.

### Expected behaviour

* When a component gets updated to major outage, subscribers receive an email with content "The component XXX has seen a status change. The component is now at Major Outage."

### Actual behaviour

* When a component gets updated to major outage, subscribers receive an email with content "The component XXX has seen a status change. The component is now at Operational." -> Operational is the old status.

### Steps to reproduce

* Create a component, add a subscriber, update the status, check the received email.

hyperr:
The status updated event used component (untouched) in the event. After adding the event the component was updated.
This fixes sending wrong (old) component status within the status updated notification email and fixes fixes #2061


ConnorVG:
I'm unsure on if we should do this or if we should change the logic in the `ComponentStatusWasUpdatedHandler` instead, as I previously mentioned.

fdhex:
@ConnorVG : As it's a bug in the latest release, I'd suggest releasing the fix, and then think about the refactoring in further releases.

ConnorVG:
Implementing inconsistent code wont get us anywhere. I believe we should just get a higher quality fix done instead (or you can do it yourself, if you like) as it's not much extra work at all.

_It'll be just like comments like `// temporary` `// @todo: implement correctly` sort of stuff._

GrahamCampbell:
No need to hack a fix out. We're in no rush to release 2.4. We should implement the correct fix long term.

fdhex:
I understand your point of view, and will be using the "refactored" fix when it comes out.

Meanwhile, I have already used the above fix in my version, so all is good.

jbrooksuk:
We'll do this for v2.4

PR #2089, opened 2016-08-30
Split css between dashboard and status page

joecohens:
/cc @jbrooksuk


joecohens:
After this I will make the same with all js

Issue #109, opened 2014-12-28
Update system


jbrooksuk:
We need some way of updating Cachet from an existing installation. This will keep Heroku users up to date and make it easier for manual installations to stay up to date too.
- [ ] Option to turn off update checking.
- [ ] Only check for updates whilst we've not got an active version to update too already.
- [x] Visual notification that an update is available.
- [x] Use GitHubs releases API to find the latest version.
- [ ] Download the update to `./app/storage/updates`
- [ ] Copy over the new files, which have changed.
- [ ] Run `composer update`
- [ ] Option to allow `dev-master` style updates that will use the `master` branch, with a **big** warning.

jbrooksuk:
I think we should do this before V1, but just before we release. The reason is that we can get get a solid release setup on the go, when we can make use of GitHub releases. The update system can then look for that using the API.

jbrooksuk:
GitHub has just added [two new API methods](http://www.programmableweb.com/news/github-adds-two-new-methods-to-releases-api/elsewhere-web/2015/02/22), which is perfect for this update system.

joecohens:
@jbrooksuk Just in time!

jbrooksuk:
@joecohens I think we'll need to implement this with some test releases on a test project I'd say.

jbrooksuk:
@joecohens I'm tempted to move this feature to V2 as well. At the moment Cachet is quite easy to get updated, plus the architecture changes that we've spoken about for V2 will make this much easier for us. I'll Slack you more of my thoughts.

ConnorVG:
Is someone working on this or does no one mind if I take a crack at it this weekend?

jbrooksuk:
Please be my guest!

Most installations should be done via Git, keep that in mind :)

ConnorVG:
I have a few diff based update ideas for this. I'll mull it over.

GrahamCampbell:
Don't take too much crack. :trollface:

jbrooksuk:
Thinking about it, upgrading Cachet should be:

```
$ cd $cachet_dir
$ git pull
$ git checkout LATEST_TAG
$ composer install --no-dev
$ php artisan app:update
```

PR #2093, opened 2016-09-04
Implement an upgrade system

jbrooksuk:


app/Console/Commands/UpgradeCommand.php
```diff
@@ -0,0 +1,88 @@
...
+
+        foreach ($paths as $path) {
+            $this->comment('You can upgrade to '.$path->getVersion().'!');
+
+            if ($this->confirm("Do you wish to continue?\n Not doing so may leave your status page in an unusable state.")) {
+                $this->info('Performing upgrade...');
+
+                if ($path->upgrade()) {
+                    $this->info('Upgrade successful!');
+                } else {
```
ConnorVG: Could we not just guard instead of using `else` everywhere? It's a bit... gross... :mask:

app/Upgrades/Paths/V240Upgrade.php
```diff
@@ -0,0 +1,41 @@
...
+    protected $version = 'v2.4.0';
+
+    /**
+     * Run the upgrade path.
+     *
+     * @return bool
+     */
+    public function upgrade()
+    {
+        echo 'This would upgrade us to '.$this->version."\n";
```
ConnorVG: What sort of processes would occur here? It seems like the way it's being done it's going to end up having to be a 'next only' upgrade system? I'm confused.

app/Console/Commands/UpgradeCommand.php
jbrooksuk: Hah, yeah.

ConnorVG:
#109

jbrooksuk:
We're not doing this now :)

PR #2110, opened 2016-09-22
- fixed metrics if you use a table prefix

karaktaka:


.gitignore
```diff
@@ -2,3 +2,4 @@
 node_modules
 phpunit.xml
 vendor
+/.idea/
```
GrahamCampbell: Please do not add this.

app/Repositories/Metric/AbstractMetricRepository.php
```diff
@@ -58,6 +58,11 @@ protected function getTableName()
         $connection = $this->config->get('database.connections.'.$driver);
         $prefix = $connection['prefix'];
 
-        return $prefix.'metrics';
+        $return = (object) array(
```
GrahamCampbell: Please update the phpdoc. Also, an array would be fine here.

GrahamCampbell: Infact, it's better to just make two methods here than returning something like this.

.gitignore
jbrooksuk: We should make a CS check that these don't get changed 😂

app/Repositories/Metric/AbstractMetricRepository.php
jbrooksuk: Also, we use the new array syntax only.

karaktaka: Feel free to modify it. It was just a suggestion.

app/Repositories/Metric/AbstractMetricRepository.php
```diff
@@ -58,6 +58,11 @@ protected function getTableName()
...
+                        'metrics_base' => $prefix.'metrics',
```
ConnorVG: I'm not a fan of concatenation. Especially when we have possible PHP7 optimisations we can have here.

ConnorVG:
Doesn't it make more sense to just do something like this? https://gist.github.com/ConnorVG/51c863929ce0f3c6800c0a4aa16b13f8

PR #2122, opened 2016-10-01
HTML Fixes

abhimanyu003:
Pull requests includes.

* Removed unwanted closing `form` tag
* Fixed closing of span tag
* Minor HTML CS fix ( Basically single quote to double quote )

**Updates**

* Lower case email before generating md5 as per #2123


jbrooksuk:
Thanks!

PR #2138, opened 2016-10-05
Api groups order

billmn:
Issue: if the request use “?sort=order” returns groups ordered by “order DESC”.

This is unexpected.
If I would a reversed order I have to use “sort=order&order=desc”

### DIFF
--- a/app/Http/Controllers/Api/ComponentGroupController.php
+++ b/app/Http/Controllers/Api/ComponentGroupController.php
@@ -62,7 +62,7 @@ class ComponentGroupController extends AbstractApiController
         $groups->search(Binput::except(['sort', 'order', 'per_page']));
 
         if ($sortBy = Binput::get('sort')) {
-            $direction = Binput::has('order') && Binput::get('order') == 'desc';
+            $direction = Binput::get('order', 'asc');
 
             $groups->sort($sortBy, $direction);
         }


jbrooksuk:
One thing we should add is multiple sorts. Fancy this @billmn?

billmn:
@jbrooksuk Multiple sorts can introduce some BC but I think we can remove the "order" parameter using only "sort":

<code>?sort=-status,order</code> == <code>ORDER BY status DESC, ORDER ASC</code>

jbrooksuk:
I like it, but this breaks BC and I don't like that.

ConnorVG:
I agree with what @jbrooksuk said but I also have a question... What would be the 'default' definition for the sort orders if you're wanting it to go this way? Seems both opinionated _and_ random to me.

billmn:
@jbrooksuk I agree too, nobody loves BC :)

@ConnorVG If I don't specify the "sort" parameter I use a default sorting (eg: for component groups -> status DESC, order ASC).

If I specify the "sort" parameter:
- with the hyphen the order is descending
- without the hyphen the order is ascending

ConnorVG:
I see, I didn't notice the hyphen. Thanks.

jbrooksuk:
I guess we have to break BC on something like this?

ConnorVG:
We don't _have_ to, especially if we use a versioned API.

jbrooksuk:
True. Although that's a ball ache to maintain.

ConnorVG:
:thinking: :thought_balloon:

PR #2142, opened 2016-10-06
Incident updates

jbrooksuk:
Backporting this to v2.4.0 because it's so awesome and Cachet lacks a seriously needed feature.


ConnorVG:
What a legend...

PR #2160, opened 2016-10-08
Minor HTML tweaks

abhimanyu003:
Just few minor HTML tweaks.

Question: I'm thinking to add `placeholder` in input area, if you are OK with that I will do that in next pull request ? Yes I will take care of translation.


jbrooksuk:
@abhimanyu003 that's perfect! Could you please only add the English translations, and then I sync it with CrowdIn. Thank you :)

PR #2171, opened 2016-10-12
Updates exception in signup to ModelNotFoundException

### LINKED ISSUE #2170: Use ModelNotFoundException instead
Why we are not using `ModelNotFoundException` instead of `BadRequestHttpException`

https://github.com/CachetHQ/Cachet/blob/2.4/app/Http/Controllers/SignupController.php#L41-L45

I guess we can refactor this to and can use `firstOrFail`

```
Invite::where('code', '=', $code)->whereNotNull('claimed_at')->firstOrFail();
```

Or Even better

```
Invite::whereCode($code)->whereNotNull('claimed_at')->firstOrFail();
```

I can see this code pattern is used over many places if you like I can create a pull request to update them this way.

Thanks

bretterer:
This PR is to address and resolve #2170 

No tests were currently created for this, so no tests were updated.

### DIFF
--- a/app/Http/Controllers/SignupController.php
+++ b/app/Http/Controllers/SignupController.php
@@ -17,6 +17,7 @@ use CachetHQ\Cachet\Bus\Commands\User\SignupUserCommand;
 use CachetHQ\Cachet\Models\Invite;
 use CachetHQ\Cachet\Models\User;
 use GrahamCampbell\Binput\Facades\Binput;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
 use Illuminate\Routing\Controller;
 use Illuminate\Support\Facades\Redirect;
 use Illuminate\Support\Facades\View;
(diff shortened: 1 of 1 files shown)


jbrooksuk:
Ping @GrahamCampbell.

bretterer:
I will fix StyleCI in a few

GrahamCampbell:
This doesn't look correct. This will give an error 500.

GrahamCampbell:
Error 400 is fine.

bretterer:
@GrahamCampbell Where would this present a 500 error?   model not found is an exception.  The PR does not change any code other than the exception that is thrown if the model is not found.

GrahamCampbell:
A model not found exception yeilds an error 500, and if it doesn't right now, it will in a later version of Cachet.

GrahamCampbell:
Just checked the code. It does already give an error 500. This is because a model not found exception doesn't necessarily mean the it happened in routing. It can happen anywhere in the app code.

PR #2174, opened 2016-10-12
Completely refactor all routes

jbrooksuk:
I've put a new routing structure in place!

We're now prefixing all Cachet core routes with `core::`, followed by the request type, `get:` and then a name. An example of linking to the dashboard incidents update form `core::get:dashboard.incidents.updates`.

This introduces a new `cachet_route` helper which is similar to `route`, but prefixes the `core` domain.

Aside from just organising them, I've actually changed some of the URLs too. So now `/add` = `/create` and there is no `/edit` route for entities now. Clean! 🚀


ConnorVG:
:gem:

PR #2176, opened 2016-10-13
Fixed redirects with “cachet_redirect” helper

billmn:
This commit https://github.com/CachetHQ/Cachet/commit/5a9f6134c48383d779535b572f6ee2076932c3ad breaks all redirects.

Created "cachet_redirect" helper to restore redirects and allow to specify a domain.


jbrooksuk:
Thanks!

Issue #2184, opened 2016-10-15
Filter for displaying component group data


nicolasfagotti:
When many components are added, is difficult to focus on specific data. It would be nice to have a way to filter the component group information without disrupting whole site functionality.

jbrooksuk:
Thank you for your input on Cachet 2.x. We are shifting our attention and resources to Cachet 3.x and will no longer be supporting the 2.x version. If your feedback or issue is relevant to the 3.x series, we encourage you to engage with the new branch.

For more information on the Cachet rebuild and our plans for 3.x, you can read the announcement [here](https://github.com/CachetHQ/Cachet/discussions/4342).

We appreciate your understanding and look forward to your contributions to the new version.

PR #2185, opened 2016-10-15
Component filter

nicolasfagotti:
Changes on this commit related to issue #2184:
- The possibility of relating a metric to a specific component.
- CRUD actions updated for the new metric field, allowing to select a component or not when a metrics is created/udpated.
- Dropdown element added on the status page for allowing selecting a component group to filter.
- Separated pages for visualizing the group selected, with their individual metric charts and incidents.
- Component label added on metric charts.


nicolasfagotti:
We are using these changes on one internal tool of our company. We were wondering if they are suitable to be considered as potential permanent changes for the CachetHQ project.

It's my first time with Laravel, so any advice is highly welcome.

Also, I need to check the tests. I didn't realize about them.

jbrooksuk:
Please can you remove the compiled files, I'll compile these separately :)

database/migrations/2016_10_06_094000_AlterTableMetricsAddComponentRelation.php
```diff
@@ -0,0 +1,37 @@
...
+
+class AlterTableMetricsAddComponentRelation extends Migration
+{
+    /**
+     * Run the migrations.
+     */
+    public function up()
+    {
+        Schema::table('metrics', function (Blueprint $table) {
+            $table->integer('component_id')->default(0);
```
jbrooksuk: Can you add this after the `order` column please.

resources/views/dashboard/metrics/add.blade.php
```diff
@@ -31,6 +31,25 @@
                             <textarea name="metric[description]" class="form-control" rows="5" placeholder="{{ trans('forms.metrics.description') }}">{{ Binput::old('metric.description') }}</textarea>
                         </div>
                     </div>
+                    @if(!$components_in_groups->isEmpty() || !$components_out_groups->isEmpty())
+                        <div class="form-group">
```
jbrooksuk: You don't need to indent beneath any Blade tags.

resources/views/partials/components.blade.php
```diff
@@ -29,3 +47,24 @@
     @endforeach
 </ul>
 @endif
+
+<script>
+(function () {
```
jbrooksuk: Really, I don't think we need any JS for the dropdown. You can use Blade to find which group is selected and pass that through to the view. This would make the groups unusable for somebody without JS, which isn't needed.

nicolasfagotti:
I pushed the changes you proposed @jbrooksuk. Thanks for your feedback! :)

Let me know if you find anything else to fix.

jbrooksuk:
Thanks @nicolasfagotti! I'll take a look at this later on :)

jbrooksuk:
Can you rebase this please?

nicolasfagotti:
Sorry for the delay. We've been working on a way to filter components too.

I know the rebase was a mess, and the number of commits are too high. It was hard after so many changes on 2.4 branch. I can try to do it cleaner if there is an intention of considering these changes to be integrated.

ConnorVG:
Don't worry about the amount of commits, just worry about the quality of them :smile:
Worst case scenario? We squash them :grimacing:

jbrooksuk:
Wow 😆 it's okay, we'll figure it out.

nicolasfagotti:
Please dismiss this pull request.

There is a new pull request (much cleaner -one commit only-) that includes this feature + the historical status transition support ---> #2243.

The component/group filter makes much more sense with the status transition charts :)

PR #2192, opened 2016-10-17
Template overview broken when templates exists

coupej:
Due to a typo in the routing table the dashboard/templates page breaks once there are
templates available. Once the typo is resolved everything works as
expected


jbrooksuk:
Awesome, thank you!

PR #2194, opened 2016-10-18
No data available status

nicolasfagotti:
We added a new status for the components, "No Data Available". This doesn't seem meaningful for manual modifications, but when components are regularly updated through Cachet API, it's pretty useful to have it.

If it's not desirable, my apologies and just cancel the pull request :)

### DIFF
--- a/app/Bus/Commands/Component/AddComponentCommand.php
+++ b/app/Bus/Commands/Component/AddComponentCommand.php
@@ -70,7 +70,7 @@ final class AddComponentCommand
     public $rules = [
         'name'        => 'required|string',
         'description' => 'required|string',
-        'status'      => 'required|int|min:1|max:4',
+        'status'      => 'required|int|min:0|max:4',
         'link'        => 'nullable|url',
         'order'       => 'nullable|int',
         'group_id'    => 'nullable|int',
--- a/app/Bus/Commands/Component/UpdateComponentCommand.php
+++ b/app/Bus/Commands/Component/UpdateComponentCommand.php
@@ -79,7 +79,7 @@ final class UpdateComponentCommand
     public $rules = [
         'name'        => 'nullable|string',
         'description' => 'nullable|string',
-        'status'      => 'nullable|int|min:1|max:4',
+        'status'      => 'nullable|int|min:0|max:4',
         'link'        => 'nullable|url',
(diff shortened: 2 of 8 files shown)


jbrooksuk:
Hmm, I like the idea of having an almost indeterminate status, but I don't know if I like the "No data available" status.

Actually, under what circumstances would you have no data?

nixmomo:
Some monitorings use the "unkown" status if they can't get informations about the service health from the monitoring client. Nagios as example use it :-)

If you use cachet as a internal status Page it can be helpful to have it to see that something is going wrong. if you match it hard to "down" and the service isn't really down so you have a status between down and up.

nicolasfagotti:
Well, for example, we query our component statuses not directly hitting the components but querying a central log application.

```Cachet -> log app <- components```

So, in our case, having no information in the logs doesn't mean anything about a component status. What we are looking for is a way to say that is not possible to define the status of the component.

jbrooksuk:
Okay, thanks for providing a better explanation.

I think I prefer "Unknown". Thoughts? Ping @CachetHQ/core.

nicolasfagotti:
Yes, it sounds better "Unknown".

jbrooksuk:
Okay and two more points here:

1. Can you _only_ update the English translation, please? The reason is that we sync the English files up to CrowdIn, then they generate the files when needed. I'd have to go through every language and re-sync them. This makes it much easier for me to deal with :+1:
2. You'll need to update the `./app/Bus/Commands/Component` command files to validate a status of `0` too, as it's currently only allowing 1-4.

nicolasfagotti:
If I'm not mistaken, the test that failed is not related to the changes on this branch.

jbrooksuk:
Can you squash into one commit, please? We have a commit adding languages, then removing them. I'll check that this does everything we expect on Sunday :)

nicolasfagotti:
Done.

Sorry about that. It's my first time with Cachet, Laravel, Github as issue tracker, etc. I hope my next pull request be cleaner :)

jbrooksuk:
Can you rebase this please? We changed some stuff in the Laravel upgrade. Once debased we will merge this immediately.

nicolasfagotti:
Done

PR #2212, opened 2016-10-27
Add incident column for when an incident occurred at

### LINKED ISSUE #2208: update_at value with dashboard
When I use the dashboard, created_at and updated_at are always equals.

My the typical workflow is:
Create an incident with the dashboard, and query the database: 

mysql> select * from incidents where id=2\G
*************************** 1. row ***************************
          id: 2
component_id: 0
        name: test_update_at
      status: 1
     visible: 1
     message: Test update_at
scheduled_at: NULL
  **created_at: 2016-10-24 14:09:15**
  **updated_at: 2016-10-24 14:09:15**
  deleted_at: NULL
1 row in set (0.00 sec)

Then I update the incident with the API (here I change the status): 
curl -X PUT -H "Content-Type: application/json" -H "X-Cachet-Token: xxxx" http://status.com/api/v1/incidents/2  -d '{"status":2}'

Now database have different values for created_at and updated_at: 
mysql> select * from incidents where id=2\G
*************************** 1. row ***************************
          id: 2
component_id: 0
        name: test_update_at
      status: 2
     visible: 1
     message: Test update_at
scheduled_at: NULL
  **created_at: 2016-10-24 14:09:15**
  **updated_at: 2016-10-24 14:14:08**
  deleted_at: NULL
1 row in set (0.00 sec)

Now if I edit the incident again with the dashboard an query the database: 

mysql> select * from incidents where id=2\G
*************************** 1. row ***************************
          id: 2
component_id: 0
        name: test_update_at
      status: 2
     visible: 1
     message: Test update_at
scheduled_at: NULL
  **created_at: 2016-10-24 14:09:00**
  **updated_at: 2016-10-24 14:09:00**
  deleted_at: NULL
1 row in set (0.01 sec)

It seems that the dashboard use the same value for created_at and updated_at.

jbrooksuk:
Closes #2208 

---

This does change the API requests, as we're now looking for a field called `occurred_at` instead of `incident_date`, or whatever.

### DIFF
--- a/app/Bus/Commands/Incident/ReportIncidentCommand.php
+++ b/app/Bus/Commands/Incident/ReportIncidentCommand.php
@@ -80,7 +80,7 @@ final class ReportIncidentCommand
      *
      * @var string|null
      */
-    public $occurredAt;
+    public $occurred_at;
 
     /**
      * A given incident template.
@@ -110,7 +110,7 @@ final class ReportIncidentCommand
         'component_status' => 'nullable|required_with:component_id|int|min:0|max:4',
         'notify'           => 'nullable|bool',
         'stickied'         => 'required|bool',
-        'occurredAt'       => 'nullable|string',
+        'occurred_at'      => 'nullable|string',
         'template'         => 'nullable|string',
     ];
 
@@ -125,13 +125,13 @@ final class ReportIncidentCommand
      * @param int         $component_status
      * @param bool        $notify
      * @param bool        $stickied
-     * @param string|null $occurredAt
+     * @param string|null $occurred_at
      * @param string|null $template
      * @param array       $template_vars
      *
      * @return void
      */
-    public function __construct($name, $status, $message, $visible, $component_id, $component_status, $notify, $stickied, $occurredAt, $template, array $template_vars = [])
+    public function __construct($name, $status, $message, $visible, $component_id, $component_status, $notify, $stickied, $occurred_at, $template, array $template_vars = [])
     {
         $this->name = $name;
         $this->status = $status;
@@ -141,7 +141,7 @@ final class ReportIncidentCommand
         $this->component_status = $component_status;
         $this->notify = $notify;
         $this->stickied = $stickied;
-        $this->occurredAt = $occurredAt;
+        $this->occurred_at = $occurred_at;
         $this->template = $template;
         $this->template_vars = $template_vars;
     }
--- a/app/Bus/Commands/Incident/ReportIncidentCommand.php
+++ b/app/Bus/Commands/Incident/ReportIncidentCommand.php
@@ -76,11 +76,11 @@ final class ReportIncidentCommand
     public $stickied;
 
     /**
-     * The date at which the incident occurred.
+     * The date at which the incident occurred at.
      *
      * @var string|null
      */
-    public $incident_date;
+    public $occurredAt;
 
     /**
      * A given incident template.
@@ -110,7 +110,7 @@ final class ReportIncidentCommand
         'component_status' => 'nullable|required_with:component_id|int|min:0|max:4',
         'notify'           => 'nullable|bool',
         'stickied'         => 'required|bool',
-        'incident_date'    => 'nullable|string',
+        'occurredAt'       => 'nullable|string',
         'template'         => 'nullable|string',
     ];
 
@@ -125,13 +125,13 @@ final class ReportIncidentCommand
      * @param int         $component_status
      * @param bool        $notify
      * @param bool        $stickied
-     * @param string|null $incident_date
+     * @param string|null $occurredAt
      * @param string|null $template
      * @param array       $template_vars
      *
      * @return void
      */
-    public function __construct($name, $status, $message, $visible, $component_id, $component_status, $notify, $stickied, $incident_date, $template, array $template_vars = [])
+    public function __construct($name, $status, $message, $visible, $component_id, $component_status, $notify, $stickied, $occurredAt, $template, array $template_vars = [])
     {
         $this->name = $name;
         $this->status = $status;
@@ -141,7 +141,7 @@ final class ReportIncidentCommand
         $this->component_status = $component_status;
         $this->notify = $notify;
         $this->stickied = $stickied;
-        $this->incident_date = $incident_date;
+        $this->occurredAt = $occurredAt;
         $this->template = $template;
         $this->template_vars = $template_vars;
     }
--- a/app/Bus/Handlers/Commands/Incident/UpdateIncidentCommandHandler.php
+++ b/app/Bus/Handlers/Commands/Incident/UpdateIncidentCommandHandler.php
@@ -19,7 +19,6 @@ use CachetHQ\Cachet\Dates\DateFactory;
 use CachetHQ\Cachet\Models\Component;
 use CachetHQ\Cachet\Models\Incident;
 use CachetHQ\Cachet\Models\IncidentTemplate;
-use Carbon\Carbon;
 use Twig_Environment;
 use Twig_Loader_Array;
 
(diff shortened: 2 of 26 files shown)


app/Bus/Commands/Incident/ReportIncidentCommand.php
```diff
@@ -110,7 +110,7 @@
         'component_status' => 'nullable|required_with:component_id|int|min:0|max:4',
         'notify'           => 'nullable|bool',
         'stickied'         => 'required|bool',
-        'incident_date'    => 'nullable|string',
+        'occurredAt'       => 'nullable|string',
```
ConnorVG: I reckon we should move to `camelCase` but I also reckon it should be done in it's own PR rather than just leaving random bits here and there.

app/Bus/Handlers/Commands/Incident/UpdateIncidentCommandHandler.php
```diff
@@ -61,18 +62,20 @@ public function handle(UpdateIncidentCommand $command)
...
-            $incidentDate = $this->dates->create('d/m/Y H:i', $command->incident_date);
-
-            $incident->update([
-                'created_at' => $incidentDate,
-                'updated_at' => $incidentDate,
-            ]);
+        if ($occurredAt = $command->occurredAt) {
+            if ($date = $this->dates->create('Y-m-d H:i', $occurredAt)) {
+                $incident->fill(['occurred_at' => $date]);
+            } else {
```
ConnorVG: Guard > `else` - looks messy imo

app/Http/Controllers/Dashboard/DashboardController.php
```diff
@@ -127,11 +127,11 @@ public function showDashboard()
...
     protected function getIncidents()
     {
-        $allIncidents = Incident::notScheduled()->whereBetween('created_at', [
+        $allIncidents = Incident::notScheduled()->whereBetween('occurred_at', [
             $this->startDate->copy()->subDays(30)->format('Y-m-d').' 00:00:00',
             $this->startDate->format('Y-m-d').' 23:59:59',
-        ])->orderBy('created_at', 'desc')->get()->groupBy(function (Incident $incident) {
-            return (new Date($incident->created_at))
+        ])->orderBy('occurred_at', 'desc')->get()->groupBy(function (Incident $incident) {
+            return (new Date($incident->occurred_at))
```
ConnorVG: Could we not use the presenter here?

app/Http/Controllers/FeedController.php
```diff
@@ -80,12 +85,12 @@ private function feedAction(ComponentGroup &$group, $isRss)
     {
         if ($group->exists) {
             $group->components->map(function ($component) use ($isRss) {
-                $component->incidents()->visible()->orderBy('created_at', 'desc')->get()->map(function ($incident) use ($isRss) {
+                $component->incidents()->visible()->orderBy('occurred_at', 'desc')->get()->map(function ($incident) use ($isRss) {
```
ConnorVG: Long line is long :stuck_out_tongue:

app/Http/Controllers/StatusPageController.php
```diff
@@ -109,7 +109,7 @@ public function showIndex()
             ->withDaysToShow($daysToShow)
             ->withAllIncidents($allIncidents)
             ->withCanPageForward((bool) $today->gt($startDate))
-            ->withCanPageBackward(Incident::notScheduled()->where('created_at', '<', $startDate->format('Y-m-d'))->count() > 0)
+            ->withCanPageBackward(Incident::notScheduled()->where('occurred_at', '<', $startDate->format('Y-m-d'))->count() > 0)
```
ConnorVG: Do we even need to format `$startDate`?

app/Presenters/IncidentPresenter.php
```diff
@@ -56,33 +56,73 @@ public function raw_message()
     }
 
     /**
+     * Present formatted occurred_at date time.
+     *
+     * @return string
+     */
+    public function occurred_at()
+    {
+        return app(DateFactory::class)->make($this->wrappedObject->occurred_at)->toDateTimeString();
```
ConnorVG: Off topic but is there any reason we don't just inject the DateFactory?

app/Presenters/IncidentPresenter.php
```diff
@@ -56,33 +56,73 @@ public function raw_message()
...
     /**
      * Present formatted date time.
      *
      * @return string
      */
-    public function created_at_formatted()
+    public function occurred_at_formatted()
     {
-        return ucfirst(app(DateFactory::class)->make($this->wrappedObject->created_at)->format(Config::get('setting.incident_date_format', 'l jS F Y H:i:s')));
+        return ucfirst(app(DateFactory::class)->make($this->wrappedObject->occurred_at)->format(Config::get('setting.incident_date_format', 'l jS F Y H:i:s')));
```
ConnorVG: This seems overly complex / long. QoL goes down.

app/Presenters/IncidentPresenter.php
```diff
@@ -56,33 +56,73 @@ public function raw_message()
...
+    }
+
+    /**
+     * Present formatted date time.
+     *
+     * @return string
+     */
+    public function created_at_formatted()
+    {
+        return ucfirst(app(DateFactory::class)->make($this->wrappedObject->created_at)->format(Config::get('setting.incident_date_format', 'l jS F Y H:i:s')));
```
ConnorVG: Same as previous.

app/helpers.php
```diff
@@ -177,3 +177,61 @@ function cachet_redirect($name, $parameters = [], $status = 302, $headers = [],
...
+     *
+     * Thanks to http://stackoverflow.com/a/30192680/394013
+     *
+     * @param string $format
+     *
+     * @return string
+     */
+    function datetime_to_moment($format)
+    {
+        $replacements = [
```
ConnorVG: This is just bizarre. I see it's point tho :laughing:

resources/views/single-incident.blade.php
```diff
@@ -7,7 +7,7 @@
 @stop
 
 @section('content')
-<h1>{{ $incident->name }} <small>{{ formatted_date($incident->created_at) }}</small></h1>
+<h1>{{ $incident->name }} <small>{{ formatted_date($incident->occurred_at) }}</small></h1>
```
ConnorVG: Why are we not using the presenter here?

app/Bus/Handlers/Commands/Incident/UpdateIncidentCommandHandler.php
jbrooksuk: Also, the exception doesn't do anything here, because `create` will throw, so we need to catch.

app/Http/Controllers/Dashboard/DashboardController.php
jbrooksuk: I was thinking this when I was looking at it this morning.

app/Http/Controllers/FeedController.php
jbrooksuk: I should add a better scope for this.

app/Http/Controllers/StatusPageController.php
jbrooksuk: Dunno.

app/Presenters/IncidentPresenter.php
jbrooksuk: Laziness.

app/Presenters/IncidentPresenter.php
jbrooksuk: :+1:

app/helpers.php
jbrooksuk: I didn't end up using this, but I like it, so I'm keeping it because I probably will use this at some point.

resources/views/single-incident.blade.php
jbrooksuk: I noticed this too, dunno why.

app/Http/Controllers/StatusPageController.php
jbrooksuk: Yes, because `occurred_at` might contain a time, but we're only interested in whole days.

jbrooksuk:
Let's merge and see what happens 👍 there might be an edge-case I've missed.

PR #2215, opened 2016-10-27
Change "order" condition when a component is edited

nicolasfagotti:
It's just me, or when the component order is false, the condition returns an empty value breaking the site after the update is confirmed? I had to change the way the condition is formatted for fixing the problem. Now the 0 is placed if the component order checking is false.

If this is happening on my case only, please ignore the pull request.

### DIFF
--- a/resources/views/dashboard/components/edit.blade.php
+++ b/resources/views/dashboard/components/edit.blade.php
@@ -66,7 +66,7 @@
                 </fieldset>
 
                 <input type="hidden" name="component[user_id]" value="{{ $component->agent_id || $current_user->id }}">
-                <input type="hidden" name="component[order]" value="{{ $component->order || 0 }}">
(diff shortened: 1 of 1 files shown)


resources/views/dashboard/components/edit.blade.php
```diff
@@ -66,7 +66,7 @@
                 </fieldset>
 
                 <input type="hidden" name="component[user_id]" value="{{ $component->agent_id || $current_user->id }}">
-                <input type="hidden" name="component[order]" value="{{ $component->order || 0 }}">
+                <input type="hidden" name="component[order]" value="{{ $component->order ? $component->order : 0 }}">
```
GrahamCampbell: should be `?:` please

nicolasfagotti:
Change applied :)

jbrooksuk:
Thanks!

PR #2219, opened 2016-10-30
Metric repository perf - 1.2

ELDHOSE104:
1.2 version is still used on legacy systems and needs to be maintained.


ConnorVG:
I'm sorry, there is no way we can maintain such an old version Cachet with such recent features.

jbrooksuk:
The metric repository perf branch is also incomplete and should not be used.

Issue #2222, opened 2016-10-31
Remove someRandomString from .env.example


seanmangar:
I was setting up this project in my local computer and noticed that if you don't remove someRandomString from .env.example, running php artisan key:generate appends the key before _=SomeRandomString_ .This is perhaps why in Laravel's .env.example it is kept blank: https://github.com/laravel/laravel/blob/master/.env.example

```
APP_ENV=production
APP_DEBUG=false
APP_URL=http://localhost
APP_KEY=base64:At8uiauIpFa9Il+e+sxAvXOnZUylqkUGN1hrfQyAJco=SomeRandomString

DB_DRIVER=mysql
DB_HOST=localhost
DB_DATABASE=cachet
DB_USERNAME=homestead
DB_PASSWORD=secret
DB_PORT=null
DB_PREFIX=null

CACHE_DRIVER=file
SESSION_DRIVER=file
QUEUE_DRIVER=sync

CACHET_BEACON=true
CACHET_EMOJI=false

MAIL_DRIVER=smtp
MAIL_HOST=mailtrap.io
MAIL_PORT=2525
MAIL_USERNAME=null
MAIL_PASSWORD=null
MAIL_ADDRESS=null
MAIL_NAME=null
MAIL_ENCRYPTION=tls

REDIS_HOST=null
REDIS_DATABASE=null
REDIS_PORT=null

GITHUB_TOKEN=null

```

PR #2223, opened 2016-10-31
Remove someRandomString from .env.example

seanmangar:
Remove someRandomString from .env.example

Removed someRandomString because it interferes with php artisan
key:generate

Resolves: #2222


jbrooksuk:
Thanks!

PR #2231, opened 2016-11-02
More spanish grammar corrections

jmarcher:


jbrooksuk:
Thanks, but this is the same as the other ones.

PR #2251, opened 2016-11-25
Don't cache settings

jbrooksuk:
I'm not happy with the friction that the settings caching is creating between users and Cachet. If we need to add caching, then we need to make it work easily, without breaking anything for existing installations.

---

Ping @joecohens @ConnorVG @GrahamCampbell


GrahamCampbell:
I think that caching is fine. We should just make sure it fails loudly for bad permissions.

jbrooksuk:
> We should just make sure it fails loudly for bad permissions.

👍  We should display a big alert in the dashboard then.

jbrooksuk:
Done in https://github.com/CachetHQ/Cachet/commit/c9c592b03328ab02376956ea8d9fa7dcca63b8bc

PR #2263, opened 2016-12-05
Components can now be supplied meta data

### LINKED ISSUE #2226: Allow for custom data fields in components
Hi!

I'm working on a tool that automatically checks certain things for components I added to Cachet (Right now, I'm only pinging and getting response codes). 

I currently have a file that simply stores the settings for each component in an array. In this I store the url, check type, expected response code, timeout, and some other details needed for tests. I tried moving this to tags, but that didn't work because you can't have a custom tag slug (as far as I know). 

I would like to be able to add custom data to a component. I imagine this being similar to tags, but instead of the tag slug being automatically generated, it is manually entered by me. That way I can store some data used by my automatic checking tool.

In the API response for GET /components/:id, I could imagine a response like

        {
            "id": 1,
            "name": "API",
            "description": "This is the Cachet API.",
            "link": "",
            "status": 1,
            "order": 0,
            "group_id": 0,
            "created_at": "2015-07-24 14:42:10",
            "updated_at": "2015-07-24 14:42:10",
            "deleted_at": null,
            "status_name": "Operational",
            "tags": [
                "slug-of-tag": "Tag Name"
            ],
            "meta": [
                "check-type": "http",
                "timeout": "15",
                "port": "80"
            ]
        }

jbrooksuk:
Closes #2226

---

I'd still like to get incidents being able to supply meta data, but for now I imagine it's more useful for components. And, if somebody wants to add this to incidents, this is how 🙂

### DIFF
--- a/app/Bus/Commands/Component/AddComponentCommand.php
+++ b/app/Bus/Commands/Component/AddComponentCommand.php
@@ -63,6 +63,13 @@ final class AddComponentCommand
     public $enabled;
 
     /**
+     * JSON meta data for the component.
+     *
+     * @var string|null
+     */
+    public $meta;
+
+    /**
      * The validation rules.
      *
      * @var string[]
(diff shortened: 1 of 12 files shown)


joecohens:
👍

PR #2265, opened 2016-12-07
Metric visibility flag was named incorrectly

mwillbanks:
The visibility flag was not properly wrapped into the array leading to
an error when adding in a metric as it would have an undefined index.

This would then crash adding in a metric from the view.   This was caused by PR #2261 in the merge commit 7dadef9


jbrooksuk:
Whoops! Thanks :)

Issue #1922, opened 2016-06-27
Metric points grouping


jbrooksuk:
Split from #1918 

1) when the data points and cut-off times don't exactly align - it lumps two data points in one period and none in the next - this pattern repeats.

2) instead of averaging, it adds. For instance, if it is metrics about latency - in milliseconds - then the sum doesn't make any sense. (FYI, I did set the "avg vs sum" parameter to "avg" - in case, you want to suggest that next.

czerwonkabartosz:
When it will be solved? ;)

jbrooksuk:
@GrahamCampbell any idea how we can fix these issues?

GrahamCampbell:
I'll think about this. :D

ConnorVG:
You could attach aggregates to metrics? Just an idea.

PR #2266, opened 2016-12-07
Metrics: Calculation set to average always sums

mwillbanks:
This commit is two part:

1. getActiveValueAttribute should always return the active attribute
without any further calculation here.  My guess is that this should also happen with sum but I was not going to modify it due to implications on the PR itself.

2. The query to create the average was incorrect and was taking the
count value * the value which we are really looking for the average of
the value.

### DIFF
--- a/app/Models/MetricPoint.php
+++ b/app/Models/MetricPoint.php
@@ -83,8 +83,6 @@ class MetricPoint extends Model implements HasPresenter
     {
         if ($this->metric->calc_type === Metric::CALC_SUM) {
             return round((float) $value * $this->counter, $this->metric->places);
-        } elseif ($this->metric->calc_type === Metric::CALC_AVG) {
-            return round((float) $value * $this->counter, $this->metric->places);
         }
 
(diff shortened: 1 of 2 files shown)


jbrooksuk:
Hmm, is this correct? The `counter` is used as a "how many times has this value been receieved in x time?" so surely we'd still need that, even for averages.

jbrooksuk:
> The query to create the average was incorrect and was taking the
count value * the value which we are really looking for the average of
the value.

No it wasn't? It was `value * counter`, then averaged that out.

mwillbanks:
@jbrooksuk so while it looks like that was the case, then there are other issues.  When I was pushing in values once per second, the value on the chart would increment, say an average of 12 would be quickly be 200+.  

So in this case, you would sum the value and sum the counter and then take the division over a specific time period.

jbrooksuk:
Do you need to change the threshold then?

mwillbanks:
Aha, this looks to be the bug that is mentioned in: #1922

PR #2271, opened 2016-12-16
Platform-2537 group status fix

harnash:
Disabled ordering components by `order` when determining group status.


harnash:
Sorry - didn't mean to merge it yet into upstream

PR #2278, opened 2016-12-22
Fixed the german translation

karaktaka:
I updated the German translation. No more mixing of formal and personal salutation and some other grammatically errors and typos.


ConnorVG:
Do we want to drop formality? (ie: "Du hast" -> "Sie haben")

karaktaka:
It was already mixed up. Sometimes "Sie" sometimes "Du". I could change it back to "Du", but i think you don't want to mix both salutations

ConnorVG:
It might be wise we pick one and stick with it, sure. Could you possibly contribute these to the [translation project](https://translate.cachethq.io/project/cachet)?

karaktaka:
Looks like it would take ages to check for typos there, but I'll try to contribute there. ;)

ConnorVG:
Thanks, we really appreciate it :smile:

PR #2289, opened 2017-01-02
Solely use Carbon

jbrooksuk:
We don't need to use jenssegers/date as Carbon should be doing everything we need.

### DIFF
--- a/app/Dates/DateFactory.php
+++ b/app/Dates/DateFactory.php
@@ -11,8 +11,8 @@
 
 namespace CachetHQ\Cachet\Dates;
 
+use Carbon\Carbon;
 use DateTimeZone;
-use Jenssegers\Date\Date;
 
 /**
  * This is the date factory class.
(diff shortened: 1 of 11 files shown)


joecohens:
Using `jenssegers/date` it's just because of the lang translations https://github.com/jenssegers/date#languages, it's a "wrapper" for carbon including all the translations ;)

jbrooksuk:
From what I can tell, Carbon has translations too? 
https://github.com/briannesbitt/Carbon/tree/master/src/Carbon/Lang

joecohens:
😱 They added them one month ago! That's very cool, now we can safely remove the package.

jbrooksuk:
Happy to merge this then? I can't see why it'd stop working. I've added `Carbon::setLocale` too, so that should help!

joecohens:
😉

GrahamCampbell:
Awesome to kill that unneeded library. :D

jbrooksuk:
Best thing is removing code!

PR #2290, opened 2017-01-02
✨ Add cachet:install command

jbrooksuk:
This magic command should install Cachet for the user, by following the given commands.

### DIFF
--- a/app/Console/Commands/InstallCommand.php
+++ b/app/Console/Commands/InstallCommand.php
@@ -11,7 +11,6 @@
 
 namespace CachetHQ\Cachet\Console\Commands;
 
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
What about the app install command?

jbrooksuk:
This does more of a setup than it does an installation, for people who don't want to use the setup screen.

GrahamCampbell:
I was meaning should this command run the install command?

jbrooksuk:
I don't think it can as we don't have any .env file settings.

sso-ot:
May I ask how do you run this to automate the setup??  I have my server up and running and I see the file but i'm not sure how to actually run this command in the background to fill out all the info

PR #2291, opened 2017-01-02
[WIP] Add user on cachet install command

joecohens:


jbrooksuk:
@joecohens I've added a way of fetching a fresh config, but the database still cannot be migrated. Perhaps @GrahamCampbell can chip in?

app/Console/Commands/InstallCommand.php
```diff
@@ -298,6 +302,33 @@ protected function configureCachet()
     }
 
     /**
+     * Configure the fisrt user.
```
GrahamCampbell: fisrt :P

app/Console/Commands/InstallCommand.php
```diff
@@ -338,6 +369,18 @@ protected function formatConfigsTable(array $config)
...
     /**
+     * Boot a fresh copy of the application configuration.
+     *
+     * @return void
+     */
+    protected function getFreshConfiguration()
+    {
+        $app = require $this->laravel->bootstrapPath().'/app.php';
+
+        $app->make('Illuminate\Contracts\Console\Kernel')->bootstrap();
```
GrahamCampbell: we could use class notation here :P

app/Console/Commands/InstallCommand.php
```diff
@@ -109,9 +111,9 @@ protected function configureDatabase(array $default = [])
         ], $default);
 
         $config['DB_DRIVER'] = $this->choice('Which database driver do you want to use?', [
-            'mysql'      => 'MySQL',
-            'postgresql' => 'PostgreSQL',
-            'sqlite'     => 'SQLite',
+            'mysql'  => 'MySQL',
+            'pgsql'  => 'PostgreSQL',
```
jbrooksuk: @joecohens we could use `PDO::getAvailableDrivers` and ignore SQL Server to generate this?

app/Console/Commands/InstallCommand.php
jbrooksuk: I copied this from the Laravel framework 🙈

app/Console/Commands/InstallCommand.php
GrahamCampbell: Instead of doing that, best to just hard code the 3 drivers we support, instead of calling getAvailableDrivers, because if laravel were to add a new driver, we wouldn't actually support it, so, best not to write code that says we do. :)

GrahamCampbell: I guess the most correct way to do it would be to filter laravel's returned results by a the 3 we do support, rather than removing the one we know we don't.

jbrooksuk:
@GrahamCampbell do you know what we were doing wrong here?

app/Console/Commands/InstallCommand.php
```diff
@@ -298,6 +303,33 @@ protected function configureCachet()
...
+    {
+        if (!$this->confirm('Do you want to create an admin user?')) {
+            return;
+        }
+
+        // We need to refresh the config to get access to the newly connected database.
+        $this->getFreshConfiguration();
+
+        // Now we need to install the application.
+        $this->call('app:install');
```
GrahamCampbell: Why is install calling install?

joecohens: Lol this is the `cachet:install` command not the `app:install`

GrahamCampbell: I wonder if it is really worth cachet loading in my laravel core package. Maybe it should just define it's own app commands, to replace the originals?

GrahamCampbell: My laravel core package is left over from the days when I was trying to create an extensible plugin system for a CMS. Harder than I thought it would be at the time. :trollface: The package still remains a kinda odd design. :P

joecohens: I think we can remove it and just use built in laravel commands and it's own

jbrooksuk: And just re-implement some of @GrahamCampbell's package?

GrahamCampbell:
What's the status of this PR @joecohens?

jbrooksuk:
This should be fixed and ready to go now. I had a user created!

PR #2314, opened 2017-01-16
Use unsigned tiny int instead of boolean for visible metrics

joecohens:
This field should be an int since we have 0, 1, 2 values https://github.com/CachetHQ/Cachet/blob/2.4/app/Models/Metric.php#L58

/cc @jbrooksuk


jbrooksuk:
Good spot!

joecohens:
Lol this is why I love Postgres, it squawks for everything :)

PR #2318, opened 2017-01-16
Component notifications can be silenced

### LINKED ISSUE #2316: Subscribers receive too much emails
The context is:

A cachet page that monitors the state of a service.
The service availability is verified every 10 minutes, and the cachet component is updated with the value.
Subscribers want to be warned whenever the service changes state.
Currently, the user is warned whenever the component state is updated, regardless of the value.

Solution: send notification email only when new information is available (for example: the component state is updated with a different status than the current one).

Workaround: update the component only when the value has actually changed. The problem with this is that the person looking at the component has no way of knowing the state of the service was actually monitored.

jbrooksuk:
Fixes #2316

---

You can now supply `silent` as a boolean parameter. If `true` then we won't notify about component status changes.

### DIFF
--- a/app/Bus/Commands/Component/UpdateComponentCommand.php
+++ b/app/Bus/Commands/Component/UpdateComponentCommand.php
@@ -79,6 +79,13 @@ final class UpdateComponentCommand
     public $meta;
 
     /**
+     * If this is true, we won't notify subscribers of the change.
+     *
+     * @var bool
+     */
+    public $silent;
+
+    /**
      * The validation rules.
      *
      * @var string[]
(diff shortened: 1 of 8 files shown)


joecohens:
@jbrooksuk Looks ok! This will silent all notifications for a component right? But what happens if I just don't want to notify users a single time and then they can keep getting updates?

jbrooksuk:
This is just per-update. If you silence once, the next time it won't be unless you specify again. Or am I misunderstanding?

joecohens:
Lol true, my bad 😉

jbrooksuk:
Merging this tonight :+1:

PR #2337, opened 2017-01-30
Laravel 5.4

jbrooksuk:
Because why the heck not?!

--- 

TODO:

- [x] Fix tests - `composer require laravel/browser-kit-testing`
- [x] Update remaining config files
- [x] Check all dependencies are up to date


base-zero:
@jbrooksuk Is it worth waiting and upgrading to Laravel 5.5 as it's an LTS version ?

ConnorVG:
@base-zero we would end up upgrading to 5.5 anyway, 5.4 would still be a middle step

jbrooksuk:
Doing the upgrade step-by-step is favoured. I'll try to finish this upgrade ASAP.

jbrooksuk:
Tests pass locally and I'm able to use the app myself. If anyone spots anything broken, hit me up!

PR #2381, opened 2017-02-20
Auto-link Twitter handles

### LINKED ISSUE #2380: Twitter Markdown extension
I've seen a few incidents around that link to Twitter pages. What do people think about auto-linking `@handle` text in Incident messages?

jbrooksuk:
Closes #2380

---

I'll merge this if enough people will find this beneficial. We could hide this behind a setting though.

### DIFF
--- a/app/Services/Markdown/Twitter/Extension.php
+++ b/app/Services/Markdown/Twitter/Extension.php
@@ -1,0 +1,51 @@
+<?php
+
+/*
+ * This file is part of Cachet.
+ *
+ * (c) Alt Three Services Limited
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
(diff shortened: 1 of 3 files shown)


mm-manageit:
I would vote for having it attached to a setting, as for our use cases I think it would be more annoying than helpful. Though on the flipside, can see it being useful for other use cases that utilize twitter alot for communication.

GrahamCampbell:
Would be better to have this as an Alt Three package.

👎 For the setting. Just have it on by default in the config, same as for emojis.

jbrooksuk:
> I would vote for having it attached to a setting

It'll be a setting.

> Would be better to have this as an Alt Three package.

Can do this.

jbrooksuk:
Replaced by #2404

PR #2394, opened 2017-03-01
Fix API bug and address visual bug in dashboard.

tmdvs:
Firstly, `UpdateIncidentCommandHandler` calls `UpdateComponentCommand`, however it was missing the `silent` parameter leading to incident updates to fail.

Secondly, The sidebar menu currently displays incorrectly on HiDPI/Retina displays. The badge on the Scheduled Maintenance link drops down and collides with the link below

Before:
Before

After:
After


jbrooksuk:
Thanks for this.

PR #2395, opened 2017-03-01
Add missing silent paramater, and cast status to Int before comparing

tmdvs:
It would appears that if an incident has no updates and you check the parent incident's status, it can often be a string rather than an Int. I noticed this was the case with incidents being created and updated via the Cachet API and a number of popular third party tools/monitors.

### DIFF
--- a/app/Bus/Handlers/Commands/Incident/UpdateIncidentCommandHandler.php
+++ b/app/Bus/Handlers/Commands/Incident/UpdateIncidentCommandHandler.php
@@ -97,7 +97,8 @@ class UpdateIncidentCommandHandler
                 null,
                 null,
                 null,
-                null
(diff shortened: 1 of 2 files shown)


tmdvs:
Hmm I thought so, however some updates created via the API seem to have strings for statuses. I had an incident with a status of 4 in the DB but if I dumped status in that comparison it was definitely a string.

jbrooksuk:
I couldn't merge this on my phone, sorry for the delay!

tmdvs:
No worries! Glad to have been able to contribute 👍🏼

jbrooksuk:
If you're into design stuff, we're looking for a new dashboard 😜 

I hope to see more contributions! :)

PR #2427, opened 2017-03-11
Update German E-Mail Text

silsha:


GrahamCampbell:
I think all translation changes need to go via crowdin.

silsha:
Ah, makes sense. I'll do that.

GrahamCampbell:
🍻

PR #2439, opened 2017-03-18
Cleanup event names

### LINKED ISSUE #2350: Standardise and simplify event names
Now that the event names are playing a more important part in Cachet, we should rename them to standardise and simplify their meaning.

For example:

- `ComponentStatusWasUpdatedEvent` could become `ComponentStatusChangedEvent`.
- `*WasAddedEvent` could become `*WasCreatedEvent`. (`Schedule` uses the `Created` scheme, whereas everything was added. Since we use `created_at` as a column name, it makes sense to follow this instead.
- `IncidentWasReportedEvent` should also be `IncidentWasCreatedEvent`.

jbrooksuk:
Closes #2350

---

I've renamed not just the events, but the commands too. This makes the whole system more standardised.


jbrooksuk:
Thanks StyleCI for fixing the class names!

PR #2440, opened 2017-03-18
Improve metric threshold

### LINKED ISSUE #2396: New Feature Enhancement: Maximum option for calculation of metrics display
It would be great if the "calculation of metrics" had an option for "max value". This would be helpful when attempting to report on the maximum or high water mark of a metric over a period of time.

image

jbrooksuk:
Closes #2396

---

@GrahamCampbell and I have previous discussed this.

### DIFF
--- a/app/Models/Metric.php
+++ b/app/Models/Metric.php
@@ -12,10 +12,12 @@
 namespace CachetHQ\Cachet\Models;
 
 use AltThree\Validator\ValidatingTrait;
+use AltThree\Validator\ValidationException;
 use CachetHQ\Cachet\Models\Traits\SortableTrait;
 use CachetHQ\Cachet\Presenters\MetricPresenter;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\MessageBag;
 use McCool\LaravelAutoPresenter\HasPresenter;
 
 class Metric extends Model implements HasPresenter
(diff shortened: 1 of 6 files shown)


app/Models/Metric.php
```diff
@@ -201,6 +210,26 @@ public function getShouldDisplayAttribute()
...
+     * @throws \AltThree\Validator\ValidationException
+     *
+     * @return void
+     */
+    public function validate()
+    {
+        $messages = [];
+
+        if (60 % $this->threshold !== 0) {
+            $messages[] = 'Threshold must be divisible by 60.';
```
GrahamCampbell: No, the threshold must **divide** 60. Not the same thing.

jbrooksuk: Fixed!

app/Models/Metric.php
```diff
@@ -221,7 +221,7 @@ public function validate()
         $messages = [];
 
         if (60 % $this->threshold !== 0) {
-            $messages[] = 'Threshold must be divisible by 60.';
+            $messages[] = 'Threshold must divide by 60.';
```
GrahamCampbell: Just `divide 60`, not by.

jbrooksuk: Not `into 60`? The wording doesn't make sense haha.

GrahamCampbell: We'd say `The threshold must divide 60.`, or we could notate this `t | 60` for `t divides 60`.

jbrooksuk: You're a mathematician though 😂

PR #2449, opened 2017-03-20
Use nicer logo

GrahamCampbell:
Does rely on the demo website tho.


jbrooksuk:
Why don't we just link to the asset?

GrahamCampbell:
Because github doesn't serve svgs. Try it. :P

PR #2481, opened 2017-04-12
add UNIX socket support for MySQL

duck-rh:
When using DB_HOST=localhost the driver already use the UNIX socket; unfortunately as I'm using an OpenShift container the socket path differs from the driver's default and installation fails. With this simple patch it is possible to pass the socket path to the driver and be happy :-).


jbrooksuk:
@GrahamCampbell could you review this change, please?

PR #2517, opened 2017-05-06
[WIP] Better inline form validation

jbrooksuk:
In my opinion, inline validation is far better than a single alert at the top of the page.


joecohens:
Agree! But I think you can still get the errors without validating directly in the controller, since we catch the validation exception and we populate the response with the errors, don't we?

jbrooksuk:
Yeah, we do. Maybe we just need to double check all of the command `$rules`?

PR #2570, opened 2017-06-13
Global Meta System

### LINKED ISSUE #2564: External Incident ID: Store ID + allow API update with external id
Hi,

I have a scenario, where the incident information comes from a different system, which is the leading system for incidents. In this system every incidents has a unique ID.

To bring this together with the really nice Cachet UI, I'd like to ask for the following:
- allow to store an external ID in the Cachet incident data &
- provide an additional API method, which allows to update an incident, which is identified by this external ID.

(The other systems would not easily be able to find out about cachet incident ids)

Kind regards,

Daniel

jbrooksuk:
Closes #2564

---

Meta is now something that can be applied anywhere thanks to polymorphic relations 🦄 

Currently the implementation is setup only on incidents (and creating them at that). Before applying this elsewhere, we should make meta its own set of commands and setup events. We also need a way to update meta but also delete a given key.

I'm sure there is a better way to insert polymorphic values too.

### DIFF
--- a/app/Bus/Commands/Incident/CreateIncidentCommand.php
+++ b/app/Bus/Commands/Incident/CreateIncidentCommand.php
@@ -97,6 +97,13 @@ final class CreateIncidentCommand
     public $template_vars;
 
     /**
+     * Meta key/value pairs.
+     *
+     * @var array
+     */
+    public $meta = [];
+
+    /**
      * The validation rules.
      *
      * @var string[]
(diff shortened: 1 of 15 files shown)


jbrooksuk:
Merging this as a beginning to a new system!

joecohens:
Nice!

jbrooksuk:
:)

PR #2571, opened 2017-06-13
Start using paratest

jbrooksuk:
Paratest for speed and as we're only going to add more and more tests, makes sense!


jbrooksuk:
Tests passed, rebase done.

PR #2572, opened 2017-06-13
[WIP] Bootstrap v4

jbrooksuk:
@joecohens what do you think about this?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -382,6 +382,15 @@
       "integrity": "sha1-g+9cqGCysy5KDe7e6MdxudtXRx4=",
       "dev": true
     },
+    "axios": {
+      "version": "0.16.2",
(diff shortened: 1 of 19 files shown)


joecohens:
It seems like a simple change :) is this working ok for you? Shouldn't we wait for a at least a beta version?

jbrooksuk:
Nah, it's nowhere near done yet, I don't think.

iamjulians:
Tested this on a dev build, looked fine and worked fine everywhere I could see 👍

jbrooksuk:
@iamjulians you wouldn't have had the compiled files, so you were only seeing source file changes. If you'd ran `gulp` you'd have seen issues.

iamjulians:
Oh. Makes sense.

PR #2581, opened 2017-06-19
System status now ignores hidden groups for guests

### LINKED ISSUE #2577: Overall status considers private components
We have some public (visible to guest users) and some private components. The public components are all working fine, just one of the internal one is currently down. Now the public status page shows the "Some systems are experiencing issues" alert though I am not logged in as a user. If someone visits the page as a guest, the overall status should not consider private components.

jbrooksuk:
Fixes #2577

---

This makes sense to me.

### DIFF
--- a/app/Integrations/Core/System.php
+++ b/app/Integrations/Core/System.php
@@ -42,7 +42,7 @@ class System implements SystemContract
      * Create a new system instance.
      *
      * @param \Illuminate\Contracts\Config\Repository $config
-     * @param \Illuminate\Contracts\Auth\Guard.       $auth
+     * @param \Illuminate\Contracts\Auth\Guard        $auth
      *
      * @return void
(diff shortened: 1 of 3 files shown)


app/Integrations/Core/System.php
```diff
@@ -31,15 +32,24 @@ class System implements SystemContract
...
+     *
+     * @var \Illuminate\Contracts\Auth\Guard
+     */
+    protected $auth;
+
+    /**
      * Create a new system instance.
      *
      * @param \Illuminate\Contracts\Config\Repository $config
+     * @param \Illuminate\Contracts\Auth\Guard.       $auth
```
GrahamCampbell: typo

jbrooksuk: 👍

PR #2583, opened 2017-06-19
Switch to using Laravel Mix

### LINKED ISSUE #2582: Use Laravel Mix

jbrooksuk:
Closes #2582

---

I really like the `extract` method, we just need to start writing the JS better before we can use it.


joecohens:
👍

Issue #2589, opened 2017-06-21
Incident Times Broken


iamjulians:
Running 2.4 branches code is mostly working, however I have noticed a strange bug when selecting date/time for Incident start (on incident edit/create screen). 

This happens both when not entering a date (which means it uses the current date) and when entering any custom date - the year (2017) and month (2017-07) work, but when entering the *time* and submitting, the hour of the time (e.g. 21) becomes the *date* (so in this case 2017-07-21), and the minute of the time is disregarded, the time becoming 08:00. 

^ Thats a little confusing, but in that example, the date, even if I enter `2017-07-20 21:00` would become `2017-07-21 08:00`.

I can provide screenshots etc. if needed, or more explanations, just tell me!

AdenFraser:
Running into the same issue on initial setup of 2.4 tag.

Will investigate tonight, fork and hopefully push up a PR — as this is currently preventing us from putting Cachet into prod.

PR #2593, opened 2017-06-23
Fixes Date issue with incident dates being incorrect (as only loading…

AdenFraser:
… out the date, not the time)

Fixes https://github.com/CachetHQ/Cachet/issues/2589


jbrooksuk:
Thanks!

PR #2595, opened 2017-06-24
Fix tests

jbrooksuk:


composer.json
```diff
@@ -33,7 +33,7 @@
         "alt-three/logger": "^1.3",
         "alt-three/throttle": "^1.0",
         "alt-three/twitter": "^1.0",
-        "alt-three/validator": "^1.5",
+        "alt-three/validator": "^2.0",
```
GrahamCampbell: Be careful with this. Things may break.

jbrooksuk: Do we know what like? :)

jbrooksuk:
Rebased. Hopefully extending the right class will fix PHP 5.6.

PR #2597, opened 2017-06-24
Shows in progress scheduled maintenances

oznorts:
This fixes an issue where an upcoming or in progress scheduled maintenance wouldn't show even when the completed date hasn't passed.

### DIFF
--- a/app/Models/Schedule.php
+++ b/app/Models/Schedule.php
@@ -160,7 +160,7 @@ class Schedule extends Model implements HasPresenter
      */
     public function scopeFutureSchedules($query)
     {
-        return $query->whereIn('status', [self::UPCOMING, self::IN_PROGRESS])->where('scheduled_at', '>=', Carbon::now());
+        return $query->whereIn('status', [self::UPCOMING, self::IN_PROGRESS])->where('completed_at', '>=', Carbon::now());
     }
 
     /**


jbrooksuk:
I'm not sure this change makes sense. Can you explain your thinking?

app/Models/Schedule.php
```diff
@@ -160,7 +160,7 @@ public function meta()
      */
     public function scopeFutureSchedules($query)
     {
-        return $query->whereIn('status', [self::UPCOMING, self::IN_PROGRESS])->where('scheduled_at', '>=', Carbon::now());
+        return $query->whereIn('status', [self::UPCOMING, self::IN_PROGRESS])->where('completed_at', '>=', Carbon::now());
```
GrahamCampbell: Is completed at always set?

oznorts:
Whoops, my bad. It doesn't make much sense!

What I was thinking was that the scheduled maintenances disappear right after the schedule time has passed, even if it's not completed. It should probably only return maintenances where the status is upcoming or in progress and not even take into account `scheduled_at` or `completed_at`.

jbrooksuk:
I think this should be part of the new design stuff instead of pushing it in like this. The current design doesn't lend itself to showing past data (aside from incidents).

PR #2640, opened 2017-07-18
Renamed Scheduled Maintenance to just Maintenance

### LINKED ISSUE #2516: Rename Scheduled Maintenance to just Maintenance?
Everything backend can remain the same, for instance the table is `schedules`  and the command namespace is `Schedule` but just for simplicity on the frontend I believe this would make it simpler?

jbrooksuk:
Closes #2516

---

There is more to rename here, but it'll do for now.


GrahamCampbell:
Meh. :trollface:

jbrooksuk:
Simplifies a bit.

PR #2642, opened 2017-07-18
Create incident_components table

jbrooksuk:
This creates the `incident_components` table and model. It's not used yet but we'll start to soon.

### DIFF
--- a/app/Models/IncidentComponent.php
+++ b/app/Models/IncidentComponent.php
@@ -1,0 +1,76 @@
+<?php
+
+/*
+ * This file is part of Cachet.
+ *
+ * (c) Alt Three Services Limited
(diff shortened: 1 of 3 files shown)


jbrooksuk:
@joecohens I'm tempted to add `status_id` for when https://github.com/CachetHQ/Cachet/pull/2643 is done. Would we want to store that data in here rather than on the component itself though?

joecohens:
If we want to "Cache" or potentially have history on the component statuses, it would be cool to have it ;)

jbrooksuk:
A history of statuses would be useful! Yep, will add it before I merge this.

jbrooksuk:
Merging.

PR #2643, opened 2017-07-18
[WIP] Custom Statuses

### LINKED ISSUE #2051: Add new component status for ongoing maintenance (e.g. UNDER MAINTENANCE)
During discussions in #2044 we came across the idea to add another status for components to be able to mark the component as "under maintenance". This would ease our maintenance workflow:
- Schedule a maintenance
- When maintenance is started set the status to "UNDER MAINTENANCE"
- When maintenance is done reset the status

We're also creating incidents via a checking service and so we are able to skip incident creation when the component is under maintenance.

jbrooksuk:
Closes #2051

---

Not a lot to see here yet. We're creating the `statuses` table and migrating existing statuses into said table.

Things to do:

- [ ] Think about whether we need to include an `order` column for each type?
- [ ] Migrate existing `incidents.status` data into new `statuses.id` value
- [ ] Drop existing `incidents.status` column
- [ ] Setup model relationships
- [ ] Update views to use new status value
- [ ] Update API

### DIFF
--- a/app/Models/Status.php
+++ b/app/Models/Status.php
@@ -1,0 +1,74 @@
+<?php
+
+/*
+ * This file is part of Cachet.
+ *
+ * (c) Alt Three Services Limited
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace CachetHQ\Cachet\Models;
+
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Model;
+
+/**
+ * This is the status class.
+ *
+ * @author James Brooks <james@alt-three.com>
+ */
+class Status extends Model
+{
+    /**
+     * The incidents type.
+     *
+     * @var string
+     */
+    const TYPE_INCIDENTS = 'incidents';
+
+    /**
+     * The components type.
+     *
+     * @var string
+     */
+    const TYPE_COMPONENTS = 'components';
+
+    /**
+     * The attributes that should be casted to native types.
+     *
+     * @var string[]
+     */
+    protected $casts = [
+        'name'  => 'string',
+        'slug'  => 'string',
+        'color' => 'string',
+        'type'  => 'string',
+    ];
+
+    /**
+     * The fillable properties.
+     *
+     * @var string[]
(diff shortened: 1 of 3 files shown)


jbrooksuk:
Note that the status values will now **not** be backwards compatible, making this a v3.0 release when the time comes.

joecohens:
Also doing this the statuses won't be translated automatically, are they?

jbrooksuk:
Correct. It's a shame because auto translations were nice, but this offers far more benefits.

jbrooksuk:
Unless we added `locale`...

ConnorVG:
Or we can just key the statuses rather than giving them (in the db) full representations. That'll just pass to trans as we sort of do now. Easy win.

This has been a long time coming ❤️

m1guelpf:
#~winter~statusesarecoming

jbrooksuk:
@ConnorVG what do you mean?

Geethree:
Any updates on this PR? This is the sole feature keeping us from using Catchet...

jbrooksuk:
@Geethree not yet I'm sorry! Needs more time and thought :)

Geethree:
Waiting.
Hoping.
Wishing.
For custom statuses!

jbrooksuk:
I promise they'll get here, there's just a lot of stuff to do elsewhere that may change the way this works :)

Geethree:
Meh. I just overwrote the appropriate lang resource and mounted that in the appropriate location as a docker volume. Perhaps that will work. /shrug

jbrooksuk:
@Geethree that's really the easiest way to do this if you're comfortable changing Cachet source files 🌈

JamesrenaudMitel:
Watching this PR, we had this on a previous system before moving to Cachet and would love to have the same functionality with Cachet

jbrooksuk:
@JamesrenaudMitel cool. I will pick this up again at some point.

Geethree:
Can it be picked up yet?

Geethree:
Is there anyway I could help with this issue?

jbrooksuk:
Note to self. Custom statuses in 3.x!

PR #2648, opened 2017-07-21
fixed and improved translations

Naspyy:
fixed all impolite translations to a more polite form to use the polite way throughout.


jbrooksuk:
Hey! Thanks for these however, we don't accept non-English translations here. Instead, please can you submit these to https://translate.cachethq.io so they can be peer reviewed etc :)

PR #2659, opened 2017-07-25
[WIP] Vue powered setup

jbrooksuk:
We'll switch to Vue.js at some point, so let's start with the setup screen.

TODO:

- [ ] Provide translations as a JSON object via an endpoint
- [ ] Pass through setup options so the form can be built
- [ ] Clean up the setup controller
- [ ] Add form-side validation

@joecohens thoughts? Anything we should do?


joecohens:
We are missing some configuration, but it's a good start! https://github.com/laravel/laravel/blob/master/resources/assets/js/bootstrap.js

jbrooksuk:
👍 

Maybe I should setup Vue on the 2.4 branch, then continue setup from here?

joecohens:
I think it would be better. Or just a PR to configure and setup Vue and from there we'll start adding functionality on different parts.

PR #2668, opened 2017-07-27
Use the includeWhen macro

jbrooksuk:
Simplify the views a little.


jbrooksuk:
Should be directive, but whatever.

GrahamCampbell:
> Should be directive, but whatever.

Just what I was thinking. 🌮

PR #2699, opened 2017-08-31
added auto-refresh on master status page

VeekeeFr:


judge2020:
Would be cool if it updated with perhaps angular but this is nice too.

jbrooksuk:
> Would be cool if it updated with perhaps angular but this is nice too.

The plan for the future is to use Vue.js on the frontend and dashboard 👍

jbrooksuk:
Thanks for this!

PR #2718, opened 2017-09-13
Fixing some additional php56 dependencies

VeekeeFr:
I just did a
composer update --no-dev -o
composer install --no-dev -o


jbrooksuk:
Thanks! Not sure what happened when I was doing it, perhaps a cache issue.

PR #2721, opened 2017-09-14
Faciliate Cachet deployment (new install or upgrade) using deploy.dog

violuke:
*Disclaimer: I'm the founder of the deploy.dog project, so will try to be as objective and unbiased as possible, but I obviously think deploy.dog is great. And I hope you do too.* :smile:

[deploy.dog](https://deploy.dog) is a recently-launched SAAS website deployment tool which is designed to deploy web applications to servers using SSH. It is incredibly flexible (and will continue to become more so) and allows for simple, single server deployments or deployments to complex infrastructure.

It works by looking at a [deploy.dog.yml](https://github.com/deploy-dog/Cachet/blob/2.4/deploy.dog.yml) file in the root of a project repository where the deployment configuration is defined. On private projects you would include [all config](http://local.deploy.dog/documentation/getting-started) in this file (with the exception of access keys, database passwords and other secrets). It means that the developers working on the project understand exactly how the deployment process works and can modify it as and when they need to. Deployments are managed using an intuitive web interface and all the heavy lifting is done by our servers (with other deployment triggers coming soon).

deploy.dog is free for small private projects and all open source projects (with no limits). We're about to start using the wonderful Cachet for our own status page and so have built the required config/script to deploy Cachet using deploy.dog. We'd like to share this with the community and would love it if deploy.dog could become a recommended deployment strategy for Cachet (along side the manual install and docker install strategies).

In this PR are the required files to facilitate the deployment though deploy.dog. I'd be more than happy to put together a documentation page showing how to easily use deploy.dog to install and update Cachet. To give you an idea, the other steps involved in using deploy.dog to run the deployment are:
1. Install Nginx/Apache/etc and point the root to `/var/www/code/public` (users can change this if they want, but that's the default that I've set). Also setup other services, such as MySQL/Redis/etc.
2. Register with deploy.dog (or login if you already have an account).
3. Add `CachetHQ/Cachet` as a public GitHub based project to your deploy.dog account.
4. Add secrets so we can securely store them for you, such as database password and ssh public key (within the Dashboard, you can generate a key pair if required).
5. Add your "Config Overrides" (set in the "Dashboard" page on the deploy.dog website) to set the things that are specific to you and cannot be shared between users. At the very minimum you need, the following, but other options can be overridden from their defaults if required:
```yaml
variables:
  cachet_url: http://status.example.com
  cachet_db_database: cachet
  cachet_db_username: cachet
  cachet_db_password: password # Here we recommend using a deploy.dog Secret instead of the actual password
  php_version: 7.1

servers:
  - name: status
    host: status.example.com
    port: 22
    user: www-data
    group: web # This is important, this is referenced in the deploy.dog.yml file
    stage: prod
    type: ssh
    sshKey: (d.d.secret:ssh-key-status) # This will be replaced with a secret value
```
6. Deploy at will.
7. Just deploy again to upgrade to the latest version.

I feel that deploy.dog is the perfect deployment tool for a project like Cachet. It can help any user install and upgrade. It can also really help on more complex deployments, such as where you have a highly-available version of Cachet running over multiple servers and you need to manage pushing out the update to multiple servers concurrently.

One major benefit to the Cachet team is that if you need to add new build/deployment steps, you simply add them to the project's deploy.dog.yml file and they'll automatically be handled by deploy.dog for anyone using it to install/update their instances of Cachet.

I'd love to hear your thoughts on making deploy.dog a suggested installation method for Cachet and any other suggestions on how to better integrate the two. I like to work closely with the Cachet team if you think we can work together to improve the deployment process.

Please let me know if you have any questions or if there is anything else that I need to explain.

Thank you.


deploy.dog/InstallOrUpdate.sh
```diff
@@ -0,0 +1,34 @@
...
+    # File found, we're upgrading
+    echo "Upgrading"
+
+    # Add key into .env file
+    KEY=$(head -n 1 ../../../shared/deploy.dog.cachet-key)
+    sed -i "s/DDREPLACECACHETKEY/$(echo $KEY | sed -e 's/\\/\\\\/g; s/\//\\\//g; s/&/\\\&/g')/" .env
+
+    # Upgrade Cachet
+    php artisan down
+    php artisan app:update
```
jbrooksuk: Don't forget `php artisan up`.

violuke: Thanks James. I've put the `php artisan up` bit in the `postSymlinkChangeTasks` section so that will mean that the site is in maintenance mode for very slightly longer, but we can ensure that when it comes back up, the code running will be the newly deployed version. See lines 10-12 and 29 of the `deploy.dog.yml` file. It might be better off in the `InstallOrUpdate.sh` file though as it does mean that that file is not the complete process in itself (which people might expect it to be)?

In fact, I don't think the `php artisan down` bit is quite right, sorry, as this will be running from the about-to-be-released directory and it should be within the currently running version (otherwise it's not going to have any impact). I'd overlooked how the up/down commands [actually work](https://www.reddit.com/r/laravel/comments/311i4q/what_is_php_artisan_down_actually_doing/).

I'll push a fix for this...

violuke: Ok, that last push should solve this. Sorry about that.

violuke:
Hi, what are your thoughts on this? Would you consider accepting this? I'm happy to write the extra documentation to accompany it. Thanks.

jbrooksuk:
@violuke sorry for the very long delay on this! I'll have a proper look soon :)

jbrooksuk:
Okay, so after thinking about this, I'm not keen on coupling deployment stuff into the core of Cachet. If you wanted to start a project wrapping this, like the Docker repository for example.

Thank you though, I appreciate the time and effort that you've put into this.

PR #2726, opened 2017-09-15
Don't send notifications when within maintenance window

### LINKED ISSUE #2720: Scheduled Maintenance Window should not email incidents
Not sure if I'm missing something but when we schedule maintenance windows and some of our components go un-operational, cachet shouldn't send mail out for these incidents during the window.

Currently we are syncing everything from zabbix to cachet and I do like the fact it still creates incidents and updates status's during the window so people can still see that things went down but the emails shouldn't send out as it alarms people. 

Either this or cachet should be put in a "maintenance" mode that displays "we are in a maintenance window" and doesn't send out incident mail.

jbrooksuk:
Fixes #2720

---

When maintenance is used correctly, this is pretty bad ass!


jbrooksuk:
Replaced by #2877

PR #2735, opened 2017-09-21
Do not rely on GLOB_BRACE as it is not available on non-glibc based systems

### LINKED ISSUE #2586: 2.4 installation fails on Alpine Linux
The installation of cachet 2.4 on Alpine Linux fails:
```
> php artisan optimize

                                                               
  [ErrorException]                                             
  Use of undefined constant GLOB_BRACE - assumed 'GLOB_BRACE'  
                                                               

Script php artisan optimize handling the post-install-cmd event returned with error code 1
```
GLOB_BRACE is not supported on musl-based systems - see https://github.com/zendframework/zend-stdlib/issues/58.

This line is possibly related:
https://github.com/CachetHQ/Cachet/blob/2.4/app/Foundation/Providers/RouteServiceProvider.php#L92

koendc:
This fixes Cachet 2.4 on eg the alpine-based [Docker](https://github.com/CachetHQ/Docker) deployment.
Fixes #2586.

### DIFF
--- a/app/Foundation/Providers/RouteServiceProvider.php
+++ b/app/Foundation/Providers/RouteServiceProvider.php
@@ -89,7 +89,10 @@ class RouteServiceProvider extends ServiceProvider
         $router->group(['namespace' => $this->namespace, 'as' => 'core::'], function (Router $router) {
             $path = app_path('Http/Routes');
 
-            foreach (array_merge(glob("$path/*.php"), glob("$path/*/*.php")) as $file) {
+            $AllFileIterator = new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($path));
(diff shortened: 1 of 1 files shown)


jbrooksuk:
I don't get why the current solution doesn't work as we only use it if it's defined, which, if you're on a non-glibc system, it'll fall back to 0.

koendc:
The glob expression contains braces which are not expanded by the glob filter if the GLOB_BRACE flag is not set. As a result the curly braces remain in the search expression and no php files are returned.

jbrooksuk:
@koendc I see... The problem with your fix is that we're still limited to one sub-directory though. Although this is fine for now, long-term we may want to introduce more routes.

koendc:
@jbrooksuk I made the search recursive in the last commit.

PR #2768, opened 2017-10-19
Fix bug Incident form not working properly on firefox (Component status field)

Diogo-Ferreira:
Hi !

When selecting the component the radio buttons didn't show on firefox. After looking at the code I found that old jquery code was still there for managing that field. This is now handled by Vue.Js (v-if).


jbrooksuk:
This just hides the issue by making the component status selector always visible, that's not what we want to do.

Diogo-Ferreira:
Sure, what do you suggest as a fix?

PR #2825, opened 2017-11-28
Add the possibility of only displaying incidents in the timeline

rarguelloF:
I added an option to just show incidents, no matter the distance between dates, in the timeline.

For example, if you had incidents on the days 2015-2-24, 2016-2-24, and 2017-4-12, and the `app_incident_days` option is set to `2`, you would see in the first page, all the incidents from 2017-4-12 and 2016-2-24, and in the second page (previous one), all the incidents from 2015-2-24.

This option takes for granted the `Only show days containing incidents in the timeline?` option (it doesn't display days without incidents).

## Screenshots

- New option in settings:

image

- Previous example illustrated:

image

image

Please let me know if you like this feature, or if you want me to change something in the code.

Kind regards,
Rodrigo

### DIFF
--- a/app/Http/Controllers/StatusPageController.php
+++ b/app/Http/Controllers/StatusPageController.php
@@ -45,36 +45,86 @@ class StatusPageController extends AbstractApiController
      */
     public function showIndex()
     {
-        $today = Date::now();
+        $only_disrupted_days = Config::get('setting.only_disrupted_days');
+        $appIncidentDays = (int) Config::get('setting.app_incident_days', 1);
+
+        // Used for the database query
         $startDate = Date::now();
+        $endDate = Date::now();
 
-        // Check if we have another starting date
-        if (Binput::has('start_date')) {
-            try {
-                // If date provided is valid
-                $oldDate = Date::createFromFormat('Y-m-d', Binput::get('start_date'));
+        $canPageForward = false;
+        $canPageBackward = false;
+        $previousDate = null;
+        $nextDate = null;
 
-                // If trying to get a future date fallback to today
-                if ($today->gt($oldDate)) {
-                    $startDate = $oldDate;
-                }
-            } catch (Exception $e) {
-                // Fallback to today
+        if ($only_disrupted_days) {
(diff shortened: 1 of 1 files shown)


jbrooksuk:
A lot of files have been changed when they don't need to be, things like translation files and `.gitignore`. Could you please undo some of these.

New translations also need to go to CrowdIn; https://translate.cachethq.io

rarguelloF:
Ok, thanks for your response!

One of the changes I did on the translation files was removing the space on the `[0,1]` and `[2,Inf]` text chains, since it's currently being displayed. Do you want me to create a separate PR with this (since these are not really changes on the translation)?

rarguelloF:
@jbrooksuk I already removed the translation changes 👍

resources/views/dashboard/settings/app-setup.blade.php
```diff
@@ -134,6 +134,17 @@
                                 </div>
                             </div>
                         </div>
+                        <div class="row">
+                            <div class="col-xs-12">
+                                <div class="checkbox">
+                                    <label>
+                                        <input type="hidden" value="0" name="app_only_display_incidents">
+                                        <input type="checkbox" value="1" name="app_only_display_incidents" {{ Config::get('setting.app_only_display_incidents') ? 'checked' : null }}>
```
jbrooksuk: We don't use the `Config` facade here, instead we pass through a value, see how the rest are done 👍

rarguelloF:
@jbrooksuk what do you think of using the "Only show days containing incidents in the timeline?" configuration option for this behaviour?

resources/views/dashboard/settings/app-setup.blade.php
rarguelloF: Done 👍

jbrooksuk:
@rarguelloF I much prefer it :)

rarguelloF:
@jbrooksuk changes already done 👍

jbrooksuk:
Looks good! Thank you.

PR #2831, opened 2017-12-01
Get app_name from env

andreyev:
A small change to setup app name from env vars.


jbrooksuk:
But the name of the application is Cachet.

We tend to substitute `app.name` with `setting.app_name`, so I'll close this and open a PR which does that reliably.

PR #2846, opened 2017-12-07
Use assertClassNotHasAttribute

carusogabriel:


jbrooksuk:
What's the benefit to this?

carusogabriel:
@jbrooksuk The error message. With `assertFalse`:
```
Failed asserting that true is false.
```
With `assertClassNotHasAttribute`:
```
Failed asserting that class "stdClass" has attribute "foo".
```

jbrooksuk:
Handy! Thanks. I've restarted Travis and it passed, so all is good.

PR #2866, opened 2018-01-12
 Makes the editor focusable.

uxen-ab:
On the incident template creation the editor was not focusable, now it
is.
The problem was the initialization of the editor that was done before
the DOM is fully loaded.
Using addEventListener and DOMContentLoaded fixes the problem and is
compatible from IE9.


jbrooksuk:
Thanks! 👍

PR #2870, opened 2018-01-14
Autofocus totp input

sedrubal:
It is quite nasty to click into the text input for the totp token after login. It should autofocus like the login form.


jbrooksuk:
Thanks!

PR #2874, opened 2018-01-17
Bugs in the average metrics graph

uxen-ab:
## Problems

1. The suffix wasn't displayed in the tooltip when mouseovering a metric point.
2. Metric points added in the future could be displayed on the graph.
3. When using the "Last 12 hours" view if it was less than 12pm the last hours of the previous day were displayed, but ChartJS put them at the end of the graph.

> The second point could seems useless, but since there are some problems with timezone when posting a metric point and the date when it's stored, it's possible to have a shift between the real date of posting, and the date stored/displayed.

## How they were fixed

1. The JavaScript syntax was fixed in order to execute the "tootip" callback.
2. A `AND created_ad < NOW()` was added in the different syntaxes for the different repositories in order to filter the future.
3. Select the datetime instead of only the time of every metric point and then cut this string in the tooltip and legend, in order to CharJS be able to order them.


jbrooksuk:
This looks great, thanks!

PR #2875, opened 2018-01-18
Incident update problems issue 2775

uxen-ab:
Fixes a problem in the API, all the updates of all the incidents were returned when requesting all updates for one incident.

It's the first item of the issue 2775.


jbrooksuk:
Thanks for this however, there is a slightly better way of doing this.

I've fixed this in https://github.com/CachetHQ/Cachet/commit/7a4ec0535d9ff024db9592fdfbf7f7b433e18c58

PR #2877, opened 2018-01-21
Suppress notifications while in maintenance mode

nstapelbroek:
Hi again :raising_hand_man: 

After reading #2720 I thought about a way to handle notifications when there is a scheduled maintenance. I cherry-picked some code of #2726 to help me get started on this.

### What has been done:
- Implemented setting to "suppress notifications when maintenance mode is enabled". I did this because  I could imagine users being happy to receive partial downtime notifications when during maintenance as a reminder.
- Used the code of @jbrooksuk and fixed a potential cause why it din't work for @DP19 in #2720 
- Every CRUD page regarding incident's now get a alert informing the users that notifications are suppressed.

### How to test:
- Make sure [your Cachet is up and running](https://docs.cachethq.io/docs/installing-cachet).
- Log into the dashboard and create a new maintenance period
- Make sure your maintenance period is occurring right now
- After saving the scheduled maintenance go to your settings and verify that the "Suppress notifications when incident occurs during maintenance period?" checkbox is checked
- Create a new incident
- Acknowledge the message informing you about notification suppression.
- Acknowledge that the notification is suppressed after saving your new incident

### Notes:
- I've only tested this locally using the mail notification channel. I don't want to assume that it also works for the Slack notifications so maybe someone could help me out on this one.
- Are we missing some generic "heads up, maintenance occurring" alerts on the dashboard and component overview pages? I would love to provide them in another PR.


jbrooksuk:
Great, thank you for this - you're on a bit of a roll at the moment aren't you!

> I've only tested this locally using the mail notification channel. I don't want to assume that it also works for the Slack notifications so maybe someone could help me out on this one.

All notifications should work the same way, it's just one class which fires them all, so this will be fine.

> Are we missing some generic "heads up, maintenance occurring" alerts on the dashboard and component overview pages? I would love to provide them in another PR.

I'd never thought about it, but yes, I'd say so!

PR #2893, opened 2018-01-27
Fix invalid amount of arguments in CreateIncidentUpdateCommandTest

nstapelbroek:
Heya :raising_hand_man: ,

I noticed how the build was failing. Turns out the arugments passed to the constructor of `CreateIncidentUpdateCommand` werent patched on two locations. This should do the Trick.

### What has been done:
- The Api/IncidentUpdateController now uses the component_id and component_status when creating an incident update.

### How to test
- The Travis build should be green now


jbrooksuk:
Awesome!

Issue #2812, opened 2017-11-15
Feature request: Need to be logged in to see status page


schniepern:
Hi everyone! :)

In our company we would first only like to make our status page available to logged in users (and be able to turn this off for a later point). I tried to implement this myself without doing too many changes to the source code but I ran into a few problems:
- The way the routing is structured, doing this in one centralized place (like RouteServiceProvider.php) isn't easily possible - it makes the code messy and since I'm not very experienced with Laravel, it would also be kind of unpredictable. Here's what I've done (in RouteServiceProvider.php):
```php
$alwaysAuthenticate = true;
if($alwaysAuthenticate) {
    $router->group(['namespace' => $this->namespace], function (Router $router) {
        $needAuth = ['FeedRoutes', 'StatusPageRoutes', 'SubscribeRoutes'];
        $needAuthBasenames = [];
        foreach (glob(app_path('Http//Routes').'/*.php') as $file) {
            $basename = basename($file, '.php');
            if(in_array($basename, $needAuth)) {
                $needAuthBasenames[] = $basename;
            } else {
                $this->app->make('CachetHQ\\Cachet\\Http\\Routes\\'.$basename)->map($router);
            }
        }
        $router->group(['middleware' => ['web', 'auth']], function (Router $router) use ($needAuthBasenames) {
            foreach($needAuthBasenames as $basename) {
                $this->app->make('CachetHQ\\Cachet\\Http\\Routes\\'.$basename)->map($router);
            }
        });
    });
} else { // original code
    $router->group(['namespace' => $this->namespace], function (Router $router) {
        foreach (glob(app_path('Http//Routes').'/*.php') as $file) {
            $this->app->make('CachetHQ\\Cachet\\Http\\Routes\\'.basename($file, '.php'))->map($router);
        }
    });
}
```
$alwaysAuthenticate would of course later be configurable in the settings.

This code defines a list of additional Routes that need authentication and creates a group with the necessary middleware for only them. It's of course just a messy workaround - it works (as far as I tested, which is to say there might be some bugs a more experienced Laravel developer would know about) but you have to add every *Routes.php file and the 'web' middleware is sometimes added twice (it needs to be added to ensure they handle the request in the right order).

Of course it would be possible to just add the 'auth' middleware in the *Routes.php files under the condition that the feature is enabled, but this would result in a lot of redundant code.

### In conclusion:
- This would be a very interesting and useful feature to implement directly into Cachet
[...]

PR #2894, opened 2018-01-27
Work in Progress: Feature always authenticate

nstapelbroek:
Hi again :raising_hand_man: 

A trial implementation of #2812. In this state I would advice against merging this because I had to implement a lot of hacks / workarounds to get this working nicely. I took a lot of shortcuts and would love some input before continuing. 

### What has been done:
- Added a new setting in the security section allowing you to enforce authentication on all routes.
- Added logic to the RouteServiceProvider that will read this configuration value and inject `auth` or `api.auth` middleware into the routing group.
- When you have the always authenticate feature enabled, we wont show you a "back to home" button on the login

### How to test:
- Make sure [your Cachet is up and running](https://docs.cachethq.io/docs/installing-cachet).
- Log into the dashboard and enable the `Always authenticate` checkbox at the security settings
- Logout
- Acknowledge you cannot visit the status page without authenticating
- Acknowledge you cannot visit the api routes like `api/v1/components` without authenticating

### Todo:
- [ ] Get rid of the hacks
- [ ] Write tests

### Notes:
- See inline notes in code

### DIFF
--- a/app/Foundation/Providers/RouteServiceProvider.php
+++ b/app/Foundation/Providers/RouteServiceProvider.php
@@ -16,6 +16,8 @@ use CachetHQ\Cachet\Http\Middleware\Acceptable;
 use CachetHQ\Cachet\Http\Middleware\Authenticate;
 use CachetHQ\Cachet\Http\Middleware\Timezone;
 use CachetHQ\Cachet\Http\Routes\AuthRoutes;
+use CachetHQ\Cachet\Http\Routes\Setup\ApiRoutes;
+use CachetHQ\Cachet\Http\Routes\SetupRoutes;
 use Illuminate\Cookie\Middleware\AddQueuedCookiesToResponse;
 use Illuminate\Cookie\Middleware\EncryptCookies;
 use Illuminate\Foundation\Http\Middleware\VerifyCsrfToken;
@@ -42,6 +44,15 @@ class RouteServiceProvider extends ServiceProvider
      * @var string
      */
     protected $namespace = 'CachetHQ\Cachet\Http\Controllers';
+
+    /**
+     * These are the route files that should always be available anonymously.
+     *
+     * When applying the always_authenticate feature, these routes will be skipped.
+     *
+     * @var string[]
+     */
+    protected $whitelistedAuthRoutes = [AuthRoutes::class, SetupRoutes::class, ApiRoutes::class];
 
     /**
      * Define the route model bindings, pattern filters, etc.
@@ -129,7 +140,8 @@ class RouteServiceProvider extends ServiceProvider
             SubstituteBindings::class,
         ];
 
-        if ($this->app['config']->get('setting.always_authenticate', false) && !$routes instanceof AuthRoutes) {
+        $applyAlwaysAuthenticate = $this->app['config']->get('setting.always_authenticate', false);
+        if ($applyAlwaysAuthenticate && !$this->isWhiteListedAuthRoute($routes)) {
             $middleware[] = Authenticate::class;
         }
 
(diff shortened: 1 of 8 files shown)


app/Foundation/Providers/RouteServiceProvider.php
```diff
@@ -149,8 +170,30 @@ protected function mapOtherwise(Router $router, $routes)
             Timezone::class,
         ];
 
+        if ($applyAlwaysAuthenticate && !$this->isWhiteListedAuthRoute($routes)) {
+            $middleware[] = 'auth.api:true';
```
nstapelbroek: It would be cool if I could use the `::class` class name resolution. But how do I implement this with arguments?

app/Http/Controllers/Dashboard/SettingsController.php
```diff
@@ -384,6 +385,10 @@ public function postSettings()
             Lang::setLocale(Binput::get('app_locale'));
         }
 
+        if (Binput::has('always_authenticate')) {
+            Artisan::call('route:cache');
```
nstapelbroek: A workaround I've had to implement because the `cachet:setup` command caches the routes. After enabling the `always_authenticate` feature your routing cache becomes invalid. If the original approach on solving this issue is correct. What should we do with this code? Some idea's: 

1. Use DI for resolving Artisan or even the `route:cache` command?
2. Let caches be handled by a different `service` that listens to some `settingsupdated` event. Because this controller has too many responsibilities.

resources/assets/js/cachet.js
```diff
@@ -16,6 +16,9 @@ $(function () {
         beforeSend: function (xhr) {
             xhr.setRequestHeader('Accept', 'application/json');
             // xhr.setRequestHeader('Content-Type', 'application/json; charset=utf-8');
+          if (typeof window.apiKey !== 'undefined') {
```
nstapelbroek: I really dislike this workaround. The problem originates from an API call that is fired whenever you open the dashboard.
Because the API is also "firewalled" you need the proper header in your request. Any tips on what to do with this?

resources/views/layout/dashboard.blade.php
```diff
@@ -62,4 +62,7 @@
 </body>
 @yield('js')
 <script src="{{ mix('dist/js/all.js') }}"></script>
+<script type="text/javascript">
+    window.apiKey = "{{ auth()->user()->api_key }}";
```
nstapelbroek: Part 2 of the workaround in `resources/assets/js/cachet.js`. My preference is to get rid of this change.

jbrooksuk:
I shall review this tonight :)

nstapelbroek:
@jbrooksuk thanks :+1: , I'm currently stuck trying to write a unit test that asserts the middleware injected when the config value is set.

nstapelbroek:
@jbrooksuk It's ready! thank you for your patience

nstapelbroek:
@jbrooksuk ping :see_no_evil:

app/Foundation/Providers/RouteServiceProvider.php
jbrooksuk: I don't think you can, @GrahamCampbell?

jbrooksuk:
This looks good!

nstapelbroek:
Alrighty :+1:

jbrooksuk:
Don’t forget to update where you use the config setting as well.

nstapelbroek:
Turns out I was using the new and correct path in [the code]( https://github.com/CachetHQ/Cachet/pull/2894/files#diff-76499814cc0f6334b1479f1102463c37R113) before I changed it.

PR #2912, opened 2018-02-14
Add ID of scheduled maintenance to list group item

shuichiro-makigaki:
This allows us to use a static URL to a scheduled maintenance.
e.g) http://localhost:8000/#scheduled-1


shuichiro-makigaki:
@jbrooksuk ping?

jbrooksuk:
Great! Thanks :)

PR #2922, opened 2018-02-23
Fixes the template editing as it was done in pull #2866 for creation.

uxen-ab:
In the pull request #2866 the template creation was fixed by loading CodeMirror only after DOM loaded, the same thing is done here with editing.


jbrooksuk:
Thanks for this!

Issue #2903, opened 2018-02-05
Update documentation about Cachet API


hugomcfonseca:
Hello. 

I would like to know if it is possible to update Cachet documentation about its API (preferably, including already new routes added for Cachet v2.4). I am trying to improve existing Golang client to be compatible with your latest Cachet's version.

Thanks and congrats about your great job with this app.

anthonybocci:
Hello,

I'll take a look if [laravel-apidoc-generator](https://github.com/mpociot/laravel-apidoc-generator) does the job. I would be really great if documentation was updated automatically.

anthonybocci:
I'm only at the beginning step and I've documented three controllers, probably not fully, but if you want to test you can use my [automatic-api-doc](https://github.com/anthonybocci/Cachet/tree/automatic-api-doc) branch.
To generate the doc, execute `php artisan api:generate --routePrefix="api/v1/*"`and open the file `public/docs/index.html`.
I'll customize that later.

anthonybocci:
This issues throws an other question: the API documentation could be updated or generated via command line, but what about the rest of the documentation? It could be really great to have a `resources/docs` directory that contains markdown files that explain all the install process and other. They could be updated easier.

jbrooksuk:
We could use something like Swagger?

hugomcfonseca:
Yes, it would be fine by me @jbrooksuk. And Swagger will help to keep documentation updated every time a new endpoint is added/changed/deleted, so it would be great.

anthonybocci:
The package [DarkaOnLine/L5-Swagger](https://github.com/DarkaOnLine/L5-Swagger) seems to be the job.
I'll update my pull request to add the SwaggerPHP annotations in addition to the documentation block I've ever added.

jbrooksuk:
I'm keen on this being generated to support Readme.io

anthonybocci:
If we use ApiDoc (as discussed in #2929) there would be no need of Swagger and we could be synced with Readme.io.

anthonybocci:
It could be great if a package would be able to use the Laravel's routes and the doc block to generate Swagger. [Swagger-php][1] needs all the information, but perhaps an other package or a wrapper would be able to do more?

[1]: https://github.com/zircote/swagger-php

jbrooksuk:
I don't know much about Swagger, but yeah it would be very cool!

anthonybocci:
[...]

PR #2929, opened 2018-03-01
Generation of the API documentation

anthonybocci:
The [online documentation](https://docs.cachethq.io/reference) seems to be out of date and as a developer it is difficult to think about update the online documentation at the same time as developing an API method. An other problem with the online documentation is that it is required to have an Internet connection to view it, so when developing in a train or something it's not possible to see the documentation. The documentation is the subject of #2903 
The ideal would be to be able to generate the documentation when we need it. Doing so, we will be able to generate our own documentation on our local machine during development, and it doesn't prevent an online documentation that may be updated more frequently and more easily.

To do so I've included the [mpociot/laravel-apidoc-generator](https://github.com/mpociot/laravel-apidoc-generator) package that generates the API documentation using the routes and the doc blocks. So, a developer updating the API only has to update the doc block.

To generate the documentation you can execute the command
`php artisan api:generate --routePrefix="api/v1/*"`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /node_modules
 /public/hot
 /public/storage
+/public/docs
 /storage/*.key
 /vendor
 /.idea
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ Homestead.json
 Homestead.yaml
 npm-debug.log
 .env
+config/swagger-constants.php
+storage/api-docs/
--- a/app/Http/Controllers/Api/AbstractApiController.php
+++ b/app/Http/Controllers/Api/AbstractApiController.php
@@ -19,6 +19,17 @@ use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\Response;
 use McCool\LaravelAutoPresenter\Facades\AutoPresenter;
 
+/**
+ * @SWG\Swagger(
+ *   schemes={"https", "http"},
+ *   host="example.com",
+ *   basePath="/api/v1",
+ *   @SWG\Info(
+ *     title="Cachet API",
+ *     version="1.0.0"
+ *   )
+ * )
+ */
 abstract class AbstractApiController extends Controller
 {
     /**
--- a/app/Http/Controllers/Api/ComponentController.php
+++ b/app/Http/Controllers/Api/ComponentController.php
@@ -34,53 +34,75 @@ class ComponentController extends AbstractApiController
      *
      * @SWG\Get(
      *   path="/components",
-     *   summary="List all components",
-     *   operationId="index",
+     *   summary="List all components.",
+     *   operationId="Component@index",
      *   tags={"Components"},
      *   produces={"application/json"}, 
      *   @SWG\Parameter(
-     *      description="ID of the component to filter",
-     *      in="query",
-     *      name="id",
-     *      required=false,
-     *      type="integer",
-     *      format="int64"
+     *     description="ID of the component to filter.",
+     *     in="query",
+     *     name="id",
+     *     required=false,
+     *     type="integer",
+     *     format="int64"
      *   ),
      *   @SWG\Parameter(
-     *      description="Name of the components to filter",
-     *      in="query",
-     *      name="name",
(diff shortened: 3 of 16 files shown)


jbrooksuk:
Interesting!

I don't have a much time at the moment, but please leave this with me and I'll take a look when I can. Perhaps @joecohens could help out here?

anthonybocci:
I'm glad to know you are interested!

I understand that you haven't lot of time, when you have a moment to touch that you can leave a comment here, I'm available.

I've set doc blocks above the API methods by _mixing_ the content of the online documentation an what I've found in the code.

jbrooksuk:
I've just checked the branch out and it generated really nicely. I'd be keen on keeping documentation in sync with the hosted docs though. Is there something we could do to sync with https://readme-sync.readme.io/docs/full-example?

anthonybocci:
It seems readme.io uses the ApiDoc standard. So I could update some things in the doc blocks in order to use ApiDoc instead. If so, the only thing to so would be generating the file `./doc/api_data.json` on the hosted doc.
It is pretty easy to do, I can do that it you're ok with that.

joecohens:
I think that Readme feature using apiDoc is kind of stale, but we could do it using swagger https://readme.readme.io/docs/swagger-1

anthonybocci:
It may be a good idea, Swagger is a well known standard. I know the package [DarkaOnLine/L5-Swagger][1] that generates Swagger from php doc, but it's compatible with PHP >= 7.1, do you know an other package that is compatible with PHP >=5.6?

[1]: https://github.com/DarkaOnLine/L5-Swagger

jbrooksuk:
I don't sorry, are any of their older versions compatible?

Do we need to support older versions of PHP to generate documentation though?

anthonybocci:
An older version of the package (4.0 I think) is compatible with our version of PHP. But as said in the associated issue, should we write all the Swagger, or may a part be generated using the Laravel's routes?

jbrooksuk:
We could write Swagger but that means it'll be out of date in two places potentially.

anthonybocci:
Yes that's the problem, we could write Swagger ourselves ignoring Laravel's routes, but the problem is we could be outdated in the swagger. It's the same.
We aren't the firsts to use Swagger with Laravel, there is probably someone who uses Laravel's routes with Swagger in order to keep being synced.

jbrooksuk:
http://blog.cjwfuller.com/2015/05/30/swaggervel/ Does this help?

anthonybocci:
@jbrooksuk as you can see in the "Controller Annotation" section, the path is in the doc block.

```
@SWG\Api(
    path="/animals",
    @SWG\Operation(
        method="GET",
	summary="Fetch animals"
    )
)
```

The `/animals` path is the URL. If in our Laravel's routes we update the route, the one configured above becomes wrong.
In an ideal world, the annotation in the php doc would look for the `path` in the Laravel's routes for the current controller/action/method.

jbrooksuk:
Wow, lots of changes here, awesome!

anthonybocci:
Yes I have not finishes yet! But it's in progress. I'll try to finish that this weekend.

GrahamCampbell:
Thanks for the PR @anthonybocci. Is this PR complete now? Is there anything outstanding we need to look at?

Also, can I ask why autopresenter has been removed?

anthonybocci:
@GrahamCampbell this pull is old now, but I just checked and some controllers are not annotated so it's not finished. I'd a bit forgotten this to be honest. I have to finish it.

For autopresenter, I don't know. I don't know what it is really. What is it doing?

GrahamCampbell:
Autopresenter is heavily used in the views. All models are replaced with their presented versions in the environment for view rendering.

anthonybocci:
Ok I understand what it is and why it's useful. I don't remember removing it. It's probably a mistake.

jbrooksuk:
@anthonybocci are you happy if we close this and go down the manual route for now?

anthonybocci:
@jbrooksuk Yes you can. I am overwhelmed these times so I didn't finish this issue.

PR #2942, opened 2018-03-17
[WIP] Add ability to force confirmations of install

### LINKED ISSUE #2924: Automating the Cachet installation
Hi,

On Slack I discussed with someone that needed to automate the installation of Cachet, and because I'm interested by this feature I create this issue in order to contribute then.

The command `app:install` creates the tables in the database, but does not fill the `settings` nor the `users` table.
I would like to add an optional parameter (or create a subclass) to `app:install` in order to give as parameter the path to a file that contains the settings to write into the database. So, the installation would be fully automated.

For now even after executing `app:install` the database is empty, so it's mandatory to launch the setup page that asks for a user, mail drivers, app name and so on.
Also if I've already set the `.env` file with Cache driver, Mail driver and so on, is it required to ask for them on the setup page? Or at least pre-fill the select box could be a help.

I've looked in the `app/Console/Command` directory but all the commands inside are `cachet:*` and I don't know where the `app:install` command is located. Also, do you have guidelines about that? Like the usage of a json or any other file type to manage this configuration.

Regards,

Cachet 2.4
Commit number 2d8bf2fa791511a1ac0fe7ff6e4b4b4e0efd0ca8

jbrooksuk:
Closes #2924

---

I'd recommend piping to `yes` but that's not available on Windows (that I know of).

### DIFF
--- a/app/Console/Commands/InstallCommand.php
+++ b/app/Console/Commands/InstallCommand.php
@@ -30,11 +30,25 @@ class InstallCommand extends Command
     protected $name = 'cachet:install';
 
     /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'cachet:install {--force}';
+
+    /**
      * The console command description.
      *
      * @var string
      */
     protected $description = 'Install Cachet';
+
+    /**
+     * Do we want to agree to all confirmations?
+     *
+     * @var bool
+     */
+    protected $force = false;
 
     /**
      * Execute the console command.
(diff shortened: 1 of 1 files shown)


jbrooksuk:
Ping @anthonybocci

jbrooksuk:
@anthonybocci can you test this and let me know if it worked as you expected, please?

anthonybocci:
@jbrooksuk I've tested and it seems there are some choices remaining, the installation is not 100% automatic.

I have the `DB_DRIVER` `mysql` in my `.env` file, but the command asks me what driver I want to use.
It guess it is because `$this->force` is used only in the `fire()` method, no? The other methods continue to ask for informations.

GrahamCampbell:
This is not the way to go. We should be already respecting the standard `--no-interaction` flag, which instructs console applications that they cannot expect to be able to read from stdin.

GrahamCampbell:
I'll file a bug report to this affect, we can check all our commands behave themselves.

PR #2946, opened 2018-03-18
Fixed sub-title

BaileyJM02:


jbrooksuk:
Heh, this often happens but I never know what causes it. Thanks for fixing.

PR #2950, opened 2018-03-19
Add perma-link to each scheduled maintenance

shuichiro-makigaki:
PR #2912 adds scheduled maintenance ID to overview HTML.
This patch introduces permanent link and icon image by the ID.


jbrooksuk:
Thanks for this! :)

Issue #2956, opened 2018-03-23
$redirectUrl is undefined with 3 uses in Dashboard/SettingsController


christianklemp-imt:
Hello,
I have Cachet deployed using Docker and am trying to upload a PNG image for my site banner. I have also tried a JPG, but get a 500 error due to `$redirectUrl` on line 332 being undefined. It also appears twice before this line, but am not hitting those.

https://github.com/CachetHQ/Cachet/blob/93d1f8784a75b5aa975f8cb89007e94ee63bb1e4/app/Http/Controllers/Dashboard/SettingsController.php#L324

anthonybocci:
Hi,
You have the problem only because your file was invalid. If the file was valid, it would worked.
I've explained that in the commit that fix it.

christianklemp-imt:
I find it odd that the image (tried as PNG and JPG formats) both fail. They are both images that I exported in those formats, so not sure what is making it fail.

anthonybocci:
I've tried to add an image on my Cachet, it works well. Have you tried with an other image?
I think it's not a problem with Cachet, come on Slack, the community will be glad to help you.

jbrooksuk:
I think it could be image dimensions?

christianklemp-imt:
Thanks for the help guys! I'll do a little more debugging, but may jump onto Slack if I can't get it figured out since my image is 930px x 100px. I've even tried it with a 900px x 100px and a 250px x 100px image and am getting the same result.

anthonybocci:
I suggest you to come on Slack indeed, it will be easier to discuss about your problem and fix it. If there is a real problem in Cachet we'll post it onto Github, but I don't think so. I'm "Anthony B" on Slack, don't hesitate to message me during the day!

jbrooksuk:
Looking at `SettingsController`, it's not based on dimensions, sorry for misleading you there.

It could be that the servers max file size is too small? 

Do you not get any error messages at all?

PR #2957, opened 2018-03-23
Redirect to setting page on image uploading error

anthonybocci:
In the settings we can upload an image as banner. There are some
verifications that are done on the file to check that it's a valid
image.
If the file is not valid, a redirection is done to the theme page with
an error message.

When redirecting to the theme page because of an error on the file, the
controller redirects to an URL set in a variable. The problem was this
variable '$redirectUrl' was undefined, so the controller thrown an
error. The problem occured only if there was an error on the file, but
it worked well if the file was a valid image.

In the controller's constructor a 'submenu' field is set with panel
items and url, so the variable '$redirectUrl' is set to the theme
page URL. Now if there is an error with the file, the user is well
redirected and the error is displayed.

See: CachetHQ/Cachet#2956

### DIFF
--- a/app/Http/Controllers/Dashboard/SettingsController.php
+++ b/app/Http/Controllers/Dashboard/SettingsController.php
@@ -397,6 +397,7 @@ class SettingsController extends Controller
     protected function handleUpdateBanner(Repository $setting)
     {
         $file = Binput::file('app_banner');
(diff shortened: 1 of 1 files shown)


jbrooksuk:
🎉

PR #2958, opened 2018-03-24
Fix metrics fill theme color application

### LINKED ISSUE #2797: Metric graph color issue
Hi Team,

I am running 2.4 ( upgraded from 2.3). However, I am noticing that the Metrics color is changed in 2.4 and the new color isn't good as it is dark/gray. 

When I compared it with https://dev.cachethq.io/ I do see a different color on the dashboard page. 
[
color
](url)

I've also compared Settings -> Theme page on both the installations and those are same values. Is there anything missing on my end that I am seeing this color instead of the color displaying on https://dev.cachethq.io/ ? 

Here is the screenshot of the Theme page 

color2

Ps: I also noticed that on https://dev.cachethq.io/  the color preview isn't working, but the color preview is working on my installation.

JordyvanDortmont:
Fixes  #2797.

This restores the application of the metrics fill theme color, which was broken in 3ea1e959f966887cb30961ba932073cf51e5555f.


jbrooksuk:
Great, thanks @JordyvanDortmont!

PR #2959, opened 2018-03-24
Fix SQLite metric points retrieval datetime

JordyvanDortmont:
This fixes the timezone issues with the retrieval of the metric points for the graphs.

Adding 'localtime' as a modifier to the ``datetime`` function in the ``SqliteRepository`` ensures equivalence to the use of ``NOW``, which uses the system timezone by default, in the ``MySqlRepository``.


jbrooksuk:
Heh, I didn't know that. Do you know of any issues that this fixes?

jbrooksuk:
No problem! Thanks :)

PR #2960, opened 2018-03-24
Use the table prefix on metric_points

anthonybocci:
In the '.env' file, a 'DB_PREFIX' sets the prefix that should be used
on every table name. When writing an SQL query the 'DB_PREFIX' value
has to be prefixed to the table name.

The repositories PgSqlRepository, MySqlRepository and SqliteRepository,
located in 'app/Repositories/Metric/' did not apply this prefix on
the 'metric_points' table. The problem occured only when we set a
'DB_PREFIX' not null, the rest of the application worked correctly but
the part about 'metric_points' couldn't work, saying the table was
inexistant.

A method was added in the repository AbstractMetricRepository to get
the 'metric_points' table name with the prefix if there is one.
This method is used in the three repositories to get the right table
name.

Note: This problem was present in 2.3, but was already fixed in 2.4 so
there is no need to apply this commit on the 2.4 branch.

See: CachetHQ/Cachet/#2955

### DIFF
--- a/app/Repositories/Metric/AbstractMetricRepository.php
+++ b/app/Repositories/Metric/AbstractMetricRepository.php
@@ -48,16 +48,37 @@ abstract class AbstractMetricRepository
     }
 
     /**
+     * Get the table names prefix.
+     *
+     * @return  string
+     */
+    protected function getPrefix()
+    {
+        $driver = $this->config->get('database.default');
+        $connection = $this->config->get('database.connections.'.$driver);
+        $prefix = $connection['prefix'];
+        return $prefix;
+    }
+
+    /**
      * Get the metrics table name.
      *
      * @return string
      */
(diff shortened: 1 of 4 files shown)


jbrooksuk:
I wish we could have used Eloquent for this, it'd have saved so many headaches.

jbrooksuk:
By the way, thank you for fixing this. I'll tag a v2.3 release.

anthonybocci:
In 2.3 (or 3.0, we'd say that in the roadmap) we could use Eloquent for all queries.

Naramsim:
Please, update the docker image on Docker Hub. It contains this bug. @anthonybocci @jbrooksuk

jbrooksuk:
@djdefi is responsible for the Docker image :)

PR #2969, opened 2018-03-27
Fix superfluous metric point

JordyvanDortmont:
Previously a metric point with the default value would always be added, because there were only 60 metric points retrieved while the loop to fill in gaps with default values executed 61 times.


jbrooksuk:
Replaced by https://github.com/CachetHQ/Cachet/pull/2970

PR #2970, opened 2018-03-27
Fix trailing default metric points

### LINKED ISSUE #2539: Do not create a tailing default point if still in threshold
Metric-Graph drops to zero (= my default value) when url-monitor has not yet reported a new point.

This is a show-stopper for metric-graphs, because our clients might get confused and think our service just died. See screenshot below.

The graph should not create a default point if there was a point reporting during threshold-time (which also might be 1min).

bildschirmfoto 2017-05-22 um 11 57 03

JordyvanDortmont:
Fixes #2539.

The last hour view now shifts to an earlier time frame by one minute for each missing or currently unavailable metric point per minute until the threshold is reached, after which default metric points are used.

In addition to taking into account the threshold this also retrieves the correct number of minutes for the last hour, 61 instead of 60 to allow even steps in the x-axis without any unnecessary leading or trailing default metric points.

### DIFF
--- a/app/Repositories/Metric/MetricRepository.php
+++ b/app/Repositories/Metric/MetricRepository.php
@@ -63,7 +63,7 @@ class MetricRepository
         $pointKey = $dateTime->format('Y-m-d H:i');
         $points = $this->repository->getPointsSinceMinutes($metric, 60)->pluck('value', 'key')->take(60);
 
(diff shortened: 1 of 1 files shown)


jbrooksuk:
Thanks! Will close your other PR.

PR #3005, opened 2018-04-10
Cast status and component_id to int

NHellFire:
/api/v1/incident is returning them as strings, should be int.


jbrooksuk:
Thanks!

Issue #3004, opened 2018-04-09
Should Component/tag be a ManyToMany relation?


anthonybocci:
**Affected feature** : Components (create and update at least)
**How to reproduce**:
- Go to `/dashboard/components/create`
- Fill the form
- Submit

**Exception thrown**: `BadMethodCallException: Call to undefined method Illuminate\Database\Query\Builder::sync()`.
**Note**: The operation is done on a fresh install and a fresh database. The component is successfully created but the exception is thrown when calling calling `$component->tags()->sync()`

The main problem is in ComponentController:210, the `sync()` method is declared only for the ManyToMany relations.
It seems Component and Tag are used as if they had a ManyToMany relation, but they are not declared like so. Also, there is no table `component_tag`.

Cachet version: 2.4 (3c51e5115a187c063efbc7a2e1167fa579b01aa0)
PHP 7.2.3

Please see the full log:
```
[2018-04-09 17:49:31] development.ERROR: BadMethodCallException: Call to undefined method Illuminate\Database\Query\Builder::sync() in /home/anthony/public_html/Cachet-test/vendor/laravel/framework/src/Illuminate/Database/Query/Builder.php:2445
Stack trace:
#0 /home/anthony/public_html/Cachet-test/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php(1254): Illuminate\Database\Query\Builder->__call('sync', Array)
#1 [internal function]: Illuminate\Database\Eloquent\Builder->__call('sync', Array)
#2 /home/anthony/public_html/Cachet-test/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Relations/Relation.php(356): call_user_func_array(Array, Array)
#3 /home/anthony/public_html/Cachet-test/app/Http/Controllers/Dashboard/ComponentController.php(210): Illuminate\Database\Eloquent\Relations\Relation->__call('sync', Array)
#4 [internal function]: CachetHQ\Cachet\Http\Controllers\Dashboard\ComponentController->createComponentAction()
#5 /home/anthony/public_html/Cachet-test/vendor/laravel/framework/src/Illuminate/Routing/Controller.php(55): call_user_func_array(Array, Array)
#6 /home/anthony/public_html/Cachet-test/vendor/laravel/framework/src/Illuminate/Routing/ControllerDispatcher.php(44): Illuminate\Routing\Controller->callAction('createComponent...', Array)
[...]

PR #3017, opened 2018-04-20
Corrected link component tag : morphMany => morphToMany cf.#3004

vincedom:
The relation between component and tag should be morphToMany rather than morphMany, else we get the error described in #3004


jbrooksuk:
Thanks!

PR #2325, opened 2017-01-23
[WIP] New design

jbrooksuk:
This is another attempt at the new design with some various changes from the original. Ping @joecohens :)

Some key changes & features:

- Unified header and navigation bar
- Simple "Powered by Cachet" footer (still optional) but can display the timezone too
- Improved component designs, including status icons
- A proper design for Incident Updates
- We only show the last `x` (default 3) days of incidents. There is a link to take you to an overview of the previous incidents

PR #3022, opened 2018-04-24
Add class to incident date area

shuichiro-makigaki:
This allows us to apply styles to incident date fields.
The day of the week classes are useful for week-dependent styles.


jbrooksuk:
Thank you for this. I’m actually planning on redesigning the frontend, so as this is for 2.4 I won’t be accepting this PR 👍🏻

shuichiro-makigaki:
@jbrooksuk I see, and I am looking forward to seeing the new design!
Where can I follow the progress? (may be #2325 ?)

jbrooksuk:
We'll likely scrap that PR and start again. I'm quite keen on splitting the partials further, so that we can re-use them as AJAX components in the future.

PR #3043, opened 2018-05-10
Metrics timezone follows application timezone (for MySQL)

shuichiro-makigaki:
Metric points are stored in UTC, but viewed in application timezone.
This patch can conform timezone of incidents and scheduled maintenances to metric points.


shuichiro-makigaki:
Sorry, this patch is not required because of my fault setting (not setting APP_TIMEZOME).
Closed.

jbrooksuk:
No problem. Glad you’re sorted.

PR #3071, opened 2018-06-14
Cast status to int on update component command

opengd:
Currently the `status` is a `string` but is need to be a `int`. Due this the check if the component `status` has changed on creation of a new incident will always be `false` in `SendComponentUpdateEmailNotificationHandler`, and it will notify all subscribers.


jbrooksuk:
Great! Thanks for fixing this.

PR #3086, opened 2018-06-16
[On Hold] Drop /v1 from endpoint

### LINKED ISSUE #3084: Drop versioning from API endpoints?
What are people's thoughts on dropping the versioning (`/v1`) chunk from the API endpoint?

Here's why I don't like it:

- It suggests that Cachet will support multiple versions of the endpoints in one installation - this is not true.
- It adds needless length to the endpoints, even though it's only three characters it adds up!
- I'm feeling restricted in what changes we can make to the endpoints before we end up having to move it into `v2` (and thus drop all existing endpoints).
- The next version of Cachet will be v3 and not v2.4 so we’re free to make backwards breaking changes where needed. 

Let me know what you think.

jbrooksuk:
Closes #3084

---

This is based on the assumption that #3084 is a go. So for now, we're on hold.

### DIFF
--- a/app/Http/Routes/ApiRoutes.php
+++ b/app/Http/Routes/ApiRoutes.php
@@ -37,8 +37,8 @@ class ApiRoutes
     public function map(Registrar $router)
     {
         $router->group([
-            'namespace'  => 'Api',
-            'prefix'     => 'api/v1',
+            'namespace' => 'Api',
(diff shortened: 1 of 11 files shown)


jbrooksuk:
Too much is changing at the moment and it's a 2 minute change.

Issue #2669, opened 2017-07-27
Remove GrahamCampbell/Laravel-Core


jbrooksuk:
The package is now abandoned.

We should move things into the `cachet` signature.

GrahamCampbell:
👍

GrahamCampbell:
All the events stuff is overkill BS that was in the core package originally. We can just have 3 console commands that do shit. :P

jbrooksuk:
I need to get on with this so we can move to Laravel 5.5 at some point.

jbrooksuk:
Closed by https://github.com/CachetHQ/Cachet/pull/3092

PR #3092, opened 2018-06-17
Replace GrahamCampbell/Core

jbrooksuk:
@GrahamCampbell for now I've basically copied the code out of your package as we already have the architecture needed to support it.

Like you said in #2669, the events are overkill so we should refactor them out at some point.


jbrooksuk:
@GrahamCampbell do you know why the tests fail on this one? The commands run just fine.

PR #3093, opened 2018-06-17
Smoke Tests

### LINKED ISSUE #3068: Smoke tests
We should add Smoke tests to every possible URL, ensuring that by default a page/endpoint will return an expected status code.

This should make it easier to know where issues are introduced, if we go from a 200 to a 4xx/5xx unexpectedly.

jbrooksuk:
Closes #3068

---

Smoke tests are a simple, but useful way of testing that endpoints work, without submitting any data.

### DIFF
--- a/app/Http/Controllers/Api/GeneralController.php
+++ b/app/Http/Controllers/Api/GeneralController.php
@@ -17,7 +17,7 @@ use CachetHQ\Cachet\Integrations\Contracts\System;
 /**
  * This is the general api controller.
  *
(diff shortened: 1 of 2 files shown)


shuichiro-makigaki:
In my opinion, this PR can be merged at this point even if it contains only tiny tests.
Because, by using this commit as skeletons, other contributors can add tests easily.
It will take long time to make complete smoke tests, and others can/should help :)

jbrooksuk:
Good point, I shall merge this in later.

PR #3100, opened 2018-06-17
Fix mail markdown notifications

### LINKED ISSUE #2966: Scheduled maintenance is not visible enough in the dashboard
When one wants to add or edit a scheduled maintenance, they need to first go to the "Incidents" page, then click on the "Scheduled maintenance" sub-section to finally find the list of current scheduled maintenance events and the button to add a scheduled maintenance.

This is too well hidden for most users. It'd be better if scheduled maintenance would have a link in the menu on the left for direct access instead of being a sub-section of Incidents.

jbrooksuk:
Closes #2966

---

Mail notifications should use Markdown views, this will need applying to **all** notifications in the near future, but does require slight refactoring.

image

### DIFF
--- a/resources/views/notifications/incident/new.blade.php
+++ b/resources/views/notifications/incident/new.blade.php
@@ -8,10 +8,8 @@
 @endcomponent
 
 Thanks,<br>
-{{ config('app.name') }}
+{{ Config::get('setting.app_name') }}
 
-@component('mail::subcopy')
-[{{ $unsubscribeText }}]({{ $unsubscribeUrl }}) &mdash; [{{ $manageSubscriptionText }}]({{ $manageSubscriptionUrl }})
-@endcomponent
+@include('notifications.partials.subscription')
 
 @endcomponent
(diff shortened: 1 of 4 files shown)


resources/views/notifications/incident/new.blade.php
```diff
@@ -0,0 +1,17 @@
...
+# {{ trans('notifications.incident.new.mail.greeting', ['app_name' => Config::get('setting.app_name')]) }}
+
+{{ $incident->message }}
+
+@component('mail::button', ['url' => $actionUrl])
+{{ $actionText }}
+@endcomponent
+
+Thanks,<br>
+{{ config('app.name') }}
```
nstapelbroek: This line is different with the greeting above: `Config::get('setting.app_name')`

jbrooksuk:
Good spot! I’ll rebase, make this change and merge.

sontt710:
Thank @nstapelbroek @jbrooksuk

Issue #2508, opened 2017-05-01
Updating Incidents Issues


gonzoinc:
Couple of issues I found when going in and Updating an Incident.

- [x] When you make the update and save, the corresponding screen states:
Awesome. The incident has been deleted and will not show on your status page.

![screen shot 2017-05-01 at 1 39 06 pm](https://cloud.githubusercontent.com/assets/27728728/25587821/9fadae92-2e73-11e7-8bee-d47d1b8885ba.png)

- [ ] if you create an incident and it is initially set to Not notify subscribers, Updates to the incident still send email notifications. No options to disable this in the Update view. Updates should abide by the initial setting or allow the option to not notify subscribers on the update screen.

- [ ] Setting an incident to fixed should open the option to allow the service to be set back to Operational. Since you allow the service status option to be available when you create the incident, it should also exist when you resolve the incident.

jbrooksuk:
What version of Cachet are you using?

jbrooksuk:
Sorry, obviously it's 2.4 since you're using Incident Updates.

jbrooksuk:
Issue one is fixed.

chronicals:
Issue 3 really is something critical in my opinion, since it's not possible to change the state of a service via the api.

jbrooksuk:
@gonzoinc @chronicals I've started point 2 in https://github.com/CachetHQ/Cachet/pull/3103

jbrooksuk:
Thank you for your input on Cachet 2.x. We are shifting our attention and resources to Cachet 3.x and will no longer be supporting the 2.x version. If your feedback or issue is relevant to the 3.x series, we encourage you to engage with the new branch.

For more information on the Cachet rebuild and our plans for 3.x, you can read the announcement [here](https://github.com/CachetHQ/Cachet/discussions/4342).

We appreciate your understanding and look forward to your contributions to the new version.

PR #3103, opened 2018-06-17
Add notifications column to incidents

jbrooksuk:
Part of https://github.com/CachetHQ/Cachet/issues/2508

With this change, Incident Updates will default to inherit the Incidents notification setting i.e. if you create an incident and opt not to notify people, then the Incident Update notifications will be silence by default (you'll be able to select to notify though).


jbrooksuk:
Merging now in prep.

PR #3120, opened 2018-06-20
Add timezones to API results

### LINKED ISSUE #2691: Add timezone to dates returned via the API
The current dates are returned as `2017-08-24 13:23:48`, so we need to know the system's timezone to be able to fully identify this timestamp. It would be very useful if the timezone was part of the date strings.

My use case is that I'm trying to set up integration between CachetHQ and UptimeRobot. I want to be able to run the script that copies data from UptimeRobot multiple times, without creating new metric points after each run (i.e. it'll be idempotent). This mean that I need a way to identify which metric points weren't added yet. The ideal would be to have a consistent id shared by UptimeRobot and CachetHQ, but as that isn't currently possible, the workaround is to only submit points with a timestamp that isn't in CachetHQ yet. This has the obvious drawback that it'll ignore multiple metric points with the same timestamp, but that's unlikely in my case as the interval is 1~5 minutes.

jbrooksuk:
Closes #2691

---

Timezones aren't returned in Cachet. We need to do something about it.

### DIFF
--- a/app/Presenters/Traits/TimestampsTrait.php
+++ b/app/Presenters/Traits/TimestampsTrait.php
@@ -30,7 +30,7 @@ trait TimestampsTrait
      */
     public function created_at()
     {
-        return app(DateFactory::class)->make($this->wrappedObject->created_at)->toDateTimeString();
+        return app(DateFactory::class)->make($this->wrappedObject->created_at)->format('c');
     }
 
     /**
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Tests are failing, and needs rebasing. I wish GitHub had a rebase button in the UI. :/

jbrooksuk:
Rebased.

PR #3134, opened 2018-06-23
Create documentation about incidents and metrics

anthonybocci:
Explain what is a metric
Explain how to create a metric
Explain what is an incident and how it could be used.


jbrooksuk:
Thanks!

PR #3137, opened 2018-06-26
Theming Support

jbrooksuk:
Theming Cachet has been something I've been thinking about for a while. I appreciate that people have different opinions on what makes a good status page and as Cachet can't be to every taste out of the box, it makes sense to allow people to theme Cachet themselves.

This system is something I've been only spent a day writing code for, so there's probably many things I've not thought about yet.

It works like this:

1. The default theme is now within `./resources/views/vendor/theme`. This is the namespace that we've given to the core code.
2. Themes are installed using `composer.
3. Now that we're using Laravel 5.6 we get access to "Package Auto Discovery" which means themes can be installed and then automatically register themselves into Cachet's theme manager.
4. There is a `SetActiveTheme` middleware that'll pull a `theme` setting and apply it. By default, if there is no theme it'll use the default.
5. The way it's currently setup is that themes can build upon the default or replace it altogether because it's set as the fallback. In the future we could add an option to the manifest (`theme.json`) which tells Cachet not to do this.

You can find an example theme (that replaces the status page index) at https://github.com/jbrooksuk/cachet-test-theme.

Note the following:

- The theme has both `composer.json` and `theme.json` manifest files. The `composer.json` is required to install the package. `theme.json` is needed to register information about the theme. In the future the dashboard will list all of themes.
- Themes can replace the dashboard.
- The theme system is essentially a plugin system, but we're only registering views. In the near future (if this works well enough) I'd like to migrate this system into a package and open it to registering migrations, routes etc to become a full plugin system.

Let me know your thoughts!

P.s. I wouldn't start building themes until we've completed the new design because we need to make it more obvious and intuitive how themes work and apply data.


jbrooksuk:
Once we're happy with how this works, I'm going to move the theming code into its own package at https://github.com/CachetHQ/Cocoon

jbrooksuk:
I might split this into a couple of PRs:

1. Move all files into a `theme` namespace.
2. Install Cocoon and replace default theme with [Phoebe](https://github.com/CachetHQ/Phoebe).

PR #3140, opened 2018-06-28
Changed app_banner_type on two locations due to banner image is currently not working

opengd:
The variables app_banner_type have not been changed on two location. Due to this banner image will not work and a error is produced on status page and on the dashbord/settings/themes page if any banner image is set.


jbrooksuk:
Great spot, thank you!

Issue #3144, opened 2018-06-29
Error 500 -> Undefined variable: appAnalyticsPiwikURL


krollebol:
Yesterday I updated to the latest 2.4 git version. But there is an issue when Piwik is configured in Settings, resulting in error 500:

[2018-06-29 12:38:47] production.ERROR: ErrorException: Undefined variable: appAnalyticsPiwikURL in /var/www/Cachet-dev/storage/framework/views/403732977efb6ebb644a
6875942728eb7c1c5356.php:37

Also the settings page for Analytics gives an error 500.

I guess something was forgotten when updating to the new Laravel and the many changes to variable names.

Thanks for looking into this!

krollebol:
I think I found the issue. At some places the variables have different capital letters.
In resources/views/partials/analytics.blade.php:
on line 34: $appAnalyticsPiwikSiteI**D** must be $appAnalyticsPiwikSiteI**d**

in resources/views/dashboard/settings/analytics.blade.php
l38: $appAnalyticsPiwikU**RL** must be $appAnalyticsPiwikU**rl**
l46: $appAnalyticsPiwikSiteI**D** must be $appAnalyticsPiwikSiteI**d**

jbrooksuk:
Sorry about that!

Are you okay to submit a Pull Request which fixes this?

jbrooksuk:
Closed in #3145

PR #3145, opened 2018-06-29
Variable correction for Piwik analytics

krollebol:
As discussed in issue #3144, corrected some typos in variables used for Piwik analytics.


jbrooksuk:
Thank you for this! Sorry that it happened. It was missed as part of the upgrade to Laravel 5.6

PR #3149, opened 2018-07-02
Fix error Undefined variable: config when running cachet:install

nstapelbroek:
Hello There, 

This PR should prevent an error when installing cachet and answering "no" to the following questions:
```
 Do you wish to use Cachet Beacon? (yes/no) [no]:
 > no

 Do you wish to use Emoji? This requires a GitHub oAuth Token! (yes/no) [no]:
 > no

```


jbrooksuk:
Thanks Nico!

PR #3150, opened 2018-07-02
Merge install commands

nstapelbroek:
Heya,

This PR will merge the available install commands into one in the hopes of preventing confusion. I also came accross some issues with the laravel backup tool. I left my opinions in the Notes section :see_no_evil: 

### What has been done
- The app:install has been removed, the contents have been migrated to the cachet:install
- cachet:install will ask to configure your cachet before running the installation steps
- Hardcoded the diskname to check for database backups to the `php artisan backup:list` fetches from the right directory

### How to test
- Checkout PR on a clean state
- Run the `php artisan cachet:install`
- Acknowledge the configuration wizzard
- Acknowledge that it installs after configuring the application

### Notes
I had some problems with the laravel-backup package. I'm not sure if you would want to take up the responsibility of making backups as cachet anyway as this creates a lot of edge cases. If you want to continue, here are some possible improvements:
- The database targets for backup are currently hardcoded to mysql. it should fetch from the value from `config('database.driver')` right? Sadly I can't change it due to file load order or closure statements giving me a hard time when caching the configuration. Might be worth investigating in the upstream package :thinking: 
- The notifications e-mail has the same problem. I would like to use a config value set in your installation somewhere.

I'm just thinking out loud here. Let me know if you have any suggestions :smile:


jbrooksuk:
I think you may be right about the backups. It's not much work to backup the database manually before running an update and requires one less overhead for us to maintain.

Do you want to scrap the backup from this branch as well?

nstapelbroek:
@jbrooksuk I'll remove laravel-backup in a seperate PR if that's ok :)

jbrooksuk:
That's fine. Merged in for you,

PR #3158, opened 2018-07-04
Fix the incident update display

uxen-ab:
When an incident update had many lines, the design was a bit broken. The
logo of the incident update was above the text, but it's supposed to be
on the left.

A 'div' has been added around them, and the width of the item has been
managed.


uxen-ab:
When inserting a link in an incident update the design is broken. It was the case before this PR, I'll check why.

jbrooksuk:
I’m on holiday at the moment, but will have this reviewed once I’m back.

PR #3191, opened 2018-07-19
Added GET incident templates to use in POST incidents template

TakeMeNL:
Atm there was no way to get the incident template slug from the API, while we can use it to create a incident through the api.


jbrooksuk:
This is great! Thanks @TakeMeNL!

PR #3195, opened 2018-07-24
Get config values from repository

shuichiro-makigaki:
When Cachet is running in multiple instances such as Docker,
changing config values is not synced among the instances.
This patch fixed the way to get config values to fix the sync problem.

### DIFF
--- a/app/Http/Controllers/StatusPageController.php
+++ b/app/Http/Controllers/StatusPageController.php
@@ -18,6 +18,7 @@ use CachetHQ\Cachet\Models\Incident;
 use CachetHQ\Cachet\Models\Metric;
 use CachetHQ\Cachet\Models\Schedule;
 use CachetHQ\Cachet\Repositories\Metric\MetricRepository;
+use CachetHQ\Cachet\Settings\Repository;
 use CachetHQ\Cachet\Services\Dates\DateFactory;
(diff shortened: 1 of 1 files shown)


shuichiro-makigaki:
hmm, the night passed, and I found that this patch was not so good.
Cachet gets config values from a cache at many places of codes.
We need to rewrite the config repository module to achieve the goal that I aimed in this patch.

> setup.

In Laravel, configurations are cached in a PHP instance, and not expired.
(maybe, this is because Laravel thinks configs as "static".)
Therefore, configs in another PHP instance can't be synced with others.

My ideas are:
* Store configs in `CACHE_DRIVER`, and set retention timerange
* Store configs in RDBMS, and get them from the DB every time

(too much work!)

jbrooksuk:
Thanks for this!

I'll try to review this, this week :)

jbrooksuk:
@GrahamCampbell I remember that you wrote the original setting repository code, have you any thoughts on this?

GrahamCampbell:
Yes, this looks like code left over from when settings used to be implemented by using the settings key directly in the config.

GrahamCampbell:
Hmm, maybe not actually. I'll need to remind myself how this stuff works.

PR #3223, opened 2018-08-20
Fixes Template not rendering when containing twig variables

### LINKED ISSUE #3219: Edit Template page doesn't render when it contains Twig syntax
My company (nameless for now) is looking at using Cachet for a public facing status page with Nagios integration, and it's looking promising.

The problem I'm encountering my break it though. The templates will be maintained by a separate team, with common variables between them all. The problem is that when a template contains any twig variables, like:
```
 {{ message}}
```
The whole page doesn't get rendered. It still exists, and parses properly, but the maintainers won't be able to modify them.

stobbsm:
Fixes #3219


stobbsm:
> What does `v-pre` do?

`v-pre` will exclude the text-area from template parsing. When placing twig template variables in the text-area, the front-end will try to yield the value of the given variable instead of the plain-text containing the variable. Most notable on template responses.

Example: Wanting to use the variable "time" in the template (to pass a variable of the same name to the template when being rendered).
```
The outage occured at {{time}}
```

Would result in a blank page, as the variable from "time" doesn't exist on the page. `v-pre` makes sure to not parse the contents of the text-area.

jbrooksuk:
Gotcha!

PR #3236, opened 2018-09-06
Show schedules till they are completed

Wouter0100:
Currently schedules are shown on the status page when they are in the status upcoming or in progress. But, when the scheduled time is in the past - it's hidden, even when it's still going on and regardless of the completed time entered.

This PR fixes this issue by hiding the maintenance when it's in the status completed, or the completed time has passed.


jbrooksuk:
This makes sense.

Thanks for this!

Wouter0100:
@thomasxd24 are you sure? We're using this in production, which seems to work as promised.

Issue #3249, opened 2018-09-21
Add a link to a Twitter page


waqark3389:
I am trying to add a link to a Twitter page just below the banner. In the older versions it was possible to add a "About this page" section but this seems to have disappeared now. Whats the alternative now? 

I'm not sure what version I am on as the VERSION file in the root directory is showing version:  2.3.9 but on the dashboard it is showing 2.4.0-dev.

anthonybocci:
This is not supposed to be in the issues, so thank you for having closed it and moved on Slack.

As you said there is an "About this page" field in the settings, and that's this information that is supposed to be shown on the status page (as you can see on the demo).
What you pointed out is there was a problem with that, this information was no longer shown.

I've fixed that point, I'll submit a pull request as soon as possible.

In de172e5defa4ac226044e3066413487fdee49de0 the VERSION file contains 2.4.0-dev, I guess you are not up to date.

PR #3251, opened 2018-09-21
Include the partial "about-app" in index

anthonybocci:
In the settings it's possible to fill a field "About this page". If we
refer to the demo the content of this field is supposed to be shown
under the status on the status page.

The problem was the information was not shown even if it was not empty.
So if we typed "We are a company ...", this was not shown, nowhere.
Technically the problem was because the partial "about-app" was not
included so the information was never shown.

To fix that I have included the partial "about-app" in the index.
Now if the field empty nothing is shown (the same as before) but if the
field is not empty so the information is shown under the status bar.

See: #3249

### DIFF
--- a/resources/views/index.blade.php
+++ b/resources/views/index.blade.php
@@ -3,6 +3,7 @@
 @section('content')
 @include('partials.modules.messages')
 @include('partials.modules.status')
+@include('partials.about-app')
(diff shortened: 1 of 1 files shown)


jbrooksuk:
Whoops! Looks like this was accidentally removed.

PR #3266, opened 2018-10-05
Revert schedule maintenance

thomasxd24:
commit of de172e5defa4ac226044e3066413487fdee49de0 have made the schedule maintenance disappear completely. 

This PR reverts the change. Please fix the change or revert back.


jbrooksuk:
Hmm, it looks like the change that broke this was wrong. Good spot.

PR #3273, opened 2018-10-11
Fixed rare case of division by 0

VeekeeFr:


VeekeeFr:
FYI: I encounter that issue when starting a CachetHQ container on a pre-existing populated database (don't know if that's the condition to reproduce it).

app/Http/Controllers/StatusPageController.php
```diff
@@ -74,7 +74,11 @@ public function showIndex()
                                       ->values();
 
             $numIncidentDays = count($allIncidentDays);
-            $numPages = round($numIncidentDays / $appIncidentDays);
+            if ($appIncidentDays > 0) {
```
jbrooksuk: I think that this can be shortened to:

```php
$numPages = max(round($numIncidentDays / $appIncidentDays), 1);
```

jbrooksuk: I think that this can be shortened to:

```php
$numPages = max(round($numIncidentDays / $appIncidentDays), 1);
```

VeekeeFr: Sorry, 0 division will still be stopping any further process.
But, using that change is working
            $numPages = round($numIncidentDays / max($appIncidentDays, 1));

jbrooksuk: Okay, that change looks better. Can you update please?

PR #3287, opened 2018-10-18
grep -rl "test\.com" . | xargs sed -i 's/test.com/example.com/g'

sideffect0:
From test.com to example.com


jbrooksuk:
@sideffect0 thanks for this. We will merge at some point, but we just need to co-ordinate updating everything so people can still login.

Issue #3244, opened 2018-09-13
Setup page requires a mail username


jcmcken:
The setup page seems to require a mail host login username (but doesn't require a password). I think the username and password should both be optional.

In my specific use case, I'm setting the mail host to ``127.0.0.1``, because my local host already has mail forwarding configured correctly. So I don't need any credentials to send mail

jbrooksuk:
We can make the username optional. It's just validation.

For the time being, update the `.env` file `MAIL_*` settings.

jbrooksuk:
Fixed in 6a53e21354ba004e0cebb133506e0c065eac639f

PR #3294, opened 2018-10-20
Set the mail username optional in setup

anthonybocci:
During the setup process email informations were asked:
  - Driver
  - Host
  - Username
  - Password

In some situations the username is not useful because the Cachet's
host may be configured to forward email to a server.

The problem is the username was required, so we had to set a username
and then update the .env file to remove it.

To fix this problem, the mail username has been set to optional in
the setup. So if someone needs a username it still can use this field,
and otherwise people can let it empty.

See: #3244


anthonybocci:
I've updated the SetupController, but I've to change because for now if `MAIL_USERNAME` is null the `Subscribe` button will never be viewable.

jbrooksuk:
@anthonybocci did we not change the subscriber test to simply be whether the setting is enabled or not?

anthonybocci:
@jbrooksuk I just checked, your right the test has been updated.
So, i've just udpated the file so the username is optional only for the sendmail driver.

anthonybocci:
@jbrooksuk I've removed the "WIP", you can check (because I'm not familiar with the Laravel validator) :)

PR #3297, opened 2018-10-23
Infrastructure As Code

alexstojda:
### Purpose
Essentially, I wanted a way to provision a CachetHQ docker image with components and component groups using a configuration file of sorts instead of having to create them one-by-one using the UI. Where I work, we have CachetHQ deployed in a Kubernetes cluster where most other services are file-driven to simplify the management of those services, especially in the case where something happens to the container and it has to be re-created. 

Having a file-driven approach allows for easy recovery from a container failure, and makes provisioning the Cachet instance with data really simple. 

This started out as a change to work in our setup, but may be useful for others who deploy Cachet in a Kube cluster, thus I created this PR!

### Approach
I create a DB seeder that updates tables in the database to match the configuration file. Ex, if a new component is added to the config, it will be added the next time the `php artisan db:seed` command is run. Same thing if an item is deleted from the config. 

The `php artisan db:seed` command would then be included as part of the [`entrypoint.sh`](https://github.com/CachetHQ/Docker/blob/master/entrypoint.sh) in the [CachetHQ Docker Image](https://github.com/CachetHQ/Docker/) so that it is run when the container starts. 

In a Kubernetes cluster, the configuration files would be included as part of the application's manifests, and their mounted path stored in the environment variable `FILE_DRIVEN_DIR` to tell CachetHQ where to look for the config files. By default, it will look in the `database/data` folder of the project. 

This "file-driven" behaviour can be of course toggled using the environment variable `FILE_DRIVEN` and setting it to either `true` or `false` 

---

I'm very receptive to comments or suggestions to make this PR better, so feel free to include them!


alexstojda:
Hey @GrahamCampbell @jbrooksuk, any updates or comments for this PR? I'd really like to get it merged :)

GrahamCampbell:
Hi Alex. Sorry for the delay here. I understand @jbrooksuk is planning to review this PR in the final week of November. :)

alexstojda:
Hey team, any updates? 

@jbrooksuk @GrahamCampbell

sebastien-prudhomme:
@alexstojda is there a public Docker image with your PR?

alexstojda:
@sebastien-prudhomme no, but when you docker compose [this repo](https://github.com/CachetHQ/Docker) you should be able to manually specify this PR version.

jbrooksuk:
Hello @alexstojda, apologies for the delay on reviewing this. 

I love the idea of these changes but I think that this PR would be better suited as a fork which is wrapped with your changes.

Couple of questions:

1. Would you like me to review the code nonetheless so that you can maintain a fork?
2. Why did you choose not to use a separate DB instance/cluster instead of flat-file (pure curiosity).

alexstojda:
Hey @jbrooksuk! 

Personally, I think with the way the changes have been implemented, it doesn't really warrant its own fork (plus i personally don't have the time to maintain a fork either :P 

The reason I chose to use a flat file is that said flat file can be included as part of a Kubernetes manifest. So when you deploy a manifest for a cachet node, the configuration is already included without having to spin up a second DB.

alexstojda:
@jbrooksuk @sebastien-prudhomme Any thoughts on my comments above?

FalkW:
Would love to see this merged but it seems there is no progress?

alexstojda:
@FalkW Sadly no progress on getting the code-owners to review :/

DAgostinoLucas:
So from now and afaik, there are no concrete ways to use this PR ? @alexstojda  Just searching a way to automate deployment of cachet, this [ansible playbook](https://github.com/JoanJetson/cachet-ansible) could be a beginning but I'm really worried about the status of CachetHQ project, it seems dead...

dasJ:
@DAgostinoLucas It's not perfect, but you could try this: https://github.com/stuvusIT/cachet

DAgostinoLucas:
@dasJ thanks for your reply and your role! I'm also looking to fully automate Cachet installation (without a setup page actually, user and components already created). Btw, I was also interested by this repo: https://github.com/ahstn/statuspage in order to inject SQL in a CachetHQ docker image :)

alexstojda:
Closing the PR since Cachet seems dead, and 2 years seems like a long enough time to wait :P

Issue #3210, opened 2018-08-06
Unnecessary requirement of external assets


telekem:
For offline development and for the sake of privacy (so that requests are only made to your own server), it should be possible to disable usage of external assets/cdns. Currently on a brand new install the home status page is making requests to cloudflare (for zxcvbn) and google fonts.

As an example where this issue has been resolved, gitea has an [offline mode](https://docs.gitea.io/en-us/config-cheat-sheet/) that disables usage of cdns for js libraries.

jbrooksuk:
There is a setting in the dashboard which disables external dependencies :)

telekem:
@jbrooksuk is this in a newer version? I'm on the v2.3.9 tag

jbrooksuk:
Sorry, I can’t remember the exact version. You’re a bit behind in the 2.3 release cycle though, so it’s worth upgrading.

PR #3315, opened 2018-11-06
Remove third-party password strength plugin

jbrooksuk:
Part of #3210 

---

I've signed up to a lot of services recently and haven't seen **anyone** using a password strength meter.


jbrooksuk:
@GrahamCampbell we need to look at the PHP 7.3 tests soon.

PR #3326, opened 2018-11-12
Let notification to the users be queue-able

lukepolo:
Currently my instance times out cause we have over 150+ subscribers. 

Adding this actually adds it to the queue.


jbrooksuk:
Good spot and a great fix, thank you1

PR #3328, opened 2018-11-14
Upgraded composer deps

GrahamCampbell:


composer.json
```diff
@@ -98,6 +104,9 @@
         ]
     },
     "config": {
+        "platform": {
+            "php": "7.1.3"
```
GrahamCampbell: This ensures that all dependencies are compatible with php 7.1.3, and not just the current system platform when generating the lock file.

composer.json
```diff
@@ -24,47 +24,53 @@
             "email": "joe@alt-three.com"
         }
     ],
+    "replace": {
+        "paragonie/random_compat": "*",
```
GrahamCampbell: Since Cachet requires PHP 7.1, we can prevent the installation of this package by saying we "replace" it. Similar for the mbstring polyfill, since we require the actual extension.

PR #3330, opened 2018-11-14
Fixed coverage runs

GrahamCampbell:
We need xdebug to be available to run coverage reports.


GrahamCampbell:
NB Still failing because of scrutinizer access issues.

PR #3333, opened 2018-11-19
Change subscribe page description to be more precise

bnahin:
"Subscribe to get the updates" can be "Subscribe to status changes and incident updates" to be less vague and more precise.


jbrooksuk:
This makes sense to me :)

PR #3342, opened 2018-11-25
Write port numbers explicitly to the .env

nstapelbroek:
Hey, 

When installing cachet for postgres and using the default port options it will fail the installation because the null value is not being handled nicely, resulting in a
`SQLSTATE[08006] [7] invalid port number: "sslmode=prefer"`.

This PR will explicitly configure port numbers for the mysql and pgsql drivers when you've selected them during configuration. The PR also adds support for finding & replacing "empty null values?" in the .env file like the following:

```
DB_PORT=
```

I'd love to know your thoughts on this. 
Have a great weekend 🎉


jbrooksuk:
Love it, as always, thanks @nstapelbroek!

PR #3343, opened 2018-11-25
Add empty array fallback when contributor fetching fails

nstapelbroek:
Hey,

Just saw a short hiccup in my internet connection and noticed that it crashed my cachet settings page. This PR will add a fallback value to the settings contributor page.


nstapelbroek:
Reading this again I actually dont like this implementation. I don't want failures in an api client to be handled in a controller. I'm closing this because some sort of Dto / Value object returned from the Credits integration class would make more sense. Having 0 contributors in the object would than be a valid edge-case.

jbrooksuk:
Agreed. This should be in the `Credits` class.

PR #3345, opened 2018-11-27
Fixed timezone of creation datetime for metric points

JordyvanDortmont:
I think this fixes a timezone issue that exists since commit 55002bb, which did not include the timezone for the creation date.

### DIFF
--- a/app/Models/MetricPoint.php
+++ b/app/Models/MetricPoint.php
@@ -120,7 +120,7 @@ class MetricPoint extends Model implements HasPresenter
         $timestamp = $createdAt->format('U');
         $timestamp = 30 * round($timestamp / 30);
 
-        $date = Carbon::createFromFormat('U', $timestamp)->toDateTimeString();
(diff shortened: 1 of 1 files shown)


jbrooksuk:
@JordyvanDortmont, thank you for this! The tests are failing though, do you have any thoughts?

JordyvanDortmont:
@jbrooksuk I thought about this for a while and looked through the code again. The `updated_at` timestamp of 18:23:09 seems correct, but it seems that the timezone offset is applied again when creating the `created_at` timestamp, because of the timezone header. I believe cachet-monitor [cachet-monitor](https://github.com/CastawayLabs/cachet-monitor) does not send the `Time-Zone` header and now I'm not sure why this fixes my use-case. I will look into this.

Do you also think the timezone offset is applied twice, and if so, is it always necessary to send the timezone header when you create a metric point with a timestamp?

jbrooksuk:
@JordyvanDortmont I'm not sure... The timezone header alone should be enough, I thought...

jbrooksuk:

Issue #1696, opened 2016-04-12
How install cachethq in subfolder ?


anouarjlidi:
when install cachet in subfolder example "http://localhost/cachet/" css and js error 404
what is the solution ?

GrahamCampbell:
We don't support subfolder installs because laravel doesn't properly support them.

GrahamCampbell:
You can get it to mostly work if you set the application url correctly in the .env file.

markperdue:
@GrahamCampbell do you have any suggestions on how to set it correctly? I am running into the same issue and have tried different variations of APP_URL that point to the external hosted url, internal localhost url including the subdirectory, etc and haven't had any luck.

gerhard-tinned:
I would need this too. How is "correct" in this case?

jbrooksuk:
@gerhard-tinned it really isn't supported, sorry.

iamjulians:
Solution:
Put it on a subdomain/domain, then user either:

a. 301/302 HTTP redirect from a subfolder to subdomain
or b. iframe to subdomain

gerhard-tinned:
This is **not a solution**, this is in best case a workaround for a badly designed web-application. 

A web-application that is not able to run from a subdirectory is actually a sign for bad web development practice from my point of view.

jbrooksuk:
> A web-application that is not able to run from a subdirectory is actually a sign for bad web development practice from my point of view.

Please could you explain your point of view?

bjoernffm:
Hey guys, actually laravel supports being in a subfolder pretty well.

My nginx configuration includes this part:

```
location /cachet {
    alias /var/www/Cachet/public;
    try_files $uri $uri/ @cachet;

    location ~ \.php$ {
        include snippets/fastcgi-php.conf;
        fastcgi_param SCRIPT_FILENAME $request_filename;
        fastcgi_pass unix:/run/php/php7.0-fpm.sock;
    }
}

location @cachet {
    rewrite /cachet/(.*)$ /cachet/index.php?/$1 last;
}
```

... and you're able to provide every laravel application in a subfolder (in this case /cachet).

From this point on, it's just a matter of "is the application programmed the way it supports being in a subfolder?".

Right now Cachet seems only to support this partly - I'll make some changes to provide a better support for both (normal and subfolder) and will start a pull request.

tvvignesh:
[...]

PR #3355, opened 2018-12-04
adding asset() helper to dist folder to respect APP_URL

jaredvacanti:
#1696 

`APP_URL` is used to adjust a root url in order to serve Cachet from a subdirectory (proxied by Nginx or Apache), however static assets `URI`s are not adjusted. This fix modifies these links to respect `APP_URL` by using the `asset()` helper around `mix()`.

### DIFF
--- a/resources/views/layout/clean.blade.php
+++ b/resources/views/layout/clean.blade.php
@@ -25,7 +25,7 @@
     @if($enableExternalDependencies)
     {{-- <link href="https://fonts.googleapis.com/css?family=Open+Sans:300,400,700&subset={{ $fontSubset }}" rel="stylesheet" type="text/css"> --}}
     @endif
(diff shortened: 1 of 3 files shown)


jaredvacanti:
Adjusting javascript to relative paths allows the application to function when served from a sub-directory. With an nginx configuration posted in the attached issue this seems to completely solve the subdirectory issue.

jbrooksuk:
This has broken some CSS?

image

GrahamCampbell:
Hmmm, I'm inclined to say that we don't support serving Cachet from a subdirectory.

jbrooksuk:
>Hmmm, I'm inclined to say that we don't support serving Cachet from a subdirectory.

@GrahamCampbell I would agree. It’s not official, but I’d be happy to link to a fork that does this.

PR #3357, opened 2018-12-07
Improve database performance by removing duplicated queries and using eager loading

AdrienPoupa:
This PR eliminates some unnecessary queries:

The `DashboardComposer` was executed as many times as there were dashboard views rendered, resulting in 5 additional queries for each execution when one was enough. To ensure that it was run only once, I used [this method](https://laracasts.com/discuss/channels/laravel/executing-a-view-composer-only-once).

Also, for some reason, the retrieval of the setting `enable_external_dependencies` was done using `setting` and not `Cache::`, thus a query to the database was done whenever an avatar was displayed.

In `IncidentController`, eager loading the user is useful because the dashboard.incident.index view queries `$incident->user` twice.

In ComponentController, eager loading the group is useful because it is queried twice per component. On my test instance that has 12 components, the dashboard/components goes from 35 queries, 5.3Mb and 60ms to 10 queries, 4.8Mb and 46ms

Finally, on the homepage, by adding eager loadings to `$allIncidents` in the `StatusPageController` and creating attributes that act like a cache, on a large database, I have reduced the number of queries from 539 to 118, the memory consumption from 18Mb to 8Mb and the loading time from 530ms to 240ms.

### DIFF
--- a/app/Composers/DashboardComposer.php
+++ b/app/Composers/DashboardComposer.php
@@ -26,6 +26,41 @@ use Illuminate\Contracts\View\View;
  */
 class DashboardComposer
 {
+    private $componentCount;
+
+    private $incidentCount;
+
+    private $incidentTemplateCount;
+
+    private $scheduleCount;
+
+    private $subscriberCount;
+
+    /**
+     * DashboardComposer constructor
+     */
+    public function __construct() {
+        if (is_null($this->componentCount)) {
+            $this->componentCount = Component::count();
+        }
+
+        if (is_null($this->incidentCount)) {
+            $this->incidentCount = Incident::count();
+        }
+
+        if (is_null($this->incidentTemplateCount)) {
+            $this->incidentTemplateCount = IncidentTemplate::count();
+        }
+
+        if (is_null($this->scheduleCount)) {
+            $this->scheduleCount = Schedule::count();
+        }
+
+        if (is_null($this->subscriberCount)) {
+            $this->subscriberCount = Subscriber::isVerified()->count();
+        }
+    }
+
     /**
      * Bind data to the view.
      *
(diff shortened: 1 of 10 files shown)


app/Composers/DashboardComposer.php
```diff
@@ -26,6 +26,42 @@
...
+    private $incidentCount;
+
+    private $incidentTemplateCount;
+
+    private $scheduleCount;
+
+    private $subscriberCount;
+
+    /**
+     * DashboardComposer constructor.
```
jbrooksuk: Typically we would comment as:

```
Create a new dashboard composer instance.
```

Also add a `@return void` docblock.

app/Composers/DashboardComposer.php
```diff
@@ -26,6 +26,42 @@
  */
 class DashboardComposer
 {
+    private $componentCount;
```
jbrooksuk: These variables are missing docblocks. For consistency, we also use `protected`.

app/Presenters/ComponentGroupPresenter.php
```diff
@@ -20,14 +20,16 @@ class ComponentGroupPresenter extends BasePresenter implements Arrayable
 {
     use TimestampsTrait;
 
+    protected $enabledComponentsLowest = false;
```
jbrooksuk: Add docblock.

app/Presenters/ComponentGroupPresenter.php
```diff
@@ -51,11 +53,25 @@ public function lowest_human_status()
...
     }
 
+    /**
+     * Return the enabled components from the wrapped object, and cache it if need be.
+     *
+     * @return bool
+     */
+    public function enabled_components_lowest()
+    {
+        if (is_bool($this->enabledComponentsLowest)) {
```
jbrooksuk: Why are we checking `is_bool` if the variable is declared as `false` by default?

app/Presenters/IncidentPresenter.php
```diff
@@ -29,6 +29,8 @@ class IncidentPresenter extends BasePresenter implements Arrayable
      */
     protected $dates;
 
+    protected $latest = false;
```
jbrooksuk: Add docblock.

app/Presenters/IncidentPresenter.php
```diff
@@ -248,9 +250,11 @@ public function latest_icon()
      */
     public function latest()
     {
-        if ($update = $this->wrappedObject->updates()->orderBy('created_at', 'desc')->first()) {
-            return $update;
+        if (is_bool($this->latest)) {
```
jbrooksuk: I think we can just say `if ($this->latest)`.

AdrienPoupa:
Hi,

Thanks for the review. I chose to use a boolean because by default, if a var is not declared in PHP, it's a null. If there is no hit in the database, Eloquent returns a null as well. Thus, we need a way to know if we queried the database at least once; if we were checking a null and Eloquent returned a null we would be checking over and over again...

By using is_bool, we know for sure that for the current request, if the attribute is a boolean, the query was not executed yet. After the execution, it can be set either to its actual value, or null if nothing was found in the database.

jbrooksuk:
Awesome, thank you!

PR #3363, opened 2018-12-16
Resolve #3288

### LINKED ISSUE #3288: Components get disabled on updates
Hello !

I have a weird issue with components getting disabled when they are updated via API.
Everytime I update a component with the PUT Method directly on `https://cachet/component/:id` and do not specify `enabled` in the body, the component gets disabled.

BR,
Pascal

TakeMeNL:
If Binput::get('enabled') is not defined this will return NULL which resulted in a false.
Also updated tests


app/Http/Controllers/Api/ComponentController.php
```diff
@@ -124,7 +124,7 @@ public function update(Component $component)
                 Binput::get('link'),
                 Binput::get('order'),
                 Binput::get('group_id'),
-                (bool) Binput::get('enabled'),
+                Binput::get('enabled') !== null ? (bool) Binput::get('enabled') : $component->enabled,
```
jbrooksuk: I think that this can be:

```php
(bool) Binput::get('enabled', $component->enabled)
```

TakeMeNL: Yes indeed.. That would be better, updated in new commit/squash

PR #3377, opened 2018-12-28
Updated doctrine and emoji packages

### LINKED ISSUE #3373: Upgrade doctrine/dbal
To be done after upgrading to Laravel 5.7.

GrahamCampbell:
Closes #3373.


config/emoji.php
```diff
@@ -23,4 +23,38 @@
...
+    |--------------------------------------------------------------------------
+    | Cache Connection
+    |--------------------------------------------------------------------------
+    |
+    | Here you may tell us what cache connection to use. To use the default,
+    | set this to null.
+    |
+    */
+
+    'connection' => null,
```
jbrooksuk: Shall we use `CACHE_DRIVER` on this?

GrahamCampbell: Nah. Setting to null means when we call `->store()` on the cache driver, we really do get the default. if we were to use env in the config file, we're making assumptions about the way the user has setup their cache config.

PR #3381, opened 2018-12-29
Add newline before timezone line

TakeMeNL:
Add newline before timezone line on index page in the footer.
It just looks a bit nicer because the div is only max pixels wide.


jbrooksuk:
I'm going to close this as the new design should eliminate this issue.

You can find more about the progress on the new design at [CachetHQ/Phoebe](https://github.com/CachetHQ/Phoebe).

PR #3382, opened 2018-12-30
Upgraded dependencies following XSS vulnerability discovery

GrahamCampbell:
Includes the fix to https://github.com/thephpleague/commonmark/issues/337.


GrahamCampbell:
Note that Cachet 2.3 is using CommonMark 0.13.x, so is unaffected by this vulnerability.

PR #3402, opened 2019-01-03
Authenticate with REMOTE_USER

### LINKED ISSUE #3243: Header-based authentication (e.g. REMOTE_USER)
Are there plans to support fronting Cachet with an authenticating web proxy? For example, a frontend web proxy authenticates the user, and passes the authenticated user name to Cachet via an HTTP header or environment variable, which Cachet can then trust implicitly.

This type of configuration allows users to use arbitrary authentication methods (including LDAP, CAS, SAML, OAuth, OIDC, etc.) without needing to add this functionality directly into Cachet's code base.

This possibility was mentioned in a few other issues that I saw, but it didn't look like it was addressed. Our team uses this pattern with a lot of applications that don't natively support things like OAuth and SAML authentication

jbrooksuk:
Closes #3243 

---

Right, so I've not written any tests for this yet, but I want to know if this is the right way of doing things since I've not got a method of testing. 

Ping @jcmcken


GrahamCampbell:
This doesn't seem very secure? All you need is someone's email, then you can login as them?

jbrooksuk:
> This doesn't seem very secure? All you need is someone's email, then you can login as them?

Which is what's leading me to believe I've done this wrong?

GrahamCampbell:
I guess the point of this header is that it can only be set by the internal proxy. This middleware should definitely not be enabled in Cachet by default, and should be turned on only by those who are using Cachet behind such a proxy.

joecohens:
What if we use the apiToken with this header as well instead of the email if the header is truly needed?

jcmcken:
There are two general ways to do this:

* Using a header
* Using the runtime environment (an environment variable)

The first option requires the end user to set up the authenticating web server in a particular way. If the proxy and the app are co-located on the same machine, the app must run on 127.0.0.1 at a minimum. If not, more work is required (e.g. you must authenticate the proxy to Cachet or use layer 4 network protections).

The second option requires that the proxy and app are co-located and is generally more secure. Clients have no ability to set environment variables on the runtime process. Only the web server / runtime launcher can do this. Although it's Python, you can read [the Django documentation](https://docs.djangoproject.com/en/2.1/howto/auth-remote-user/) which explains some of these issues (which are of course not language specific).

jbrooksuk:
@jcmcken are you able to test this? :)

jbrooksuk:
Ping @jcmcken!

jbrooksuk:
@jcmcken I've fixed the conflicts, if you can please give this another go?

jbrooksuk:
I'm going to merge this so we can move on with other bits. We can always look at creating accounts later on.

PR #3481, opened 2019-02-19
API cache control

### LINKED ISSUE #3479: Caching / rate limiting for API
I just had an issue where my server got a huge number of hits to the `/api/v1/status` endpoint, which basically killed my PHP-FPM server (it hit the maximum number of workers, preventing people from connecting to any other sites in the same pool).

Does Cachet cache the response to this API call? It would be good to cache its response (even if only for a short period of time, like 30 seconds), so that a large number of hits don't cause any perf issues. If a lot of hits come in at the same time, they could all receive the cached response rather than having to recompute it + hit the DB.

jbrooksuk:
Closes #3479 

---

This adds the `Cache-Control` middleware to the status endpoint of a max-age of now + 30 seconds. Before this gets merged, the added time needs to be variable (with a default of 30?) so that other endpoints can use it as needed.

Ping @Daniel15 on this one.

### DIFF
--- a/app/Http/Middleware/CacheControl.php
+++ b/app/Http/Middleware/CacheControl.php
@@ -19,8 +19,9 @@ class CacheControl
     /**
      * Handle an incoming request.
      *
-     * @param  \Illuminate\Http\Request  $request
-     * @param  \Closure  $next
+     * @param \Illuminate\Http\Request $request
+     * @param \Closure                 $next
+     *
      * @return mixed
      */
     public function handle(Request $request, Closure $next)
--- a/app/Http/Middleware/CacheControl.php
+++ b/app/Http/Middleware/CacheControl.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ * This file is part of Cachet.
+ *
+ * (c) Alt Three Services Limited
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace CachetHQ\Cachet\Http\Middleware;
+
+use Closure;
+use Illuminate\Http\Request;
+
+class CacheControl
+{
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @return mixed
+     */
+    public function handle(Request $request, Closure $next)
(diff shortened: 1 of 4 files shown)


app/Http/Middleware/CacheControl.php
```diff
@@ -0,0 +1,37 @@
...
+     *
+     * @return mixed
+     */
+    public function handle(Request $request, Closure $next)
+    {
+        $response = $next($request);
+
+        $maxAge = time() + 30;
+
+        $response->header('Cache-Control', 'public,max-age='.$maxAge);
```
Daniel15: I wonder if it's also worth sending a `Last-Modified` date, and handling requests with an `If-Modified-Since` header (return `304 Not Modified` if the content hasn't changed). Perhaps that's overkill though, and it'd require a bit more work, as ideally you'd want to bail out of most of the work if you can quickly determine that there have been no changes since the specified date.

tests/Api/GeneralTest.php
```diff
@@ -42,4 +44,36 @@ public function test_non_acceptable_content_type()
...
+                         'status'  => 'success',
+                         'message' => 'System operational',
+                     ],
+                 ]);
+    }
+
+    public function test_can_get_system_status_not_success()
+    {
+        factory(Component::class)->create([
+            'status' => 3,
```
Daniel15: Is there a constant you can use here, instead of the magic number `3`?

app/Http/Middleware/CacheControl.php
jbrooksuk: `Last-Modified` could be worked out by the latest component or incident that affected the status, yes.

tests/Api/GeneralTest.php
jbrooksuk: Not currently. There will be added difficulty in this once we allow people to configure their own status values as well.

app/Http/Middleware/CacheControl.php
jbrooksuk: @Daniel15 actually, what I said above is a bit more complicated as we don't store the global state anywhere, so we'd have to re-fetch the components in the middleware. As you say, this may be overkill anyway, so I'm going to merge this as-is.

PR #3520, opened 2019-03-14
Show current ongoing maintenance, not just future ones.

blontrock:
Without this, the maintenance on the dashboard disappears once it starts. It should only disappear when the maintenance is completed.


jbrooksuk:
I don't think this is needed anymore.

Issue #3314, opened 2018-11-06
Add incident update description in notifications sent by email


labolanimichele:
At the current status of 2.4, the email notifications sent out when an incident is updated do not contain the "Message" one enters in the UI for the incident update.

What I enter in the UI is:
![image](https://user-images.githubusercontent.com/22618797/48074953-5f263c80-e1e2-11e8-8fc8-936eea799a9a.png)

-----------------
Here what I receive right now (w)

**Test Incident 1 was updated to Identified**
Test Incident 1 was updated

Thanks,
Status Page

Unsubscribe — Manage subscription

-----------------

I would recommend the following changes:
- add the update message
- add information on the status of the component (currently only the status of the incident is reported)
- add an hyperlink easy to find to the incident page on the UI
- remove the repetition of the title ("Test Incident 1 was updated")

jbrooksuk:
Makes sense to me!

jbrooksuk:
Please feel free to take this issue up as well :)

MinThaMie:
This is fixed right? Otherwise I could take a look at the stuff that is still left over, but I'm not sure what that is atm.

jbrooksuk:
We need to get some tests in that prove that notifications are sent etc.

jbrooksuk:
Thank you for your input on Cachet 2.x. We are shifting our attention and resources to Cachet 3.x and will no longer be supporting the 2.x version. If your feedback or issue is relevant to the 3.x series, we encourage you to engage with the new branch.

For more information on the Cachet rebuild and our plans for 3.x, you can read the announcement [here](https://github.com/CachetHQ/Cachet/discussions/4342).

We appreciate your understanding and look forward to your contributions to the new version.

PR #3540, opened 2019-04-10
Include incident update message in mail notification

jzinkweg:
Mostly revolves #3314.

As component updates are notified separately and they are optional in the incident I haven't included that requested item.


jbrooksuk:
Awesome, thanks!

Issue #3559, opened 2019-04-24
Completed_at date maintenance is should be required


MinThaMie:
The `completed_at` date  for maintenance is not required, but this will break the showing of the maintenance message on the dashboard. 

There is also no validation on the `completed_at` being after the `scheduled_at`, this should probably also be added.

Pimorez:
I think this issue isn't completely correct. If you fill in the 'completed_at' field, the maintenance will not be shown rather than what you're saying that it is not shown when not filled in.

jbrooksuk:
Also, the completed at is the time that the maintenance completed, not the due date for completion. Maybe we could separate this information though.

jbrooksuk:
Thank you for your input on Cachet 2.x. We are shifting our attention and resources to Cachet 3.x and will no longer be supporting the 2.x version. If your feedback or issue is relevant to the 3.x series, we encourage you to engage with the new branch.

For more information on the Cachet rebuild and our plans for 3.x, you can read the announcement [here](https://github.com/CachetHQ/Cachet/discussions/4342).

We appreciate your understanding and look forward to your contributions to the new version.

PR #3560, opened 2019-04-24
Make enddate required and validated

MinThaMie:
Fixes https://github.com/CachetHQ/Cachet/issues/3559 

It makes the end date of maintenance required and it also validates if the end date is after the begin date.

### DIFF
--- a/app/Bus/Commands/Schedule/CreateScheduleCommand.php
+++ b/app/Bus/Commands/Schedule/CreateScheduleCommand.php
@@ -70,7 +70,7 @@ final class CreateScheduleCommand
         'message'      => 'nullable|string',
         'status'       => 'required|int|min:0|max:2',
         'scheduled_at' => 'required|string',
-        'completed_at' => 'nullable|string',
+        'completed_at' => 'required|string',
         'components'   => 'nullable|array',
     ];
 
--- a/app/Models/Schedule.php
+++ b/app/Models/Schedule.php
@@ -103,6 +103,7 @@ class Schedule extends Model implements HasPresenter
         'name'         => 'required|string',
         'message'      => 'nullable|string',
         'status'       => 'required|int|between:0,2',
+        'completed_at' => 'after_or_equal:scheduled_at'
     ];
 
     /**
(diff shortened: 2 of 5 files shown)


app/Bus/Commands/Schedule/CreateScheduleCommand.php
```diff
@@ -70,7 +70,7 @@ final class CreateScheduleCommand
         'message'      => 'nullable|string',
         'status'       => 'required|int|min:0|max:2',
         'scheduled_at' => 'required|string',
-        'completed_at' => 'nullable|string',
+        'completed_at' => 'required|string',
```
jbrooksuk: I don't think this makes sense. Maintenance could (and should) be scheduled ahead of time, so the completed time can't be known.

app/Models/Schedule.php
```diff
@@ -103,6 +103,7 @@ class Schedule extends Model implements HasPresenter
         'name'         => 'required|string',
         'message'      => 'nullable|string',
         'status'       => 'required|int|between:0,2',
+        'completed_at' => 'after_or_equal:scheduled_at',
```
jbrooksuk: I didn't know about `after_or_equal`, that's cool! Should probably be `nullable` though (see above comment).

app/Bus/Commands/Schedule/CreateScheduleCommand.php
MinThaMie: But this PR is also cool and would break if it is `nullable`: https://github.com/CachetHQ/Cachet/pull/3520 What is your opinion on that?

app/Models/Schedule.php
MinThaMie: Cool :) Will keep this, and based on our discussion above I'll add that

app/Bus/Commands/Schedule/CreateScheduleCommand.php
jbrooksuk: I think we should revisit that bit of the system then, because it clearly isn't making sense :)

Wouter0100: Personally we use `completed_at` as a maximum completion time for the maintenance window.

MinThaMie:
I'm no longer using Cachet so I'll close the PR

PR #3604, opened 2019-05-17
Ignore `completed_at` time for schedules

Wouter0100:
We always set a `completed_at` time for schedules, as a maximum time when the schedule should be finished. With these where-statements, schedules with `completed_at` time aren't displayed.

### DIFF
--- a/app/Models/Schedule.php
+++ b/app/Models/Schedule.php
@@ -152,9 +152,7 @@ class Schedule extends Model implements HasPresenter
      */
     public function scopeUncompleted(Builder $query)
     {
-        return $query->whereIn('status', [self::UPCOMING, self::IN_PROGRESS])->where(function (Builder $query) {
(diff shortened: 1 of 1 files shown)


jbrooksuk:
@Wouter0100 this feels like maybe bespoke logic? I'm really unsure on how schedules should work for the general.

Wouter0100:
Me neither, but if I'm correct - this field was required in the past. Because of that, we started using as an estimation on when the schedule is about to end. Not sure what more information is needed? :-P

Wouter0100:
Reopening as this is still an on-going thing. We manually remove this code each time we update :-).

The thought for us is that we have a maintenance window from x to y. Where y is an end date/time estimation, so customers know when to expect the maintenance to end.

jbrooksuk:

Issue #3378, opened 2018-12-28
Upgrade pragmarx/google2fa


GrahamCampbell:

jbrooksuk:
Closed by https://github.com/CachetHQ/Cachet/pull/3619

PR #3619, opened 2019-05-24
upgraded google2fa (0.7->5.0)

mxss:
#3378


jbrooksuk:
Hey, thanks for opening this @mxss. I'm unsure how I feel about requiring Imagemagick to be installed as well. 

Is there an alternative?

mxss:
@jbrooksuk in recent versions of used package (pragmarx/google2fa) image generation is disabled that why I had to install additional package Bacon/QRCode which requires imagick

You can check package readme for more details:
https://github.com/antonioribeiro/google2fa

there's a few alternatives to that package but I'm not sure yet if we can get rid of imagick, I'll take a look

jbrooksuk:
Thanks @mxss!

GD is much easier to install and is typical of most installations that I've come across, whereas I find Imagemagick less common.

jbrooksuk:
It looks like https://github.com/chillerlan/php-qrcode has SVG support and doesn't list GD or Imagemagick as a requirement in `composer.json`.

mxss:
@jbrooksuk done

jbrooksuk:
Nice! Could you rebase this please and we'll merge it in?

Issue #3618, opened 2019-05-23
404 on resources when custom path


anthosz:
Hello,

Following issue #2977 and a request to slack. I cannot install cachet with a custom path.

Example: http://127.0.0.1/cachet
APP URL is the same in all case. 

If I add RewriteBase, only page seems to works but 127.0.0.1/cachet/setup/ cannot load ressources (like js/css...etc). Indeed, the browser try to get resources from 127.0.0.1/build/... instead 127.0.0.1/cachet/build/...

If I don't add RewriteBase, I have 404 for all pages but ressources links are ok..

Versions:
Apache 2.4
Php 7.2
Cachet 2.3.18

Best regards,

anthosz:
I just find public/build/rev-manifest.json but it's not absolute path.

The goal is to find where is generated "/build" & to force the root path before..

anthonybocci:
Hello,

Thank you for coming from Slack to here. I'll try to illustrate the issue, and check in the code what we have to do.

Firstly, to illustrate your problem (if I correctly understand it), here is a screenshot of the setup page.

![cachet-setup-no-rewrite-base](https://user-images.githubusercontent.com/7780244/58348087-e629c200-7e5f-11e9-9b19-51e0da6267fc.png)

On my fresh install the URL is `http://localhost/cachet/Cachet/public/index.php/setup`. I've set no virtual host in Apache, I use the default one here.  
The problem on this page is some requests are not using the `APP_BASE`  environment variable as a base path. An example of request is `/dist/css/all-2812406e36.css`, the path is given as absolute and the environment variable is not used.

anthosz:
Thank you @anthonybocci ,

This is exactly the issue that I have.

For now I use this workaround but it's pretty dirty:
>   AddOutputFilterByType SUBSTITUTE text/html text/css application/javascript
>   Substitute "s@/(auth|setup|build|img)@/cachet/$1@i"

After that, once installed & logged, I have an issue with the root page (but seems not related to this topic):
> [2019-05-24 07:33:35] dev.ERROR: ErrorException: count(): Parameter must be an array or an object that implements Countable in /cachet/vendor/laravel/framework/src/Illuminate/Database/Eloquent/
Builder.php:1161


Best regards,

anthonybocci:
@anthosz thank you for sharing your workaround. I've understood the problem, I just need to fix one thing and it will be ok.

[...]

PR #3621, opened 2019-05-24
Use APP_URL env var in assets URI

anthonybocci:
Cachet may be installed at the root of its vhost or in a subdirectory,
for example in "/" or under "/status".

The URI we found pointing to some assets were usually not prefixed with
the APP_URL value, so if Cachet was installed under "/status" the asset
URI pointed to the server root like "/my-asset.js".
It is a problem because that means the behaviour is broken in this case.
The problem was present from the setup, it was not possible to fill the
setup since the path to the scripts and CSS weres wrong.

The content of the APP_URL environment variable is now read and used
in the assets URI.

See: CachetHQ/Cachet#3618


jbrooksuk:
I believe we should be using the `asset` helper, instead :)

anthonybocci:
I'd forgotten this helper in Laravel. You can close this PR so, I'll check that and do it using `asset`. Sorry :)

PR #3633, opened 2019-06-01
Demo link should point to demo dashboard

ajaissle:


ajaissle:
Nvm, dev is the real demo

PR #3645, opened 2019-06-08
Allow Cachet setup not to be at server's root

anthonybocci:
Cachet may be installed at the root of its vhost or in a subdirectory,
for example in "/" or under "/status".

The URI we found pointing to some assets were usually not prefixed with
the path, so if Cachet was installed under "/status" the asset
URI pointed to the server root like "/my-asset.js".
It is a problem because that means the behaviour is broken in this case.
The problem was present from the setup, it was not possible to fill the
setup since the path to the scripts and CSS weres wrong.

The "asset" helper is now used and resolves the URI.

See: CachetHQ#3618


jbrooksuk:
Thanks for backporting this to 2.3 too.

PR #3646, opened 2019-06-08
Allow Cachet to be setup not from the server's root

anthonybocci:
The `asset()` helper was missing at some places so it was not possible
to install it in an other place than the server's root (/).

The paths have been fixed and now use the `asset()` helper. Also,
The steps process takes care of the current path.

See: #3618


anthosz:
@anthonybocci, I confirm that css/js/images are now OK (connected or anonymous) but the fonts path are not fixed: /fonts/ionicons.woff and /fonts/ionicons.ttf doesn't use custom path

Thank you!

agentcobra:
link with #3645?

anthonybocci:
@agentcobra yes it's related to #3645. Here I target 2.4 branch, the other PR targets 2.3.

anthonybocci:
Hi @anthosz, I checked by loading the setup page, but I don't find any 404 error in my browser. Could you explain me a bit more?
Here is my browser's console when I load `/setup`:

cachet-setup

anthosz:
Hi @anthonybocci

Maybe the fonts are not used in setup page, do you have the 404 after setup? (status and dashboard page)

anthonybocci:
@anthosz Ok I didn't understood. I'd tested only the setup part. I don't know where the font is included, probably somewhere from the CSS. I'll have  a look, thanks for the explanation :)

jbrooksuk:
Yeah, the fonts are loaded from CSS itself. I'd quite like to remove the icon font and use SVG at some point anyway.

I know @ojdon is working on implementing the redesign.

anthonybocci:
So is the PR mergeable @jbrooksuk? I don't touch the CSS since the design is under work.

jbrooksuk:
This seems to work locally, thanks @anthonybocci!

PR #3669, opened 2019-06-27
fixing .htaccess

cubex-hro:
Fixed a bug that caused apache2 only served _Cachet_ from **.../index.php/page**
e.g. **example.com/index.php/dashboard**


jbrooksuk:
Cachet isn't designed to be served from `index.php` - what reason would you be using this?

The `.htaccess` file is the one provided by Laravel at https://github.com/laravel/laravel/blob/master/public/.htaccess

PR #3670, opened 2019-06-29
Change app:install reference to cachet:install

chrisforrence:
It looked like there were a few issues about app:install not working; this MR updates the documentation to refer to the correct command.


chrisforrence:
Also made the suggested changes on the documentation page

jbrooksuk:
Thanks!

PR #3706, opened 2019-07-12
Fix timeout with large number of subscribers

hillale:
We have Cachet with 1499 subscribers, and this page was timing out.  We troubleshot the issue and found this code running thousands of mysql queries and hitting the 60 second nginx timeout.

Where the template hits this inside a loop, triggering many queries:
https://github.com/CachetHQ/Cachet/blob/2.4/resources/views/dashboard/subscribers/index.blade.php#L43

Helpful page on optimizing code for ORM utilization:
https://laravel-news.com/eloquent-eager-loading


jbrooksuk:
Thank you, I’m sure there are a few instances where we can make use of eager loading.

PR #3710, opened 2019-07-15
Improved fetching of settings

jbrooksuk:
Improve fetching of settings by caching into an array once


app/Settings/Repository.php
```diff
@@ -112,9 +112,15 @@ public function set($name, $value)
      */
     public function get($name, $default = null)
     {
+        static $settings = [];
```
GrahamCampbell: This will incorrectly cache settings.

PR #3711, opened 2019-07-15
App banner is now an uploaded file

### LINKED ISSUE #3709: Add support for vector images as custom logo
Under the 'Theme' settings tab you can upload images. It would be great if vector images can be added so it can scale without causing artifacts.

image

jbrooksuk:
Fixes #3709

---

This requires the `php artisan storage:link` command to be ran so that a banner image can be uploaded instead of stored in the DB. It also allows us to upload vector images.

Ping @Pimorez.


jbrooksuk:
@Pimorez are you happy with how this works?

Pimorez:
The files are correctly uploaded to `/storage/public/uploads` but are not being displayed.
Ownership is set to the webserver user, permissions are 0644.

It looks like it is displaying an empty image as if it's having issues reading the file. Am I missing something? (tried with both .svg and .png)

jbrooksuk:
Hmm, I’m unsure why this would be the case. If it can upload it, it should read it. I wonder if your web server is incorrectly serving the images?

jzinkweg:
This would make it harder to run Cachet in a docker container, as it now requires persistent storage, having base64 as an option would be good.

Issue #3046, opened 2018-05-15
[Enhancement] Better user accountability


RyoshiKayo:
- Ability to add user who created/updated/modified incident to email notification
- Add `- Updated by {username} at {time}` to incidents.

jbrooksuk:
👍 

Technically, we do know who updated an incident through the `actions` table, we just need to pull this information back.

jbrooksuk:
Thank you for your input on Cachet 2.x. We are shifting our attention and resources to Cachet 3.x and will no longer be supporting the 2.x version. If your feedback or issue is relevant to the 3.x series, we encourage you to engage with the new branch.

For more information on the Cachet rebuild and our plans for 3.x, you can read the announcement [here](https://github.com/CachetHQ/Cachet/discussions/4342).

We appreciate your understanding and look forward to your contributions to the new version.

PR #3809, opened 2019-10-02
Add username to the dashboard of who logged an incident or any incident updates.

Data-Kiss:
In relation to #3046 I've added the username of the person who logged an incident or an incident update on the dashboard.

I've only made the change in the English translation. I've noticed that viewing in other languages it still says "**by** {username}" - hopefully this is okay as I see some other languages aren't fully translated anyway.

I haven't added the name of the person to the notifications that go out as originally specified. I was wondering if that needed more thought? My only thinking is you may not want end users to know who posted it so they don't badger individuals in a company about an outage. I think it's good internally (on the dashboard) for accountability, but think end users perhaps may not need to know who. It may be better to make an option somewhere "Include username in notification emails" so the owner can decide.

Let me know if anything I've said doesn't make sense, or requires amending.

### DIFF
--- a/resources/lang/en/cachet.php
+++ b/resources/lang/en/cachet.php
@@ -34,7 +34,7 @@ return [
         'stickied'     => 'Stickied Incidents',
         'scheduled'    => 'Maintenance',
         'scheduled_at' => ', scheduled :timestamp',
-        'posted'       => 'Posted :timestamp',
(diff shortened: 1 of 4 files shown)


thewillwilson:
This doesn't work on v2.4, it just displays 'Posted 6 minutes ago by :username'

Issue #3566, opened 2019-04-26
Choose template when updating an incident


MinThaMie:
When you create an incident you can choose a template, but when you create an update for the incident you cannot choose one. It would be nice if this would be possible.

jbrooksuk:
Agreed! We can add this to the list.

jbrooksuk:
Thank you for your input on Cachet 2.x. We are shifting our attention and resources to Cachet 3.x and will no longer be supporting the 2.x version. If your feedback or issue is relevant to the 3.x series, we encourage you to engage with the new branch.

For more information on the Cachet rebuild and our plans for 3.x, you can read the announcement [here](https://github.com/CachetHQ/Cachet/discussions/4342).

We appreciate your understanding and look forward to your contributions to the new version.

PR #3815, opened 2019-10-09
Add the ability to select a template when adding an incident.

Data-Kiss:
Add the ability to select a template when adding an incident, as requested in #3566.


Data-Kiss:
Note that the entire section in resources/assets/js/cachet.js was commented out for some reason. I uncommented it to make this build, so if anyone knows why it needs to be commented out let me know. All seems to be fine with it included.

PR #3823, opened 2019-10-18
Generate a signed url for the verify redirect

### LINKED ISSUE #3714: 2.4 - rev c74193e - 403 Invalid signture when we click on the url to verify the email
Hello,

With the last release of Cachet (2.4), we have this 403 when we try to verify the email subscription:
> 
> Error 403
> Invalid signature.
> Forbidden
> What does this mean?
> 
> Something went wrong on our servers while we were processing your request. Invalid signature. This occurrence has been logged, and a highly trained team of monkeys has been dispatched to deal with your problem. We're really sorry about this, and will work hard to get this resolved as soon as possible.

Moreover, the email is well in "verified" mode in the subscribers tab.

If I try to rollback to 8f91f6d9 (23 june), works.

craigballinger:
This should resolve #3714. It appears what is happening is that both verify and manage are now signed routes, but when redirecting to manage after verifying the new subscriber, the redirect url was not signed and the middleware was generating a 403.

### DIFF
--- a/app/Http/Controllers/SubscribeController.php
+++ b/app/Http/Controllers/SubscribeController.php
@@ -28,10 +28,10 @@ use Illuminate\Contracts\Auth\Guard;
 use Illuminate\Contracts\Config\Repository;
 use Illuminate\Routing\Controller;
 use Illuminate\Support\Facades\Config;
+use Illuminate\Support\Facades\URL;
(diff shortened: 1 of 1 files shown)


djanand2000:
@jbrooksuk  Could you please let me know how to resolve the issue if we are using the Docker version of cachet.
Thanks.

Issue #3626, opened 2019-05-29
2.4: Maintenance doesn't works


anthosz:
Hello,

In 2.4 (last commit as this day), maintenance seems doesn't works.

I try to insert a maintenance event between 28 & 31 May 2019 with differents status and no maintenance are showed on status page.

- Content of _schedules_ table:

> mysql> select * from schedules;
> | id | name | message | status | scheduled_at        | completed_at        | created_at          | updated_at          | deleted_at |
> |  1 | test | test    |      0 | 2019-05-28 11:56:00 | 2019-05-31 21:30:00 | 2019-05-29 09:22:32 | 2019-05-29 09:56:51 | NULL       

Best regards,

pehlert:
Not sure if it helps, but I can confirm this on a fresh installation

ArthurKnoep:
Same problem here, also when I create the maintenance from the dashboard, it end on a 500 error

ArthurKnoep:
Why have you closed the issue, have you fixed your instance ?

anthosz:
@arthurknoep, oups bad manipulation on my smartphone, now open ^^ thx for notification

lunarthegrey:
Same issue, can't create any maintenance.

jbrooksuk:
I'm going to investigate this now.

jbrooksuk:
This is weird, I've just created maintenance without issue:

![image](https://user-images.githubusercontent.com/246103/59973156-1fef0500-9593-11e9-9b21-3191d16ef3bc.png)

Can someone please try this again and then provide me with your error logs?

lunarthegrey:
@jbrooksuk when trying to create a maintenance I get this 500 error page
![Screenshot_2019-06-23 Error 500](https://user-images.githubusercontent.com/8747426/59981469-37e17b80-95c9-11e9-9056-528ce2238f97.png)

When I run `docker-compose logs` all I see is the GET and POST requests.

```
cachet_1    | 172.24.0.1 - - [23/Jun/2019:20:09:04 +0000] "GET /dashboard/schedule/create HTTP/1.0" 200 9810 "https://status.domain.com/dashboard/schedule" "Mozilla/5.0 (X11; Linux x86_64; rv:67.0) Gecko/20100101 Firefox/67.0" "xx.xx.xx.xx"
[...]

PR #3890, opened 2019-12-12
Remove unused "reserved" column from jobs table

jzinkweg:
Fixes "null value" error as mentioned in #3626:
SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "reserved" violates not-null constraint

This mirrors the removal of the column in illuminate/queue:
 illuminate/queue@061851b#diff-fd99f8447bff90123d06344bfe84dd89

### DIFF
--- a/database/migrations/2019_12_12_131400_AlterJobsDropReserved.php
+++ b/database/migrations/2019_12_12_131400_AlterJobsDropReserved.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of Cachet.
+ *
+ * (c) Alt Three Services Limited
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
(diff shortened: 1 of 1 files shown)


jzinkweg:
I made the rollback restore the original broken situation, but the CI tests don't like it.
Alternatively the rollback could create a nullable column, this is not identical but functionally identical?

langemeijer:
Thank you @jzinkweg for your solution. Adding the file to my cachet install then running `php artisan migrate` solved this problem for me.

venkz:
why can't we just merge this change?

jzinkweg:
That would be up to @jbrooksuk .

venkz:
@jbrooksuk  Please let us know if you need any help with this one. Happy to contribute.

This is a critical feature and would be helpful to have it functional.

PR #3950, opened 2020-02-27
Fix RemoteUserAuthenticate running after Authenticate

svensp:
Having the middleware RemoteUserAuthenticate run after Authenticate does nothing - a request requiring a login was already redirected away before reaching the logic to load the user from the REMOTE_USER server setting.

Moving RemoteUserAuthenticate before Authenticate fixes this.


jbrooksuk:
You're right! Good fix :)

PR #3980, opened 2020-04-11
add a CSS class for status to each component element

luto:
Cachet offers a feature to add additional CSS. This PR builds upon that functionality by enabling users to style components on the main page based on their status. Each component's `<li>` now gets a `status-0` (... `status-4`) class based on their status.

This can be matched against in CSS using `.component.status-1` to target all operational components or `.component.status-4` to match ones that are experiencing a major outage.


ibotty:
Any update?

PR #4020, opened 2020-05-12
Fix for: Click on "manage subscription" causes "invalid signature"

langemeijer:
This fixes issue: https://github.com/CachetHQ/Cachet/issues/4013

PR #4029, opened 2020-05-21
No components -> no manage subscription

langemeijer:
I count the number of components (active ones) to determine if the manage link should be in the notifications. Only if the number of components is higher than 1 it is shown.

Obviously, having a single component is the same as having no components. There is nothing to choose for a subscriber. Either you do or you don't receive notifications.

Note that I build this on my other PR https://github.com/CachetHQ/Cachet/pull/4020


jbrooksuk:
I think we should move this code into either `helpers.php` or `System`?

jbrooksuk:

PR #4093, opened 2020-08-13
2.4 ssl options for mysql

zchef2k:


jbrooksuk:
This should be added manually.

PR #4137, opened 2020-10-19
Fix language file for de.

grossmane:
New stickied incidents will not tranlated if browser language is "de".


jbrooksuk:
Thanks for this, but please make translation changes at https://translate.cachethq.io

PR #4178, opened 2020-12-04
fix(dashboard): RSS feed is no more avariable

lermit:
Hi

As the RSS feed referenced here https://github.com/CachetHQ/Cachet/blob/2.4/app/Integrations/Core/Feed.php#L31 is no more avariable the retreiving of last feed failed.

This PR handle this by disabling the feed if not avariable.

Regards


jbrooksuk:
Thanks, I've applied this manually.

PR #4203, opened 2021-01-15
Fix API search with pagination when specifying page

sedan07:
Currently if you use the search feature of the API along with pagination, specifying the page you want (`page` param) Cachet throws a 500 error as it adds `page` and value you gave the param as a WHERE clause, even though the table in question doesn't have a `page` column.

**Broken:** `/api/v1/incidents?per_page=6&page=1&status=3`
**Works:** `/api/v1/incidents?per_page=6&page=1`
**Works:** `/api/v1/incidents?per_page=6&status=1`
**Works:** `/api/v1/incidents?status=1`

### DIFF
--- a/app/Models/Traits/SearchableTrait.php
+++ b/app/Models/Traits/SearchableTrait.php
@@ -34,10 +34,11 @@ trait SearchableTrait
             return $query;
         }
 
-        if (!array_intersect(array_keys($search), $this->searchable)) {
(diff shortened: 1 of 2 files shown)


jbrooksuk:
Great, thanks!

phith0n:
Hi, this commit is also a security patch, which is fix a SQL injection vulnerability.

I have sent a detail to you, release a new security advisories please.

PR #4219, opened 2021-02-16
Support binding cachet into a subpath.

w4-sjcho:


jbrooksuk:
Thank you for your PR however, Laravel does not support this, so we do not advise doing this.

PR #4220, opened 2021-02-19
Implement masonry in status page.

maddes:
Much more consumable when the amount of cards is large.

Current:

image

With this PR changes:

image


maddes:
Shameless plug: I only tried it live on the demo page and as an settings override on our organization, haven't really built the .scss so it might have typos or .scss issues if builds to a different css.

I'll try building the project later but if someone else has it at hand and can try, it will be greatly appreciated as I'm not a maintainer of this repo (just created this out of necessity, we have many components).

Issue #3952, opened 2020-02-27
APP_URL not taken into account : ugly hotfix


thomvaill:
Hello,

I am running Cachet behind an ALB, which is itself behind an HAProxy (because exposed on an intranet).
The ALB performs the SSL termination and the HAProxy exposes Cachet on a custom port.
As a result, Cachet is completely confused when generating its URLs:

- uses `http://` instead of `https://`
- does not take into account the custom port

I thought the `APP_URL` would solve my issue, but it is unfortunately not taken into account at all.
A lot of issues reference this problem:

- https://github.com/CachetHQ/Docker/issues/262
- https://github.com/CachetHQ/Docker/issues/271
- https://github.com/CachetHQ/Cachet/issues/1545
- https://github.com/CachetHQ/Cachet/issues/2522

Each time, we see only one solution: patch a file with `\URL::forceSchema('https');`...
Unfortunately, the solution, in addition to being ugly, is not complete because the redirection after login to the dashboard is still not fixed.

So, for those who might be interested, here is how I managed to patch the Docker image to make it fully work:

```Dockerfile
FROM cachethq/docker:2.3.15

RUN set -e; \
    # HOTFIX #1: force root URL and schema
    # We locate the boot() method in AppServiceProvider.php
    line=$(grep -n 'public function boot(Dispatcher $dispatcher)$' /var/www/html/app/Foundation/Providers/AppServiceProvider.php | tail -n1 | cut -f1 -d:); \
\
    # We insert the code 2 lines after because there is a "{" on a separate line
    insertAtLine=$((line+2)); \
\
    # We insert the hotfix
    sed -i "$insertAtLine i \\
        // Begin hotfix \n \
        if (getenv('APP_URL')) { \n \
            if (strpos(getenv('APP_URL'), 'https') === 0) { \n \
                \\\URL::forceSchema('https'); \n \
            } \n \
            \\\URL::forceRootUrl(getenv('APP_URL')); \n \
        } \n \
        // End hotfix \n \
    " /var/www/html/app/Foundation/Providers/AppServiceProvider.php; \
\
    # We forward the APP_URL environment variable to FPM
    echo '[www]' > /etc/php7/php-fpm.d/app-url-fix.conf; \
    echo 'env[APP_URL] = $APP_URL' >> /etc/php7/php-fpm.d/app-url-fix.conf; \
\
    # HOTFIX #2: fix login redirection to dashboard
[...]

PR #4293, opened 2021-07-14
feat: FORCE_HTTPS_URLS env to force \URL schemes to https

lbennett-stacki:
For https://github.com/CachetHQ/Docker/issues/158

 I have cachet behind SSL terminating load balancer. Cachet is rendering pages with assets to request using the `http` protocol. Browsers will block these assets if you are accessing over https, even if the asset ingress forces redirects.


ChrisTG742:
We use this fix here and it's working fine. Please consider merging it into base repo.

arrfab:
Came here because of the same issue and wondering when that simple fix can be merged as it seems it's affecting quite some users when running cachethq behind TLS balancer/reverse proxy.
Don't know what's the best approach for the time being to easily deploy the container, maybe apply the patch and rebuild a container and so not deploy the official one ?

arrfab:
some links for issues : 
#3952

arrfab:
FWIW (for people willing to use it as container), I just built this for my own usage : https://quay.io/repository/farrotin/cachethq
So basically : 
 * using original/upstream container image
 * applying this patch

PR #4369, opened 2024-01-14
Added a one-click deploy template

cosark:
Added a third-party hosting section and a link to the 1-click installer template at RepoCloud.io - allowing users access to their affordable cloud hosting.

(This link also qualifies the cachet team to earn 25% recurring lifetime referral commissions)


jbrooksuk:
Thank you, but we've tried one-click deploys previously with Heroku and that didn't work too well.

PR #4371, opened 2024-01-29
[3.x] Implement the filament user contract

ashleyshenton:
This PR adds the FilamentUser interface to the user model to prevent a 403 when logging in on a non-local environment. I've set it to just return true to allow all users access to the cachet dashboard as I believe this is the desired functionality. I can amend this if needed.

I've updated the test account email in the README to reflect the user that is seeded and mentioned in the recent blog post.


jbrooksuk:
Thanks, @ashleyshenton!

jbrooksuk:
This is now deployed and all fixed.

Issue #4047, opened 2020-06-16
Component not updating the Updated_at if its was in same state


ritheshrithu:
Hi All,

I am using the Cachet API to update the component, but let say if the component name is "API" and its current status is "OPERATIONAL", if i update it again to "OPERATIONAL" the last_updated_at time is displaying old time when the component updated from one state to other.

I have tried to modify the **Bus\Handlers\Commands\Component\UpdateComponentCommandHandler** by removing the conditional check,

`if ($command->status && (int) $originalStatus !== (int) $command->status) {
event(new ComponentStatusWasChangedEvent($this->auth->user(), $component, $originalStatus, $command->status, $command->silent));
}`

But still the same. Could someone please assist here

jbrooksuk:
Thank you for your input on Cachet 2.x. We are shifting our attention and resources to Cachet 3.x and will no longer be supporting the 2.x version. If your feedback or issue is relevant to the 3.x series, we encourage you to engage with the new branch.

For more information on the Cachet rebuild and our plans for 3.x, you can read the announcement [here](https://github.com/CachetHQ/Cachet/discussions/4342).

We appreciate your understanding and look forward to your contributions to the new version.

Issue #2854, opened 2017-12-19
cachet.components.status.0 upon updating Component from API


nadeaus:
After digging around the internet for this known issue - I was unable to find a solution that works for us.

I'm trying to do a simple CURL request to update the status of a certain component. Here is the command:

curl -H "Content-Type: application/json;" -H "X-Cachet-Token: xxxxxxxxxxx" --request PUT --url http://status.xxxxxx.com/api/v1/components/29 --data '{"status":2}'

Each time I run this command, the component status gets set to 0, and looks like this:

cachet.components.status.0

Can someone please help me shine some light on this issue?

aojari:
I am having the same issue but in addition am getting problems creating incidents.  I get a message back saying that name field is required etc despite the fact that I have defined them.

jbrooksuk:
What version of Cachet are you using?

aojari:
I have tried it on 2.4 and 2.3.13 both with the same result

2Belette:
hi,
I am struggling after multiple hours of testing with no result and I am wondering if I am not on the same situation.
I am trying to simply update a component with a PUT request.
My idea is to simply update the status of the component (but I also tested with other fields) and no success but no error...nor authentication issue nothing only a 200 OK :( :)

Using requests from python & Cachet 2.4.0-dev

Not sure it is related to this issue.
If you have any idea! will spend more hours on it anyway...

nadeaus:
Currently on version 2.3.9 of Cachet @jbrooksuk

2Belette:
@jbrooksuk : any idea / tips regarding how to update a component status? Is it possible to do it like it is documented ? (see my previous comment here).
I tried hard but I am unable to make it work :(

finndev:
Same issue, latest version of cachet. Unable to update component's status using a PUT request as documented per API.

nadeaus:
I'm on version 2.3.9. Any ideas?

tomaik:
@nadeaus @aojari @finndev  @2Belette 

I'm using 2.3.13 version (docker image) and it works for me.

`curl -X PUT \
  http://example.com/api/v1/components/25 \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  -H 'X-Cachet-Token: xxxxxxxxxxx' \
  -d status=1`

marcel-dempers:
Having the same issue even with 2.3.13 version of the docker image. 
[...]

PR #4403, opened 2024-06-18
Dashboard Api Controller: call UpdateComponentCommand only with modified params

ppomes:
After merging #4402, changing a component state with no tags creates a tag "[]" in DB (an empty array to string).

Fix: calls to UpdateComponentCommand from dash controller should be done only on modified params, as it is done on incident creation/updates.

### DIFF
--- a/app/Http/Controllers/Dashboard/ApiController.php
+++ b/app/Http/Controllers/Dashboard/ApiController.php
@@ -38,15 +38,15 @@ class ApiController extends AbstractApiController
         try {
             execute(new UpdateComponentCommand(
                 $component,
-                $component->name,
-                $component->description,
+                null,
+                null,
                 Binput::get('status'),
-                $component->link,
-                $component->order,
-                $component->group_id,
-                $component->enabled,
-                $component->meta,
-                $component->tags,
+                null,
+                null,
+                null,
+                null,
+                null,
+                null,
                 true   // Silent mode
             ));
         } catch (QueryException $e) {
@@ -71,15 +71,15 @@ class ApiController extends AbstractApiController
 
                 execute(new UpdateComponentCommand(
                     $component,
-                    $component->name,
-                    $component->description,
+                    null,
+                    null,
                     $component->status,
-                    $component->link,
+                    null,
                     $order + 1,
-                    $component->group_id,
-                    $component->enabled,
-                    $component->meta,
-                    $component->tags,
+                    null,
+                    null,
+                    null,
+                    null,
                     true   // Silent mode
                 ));
             } catch (QueryException $e) {


app/Http/Controllers/Dashboard/ApiController.php
```diff
@@ -38,15 +38,15 @@ public function postUpdateComponent(Component $component)
...
-                $component->group_id,
-                $component->enabled,
-                $component->meta,
-                $component->tags,
+                null,
+                null,
+                null,
+                null,
+                null,
+                null,
```
jbrooksuk, line 49: This definitely doesn't look right to me...

ppomes: Hum.. that's the same logic implemented anywhere else?  Example from ./app/Bus/Handlers/Commands/Incident/CreateIncidentCommandHandler.php:


```
        // Update the component.
        if ($component = Component::find($command->component_id)) {
            execute(new UpdateComponentCommand(
                Component::find($command->component_id),
                null,
                null,
                $command->component_status,
                null,
                null,
                null,
                null,
                null,
                null,
                true  // Silent mode
            ));
```

To see the problem, try updating a component with no flag in the dashboard multiple times, and have a look to flags in DB after each update. 

Pierre

ppomes: > To see the problem, try updating a component with no flag in the dashboard multiple times, and have a look to flags in DB after each update.

I mean: with my PR's merged last week. They broke components updates from the gui. The current PR fix them.

PolNavarro:
## Excellent fix\! Here are some suggestions to make it even more robust

This PR addresses a real architectural issue with component updates. I've analyzed the codebase and have some specific suggestions to improve the implementation:

### 1. **Enhanced Parameter Filtering**

The current approach of passing all parameters can be improved with more precise filtering:

```php
// Instead of passing all parameters, build only changed ones
private function buildComponentUpdateParams($request, $component): array 
{
    $params = [$component]; // Always include component
    
    // Only add parameters that have actually changed
    if ($request->has('status') && $request->get('status') \!== $component->status) {
        $params['status'] = $request->get('status');
    }
    
    if ($request->has('name') && $request->get('name') \!== $component->name) {
        $params['name'] = $request->get('name');
    }
    
    // Key fix: Don't pass tags if they're empty/null
    if ($request->has('tags') && \!empty($request->get('tags')) && $request->get('tags') \!== '[]') {
        $params['tags'] = $request->get('tags');
    }
    
    return $params;
}
```

### 2. **Prevent Empty Tag Arrays**

Add specific validation to prevent the empty tag issue:

```php
private function shouldUpdateTags($requestTags, $currentTags): bool
{
    // Don't update if tags are empty, null, or empty array string
    if (empty($requestTags) || $requestTags === '[]' || $requestTags === 'null') {
        return false;
    }
    
    // Only update if tags have actually changed
    return json_encode($requestTags) \!== json_encode($currentTags);
}
```

### 3. **UpdateComponentCommand Enhancement**

The command class should handle optional parameters gracefully:

```php
public function __construct(
    Component $component,
    ?string $name = null,
    ?string $description = null,
    ?int $status = null,
    ?array $tags = null,
    // ... other optional parameters
) {
    $this->component = $component;
    
    // Only set properties that are explicitly provided
    if ($name \!== null) $this->name = $name;
    if ($status \!== null) $this->status = $status;
    if ($tags \!== null && \!empty($tags)) $this->tags = $tags;
}
```

### 4. **Testing Scenarios**

To prevent regression, consider testing these scenarios:

```php
// Test case 1: Update component without tags
$response = $this->put('/dashboard/api/components/{id}', [
    'status' => 1
    // No tags provided
]);
// Should NOT create empty tag arrays

// Test case 2: Update with same values
$response = $this->put('/dashboard/api/components/{id}', [
    'status' => $component->status, // Same as current
    'name' => $component->name      // Same as current
]);
// Should not trigger unnecessary updates

// Test case 3: Clear existing tags
$response = $this->put('/dashboard/api/components/{id}', [
    'tags' => null // Explicitly clear tags
]);
// Should properly clear tags without creating '[]'
```

### 5. **Database Impact**

This fix will prevent:
- ✅ Empty tag arrays  in the database
- ✅ Unnecessary  timestamp changes
- ✅ Side effects from passing unchanged parameters
- ✅ Inconsistency between GUI and API updates

### Related Issues
This also addresses similar problems mentioned in:
- #3288 (components getting disabled)
- #4047 (updated_at not updating correctly)
- #2854 (invalid component status values)

Great work identifying this pattern inconsistency\! The incident update approach is definitely the right model to follow. 👍

PolNavarro:
## Excellent fix\! Here are some suggestions to make it even more robust

This PR addresses a real architectural issue with component updates. I've analyzed the codebase and have some specific suggestions to improve the implementation:

### 1. Enhanced Parameter Filtering

The current approach of passing all parameters can be improved with more precise filtering:

```php
// Instead of passing all parameters, build only changed ones
private function buildComponentUpdateParams($request, $component): array 
{
    $params = [$component]; // Always include component
    
    // Only add parameters that have actually changed
    if ($request->has('status') && $request->get('status') \!== $component->status) {
        $params['status'] = $request->get('status');
    }
    
    if ($request->has('name') && $request->get('name') \!== $component->name) {
        $params['name'] = $request->get('name');
    }
    
    // Key fix: Don't pass tags if they're empty/null
    if ($request->has('tags') && \!empty($request->get('tags')) && $request->get('tags') \!== '[]') {
        $params['tags'] = $request->get('tags');
    }
    
    return $params;
}
```

### 2. Prevent Empty Tag Arrays

Add specific validation to prevent the empty tag issue:

```php
private function shouldUpdateTags($requestTags, $currentTags): bool
{
    // Don't update if tags are empty, null, or empty array string
    if (empty($requestTags) || $requestTags === '[]' || $requestTags === 'null') {
        return false;
    }
    
    // Only update if tags have actually changed
    return json_encode($requestTags) \!== json_encode($currentTags);
}
```

### 3. Testing Scenarios

To prevent regression, consider testing these scenarios:

- Update component without tags (should NOT create empty arrays)
- Update with same values (should not trigger unnecessary updates)  
- Clear existing tags explicitly (should clear properly)

### 4. Database Impact

This fix will prevent:
- Empty tag arrays in the database
- Unnecessary timestamp changes
- Side effects from passing unchanged parameters
- Inconsistency between GUI and API updates

Great work identifying this pattern inconsistency\! The incident update approach is definitely the right model to follow. 👍

PR #4417, opened 2024-09-12
Update lang es-ES

martinblascop:
Fix typos


jbrooksuk:
Thanks! Can you explain how this difference makes an impact?

martinblascop:
at my company, we are evaluating cachet, so when I switched to spanish I noticed these typos.

PR #4591, opened 2026-01-03
Bump PHP version due to installation issues

felipee-monteiro:
I was trying to install cachet 3.x with the minimum version (8.2), but the installation process break. Its seems that a 
dependencie called openspout requires a minimum PHP version (8.3). I also update the extensions requirements. Maybe can i update the docs ?

Captura de tela de 2026-01-03 16-31-58

image


LOFWagner:
+1, for my docker build I also had to even use 8.4, 8.3 did not work for me

